PR #1, opened 2018-09-18
String and Cow on `no_std`

bkchr:
@gavofyork could you also please release the new version to crates.io?

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -16,6 +16,8 @@
 
 use alloc::vec::Vec;
 use alloc::boxed::Box;
+use alloc::borrow::{Cow, ToOwned};
+use alloc::string::String;
 use core::{mem, slice};
 use arrayvec::ArrayVec;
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -32,6 +32,8 @@ extern crate arrayvec;
 pub mod alloc {
 	pub use std::boxed;
 	pub use std::vec;
+    pub use std::string;
```
sorpaas: Should use `<tab>` for indentation here.

src/codec.rs
```diff
@@ -258,36 +260,32 @@ impl<'a> Encode for &'a str {
...
+impl<'a, T: ToOwned + ?Sized> Decode for Cow<'a, T> where
 	<T as ToOwned>::Owned: Decode
 {
 	fn decode<I: Input>(input: &mut I) -> Option<Self> {
-		Some(::std::borrow::Cow::Owned(Decode::decode(input)?))
+		Some(Cow::Owned(Decode::decode(input)?))
 	}
 }
 
-#[cfg(feature = "std")]
```
gavofyork: There's a reason for this: we don't want `String` in wasm environments.

gavofyork:
There's a reason these things were disabled in `no_std` - we don't want to expose them in Wasm.

src/codec.rs
pepyakin: But what the reason exactly?

PR #4, opened 2018-09-20
Add Support for PhantomData

gnunicorn:
Allow us to derive Struct that contain `::std::marker::PhantomData`

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -281,6 +281,21 @@ impl<'a, T: ToOwned + ?Sized> Decode for ::std::borrow::Cow<'a, T> where
 }
 
 #[cfg(feature = "std")]
+impl<T> Encode for ::std::marker::PhantomData<T> {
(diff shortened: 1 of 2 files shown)


src/codec.rs
```diff
@@ -280,6 +280,21 @@ impl<'a, T: ToOwned + ?Sized> Decode for ::std::borrow::Cow<'a, T> where
 	}
 }
 
+#[cfg(feature = "std")]
```
pepyakin: Why std only?

gnunicorn: Just sticked with the convention that everything for `::std::` (like the two above) is featured gated behind that flag. I suppose in order to provide no-std/wasm compilation support (where those can't be needed anyways)!?!

pepyakin: Yeah, I just wonder because it seems that using `PhantomData` in `no-std` environment should not harm and I also can imagine legit uses of it.

PR #5, opened 2018-09-21
Compact encoding

gavofyork:


src/codec.rs
```diff
@@ -113,6 +131,108 @@ pub trait Decode: Sized {
 /// Trait that allows zero-copy read/write of value-references to/from slices in LE format.
 pub trait Codec: Decode + Encode {}
 
+/// Compact-encoded variant of T. This is more space-efficient but less compute-efficient.
+pub struct Compact<T>(T);
```
pepyakin: I think we probably want to make the field `pub`. Otherwise the users of the crate won't be able to crate instances of it nor read the value.

rphmeier: `From` and `Into` implementations would be great too.

gavofyork: tried that but didn't work.

gavofyork:
```
error[E0119]: conflicting implementations of trait `std::convert::Into<_>` for type `codec::Compact<_>`:
   --> src/codec.rs:141:1
    |
141 | impl<T> Into<T> for Compact<T> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: conflicting implementation in crate `core`:
            - impl<T, U> std::convert::Into<U> for T
              where U: std::convert::From<T>;
```

gavofyork: (and, no, there's no corresponding impl for `From` in this module)

pepyakin: yeah, I think you have to define them explicitly.

src/codec.rs
```diff
@@ -113,6 +131,121 @@ pub trait Decode: Sized {
 /// Trait that allows zero-copy read/write of value-references to/from slices in LE format.
 pub trait Codec: Decode + Encode {}
 
+/// Compact-encoded variant of T. This is more space-efficient but less compute-efficient.
+pub struct Compact<T>(pub T);
```
dvdplm: This is not exported in `lib.rs` – I think it should be?

PR #6, opened 2018-09-30
Impl Codec for Compact<u64> and Compact<u128>

pepyakin:
Should be useful for https://github.com/paritytech/substrate/issues/847


bkchr:
Could you please rebase with master, so that we can test that CI works?

PR #7, opened 2018-10-02
Add CI build config

ddorgan:


.gitlab-ci.yml
```diff
@@ -0,0 +1,37 @@
...
+    - ./target/
+    - ./.cargo/
+
+
+#### stage:                        test
+
+test:rust:stable:
+  stage:                           test
+  script:
+    - time cargo test --all --release
```
pepyakin: I would rather go without `--release`

.gitlab-ci.yml
```diff
@@ -0,0 +1,37 @@
...
+variables:
+  CI_SERVER_NAME:                  "GitLab CI"
+  CARGO_HOME:                      "${CI_PROJECT_DIR}/.cargo"
+
+
+cache:
+  key:                             "${CI_JOB_NAME}"
+  paths:
+    - ./target/
+    - ./.cargo/
```
pepyakin: Do we actually need a cache here?

.gitlab-ci.yml
ddorgan: Removed --release

.gitlab-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+  CARGO_HOME:                      "${CI_PROJECT_DIR}/.cargo"
+
+#### stage:                        test
+
+test:rust:stable:
+  stage:                           test
+  script:
+    - time cargo test --all
+  tags:
+    - rust-stable
```
bkchr: What does this tag do? Above you specify `rust:nightly` as image and the tag says stable?

.gitlab-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+
+
+#### stage:                        build
+
+build:linux:ubuntu:amd64:
+  stage:                           build
+  script:
+    - cargo build
+  tags:
+    - rust-stable
```
bkchr: Same as above.

.gitlab-ci.yml
ddorgan: This image actually has more than one toolchain:

```
# rustup toolchain list
stable-x86_64-unknown-linux-gnu (default)
beta-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu
```

We only use the tag to assign to a gitlab-runner. But maybe we should just use the tag rust instead. Will change.

PR #12, opened 2018-10-12
Implements `encoded_as` and `compact`

bkchr:
Fixes: https://github.com/paritytech/substrate/issues/879

cc: @shawntabrizi

### DIFF
--- a/derive/src/encode.rs
+++ b/derive/src/encode.rs
@@ -39,7 +39,7 @@ fn encode_fields<F>(
 		let encode_as = super::get_encode_type(f);
 
 		if encode_as.is_some() {
-			let ts = quote!{ <Compact<u64>> };
+			let ts = quote!{ <<Foo as HasCompact>::Type> };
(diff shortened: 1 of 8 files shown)


bkchr:
Yep. I tested it with substrate.

derive/src/encode.rs
```diff
@@ -36,9 +34,22 @@ fn encode_fields<F>(
...
 
-		quote_spanned! { f.span() =>
-			#dest.push(#field);
+		if encoded_as.is_some() && compact {
+			panic!("`encoded_as` and `compact` can not be used at the same time!");
+		}
+
+		if compact {
+			let field_type = &f.ty;
+			quote_spanned! { f.span() => { #dest.push(&Compact::<#field_type>::from(#field)); } }
```
0x7CFE: It would be a good idea to add comment on what particular `push` is called, since it's quite ambigious. Another option would be to rewrite the whole line as `<#dest as Output>::push(#field)` which would be self explanatory.

PR #15, opened 2018-11-06
Fixes tests and reduce required scope

bkchr:


src/lib.rs
```diff
@@ -44,3 +44,13 @@ mod keyedvec;
...
 pub use self::joiner::Joiner;
 pub use self::keyedvec::KeyedVec;
+
+/// Decode an object from the given slice.
+pub fn from_slice<R: Decode>(data: &[u8]) -> Option<R> {
+	R::decode(&mut &data[..])
+}
+
+/// Decode an object from the given vector.
+pub fn from_vec<R: Decode>(data: Vec<u8>) -> Option<R> {
```
pepyakin: Do we actually need this one? Seems like superflous to me

src/codec.rs
```diff
@@ -187,12 +187,13 @@ impl<'de, T> ::serde::Deserialize<'de> for Compact<T> where T: ::serde::Deserial
 impl_from_compact! { u8, u16, u32, u64, u128 }
 
 /// Trait that tells you if a given type can be encoded/decoded in a compact way.
-pub trait HasCompact: Sized {
-	/// The compact type; this can be 
+pub trait HasCompact: Sized + Clone {
```
pepyakin: Just out of curiousity why is this change needed?

src/lib.rs
bkchr: I think in certain situations this can be helpful. 
I wanted to `decode` something in a macro, without wanting to know the type it should decode to.
With this function I can just write `from_vec(vec)` and the compiler itself finds out the return type.
`Serde` for example provides similar functions.

Robbepop: Such APIs can easily become a performance trap.

Can you please elaborate on the helpful situation a bit? Maybe an example.

bkchr: Why should that be a performance trap? I would never expect the compiler to not optimize it away?

Robbepop: For example a user has a vector `v`, calls this without knowing about the `from_slice` thus cloning `v`. It is just not intuitive taking something non-Copy by value if you do not need it.

bkchr:
As I don't require the functions anymore, I will remove them. I will update the pr to just fix the tests.

bkchr:
@Robbepop @pepyakin I pushed a fix for the tests.

Besides fixing the tests it also fixes: https://github.com/paritytech/parity-codec/issues/17

Robbepop:
I think this PR needs renaming now. :D

src/codec.rs
```diff
@@ -207,31 +227,43 @@ impl<T> HasCompact for T where
...
+	}
+}
+
+impl<'a> Encode for CompactRef<'a, u16> {
 	fn encode_to<W: Output>(&self, dest: &mut W) {
 		match self.0 {
-			0...0b00111111 => dest.push_byte((self.0 as u8) << 2),
-			0...0b00111111_11111111 => ((self.0 << 2) | 0b01).encode_to(dest),
-			_ => (((self.0 as u32) << 2) | 0b10).encode_to(dest),
+			0...0b00111111 => dest.push_byte((*self.0 as u8) << 2),
```
Robbepop: It would be very nice to have some more comments on code like this.
However, I can understand that this might be out of range for this PR.
Not stopping merging because of this.

src/codec.rs
```diff
@@ -184,16 +194,26 @@ impl<'de, T> ::serde::Deserialize<'de> for Compact<T> where T: ::serde::Deserial
 	}
 }
 
-impl_from_compact! { u8, u16, u32, u64, u128 }
-
 /// Trait that tells you if a given type can be encoded/decoded in a compact way.
-pub trait HasCompact: Sized {
-	/// The compact type; this can be 
-	type Type: Encode + Decode + From<Self> + Into<Self> + Clone + PartialEq + Eq + MaybeDebugSerde;
+pub trait HasCompact: Sized + 'static {
```
Robbepop: For what is the `'static` restriction required?

src/codec.rs
bkchr: There is already documentation for it in the source file.

src/codec.rs
Robbepop: Resolved since removed.

Robbepop:
LGTM!

src/codec.rs
```diff
@@ -184,16 +194,26 @@ impl<'de, T> ::serde::Deserialize<'de> for Compact<T> where T: ::serde::Deserial
...
 }
 
-impl_from_compact! { u8, u16, u32, u64, u128 }
-
 /// Trait that tells you if a given type can be encoded/decoded in a compact way.
 pub trait HasCompact: Sized {
-	/// The compact type; this can be 
-	type Type: Encode + Decode + From<Self> + Into<Self> + Clone + PartialEq + Eq + MaybeDebugSerde;
+	/// The compact type; this can be
+	type Type: for<'a> EncodeAsRef<'a, Self> + Decode + From<Self> + Into<Self> + Clone +
```
0x7CFE: Isn't it redundant to require `Into<Self>` when `From<Self>` is also required? Standard library contains a trivial implementation of `Into<B>` for every `A` if there exists an `impl From<A> for B`.

PR #16, opened 2018-11-09
provide  BTreeMap Encode/Decode

JKingdom:


bkchr:
Could you please provide a test?

JKingdom:
> Could you please provide a test?

Sure, I have add a test now, if it's not enough, please tell me, I would add more tests, ths

PR #19, opened 2018-12-04
.gitlab-ci.yml build stage on master only

gabreal:


bkchr:
Why do we have this `build` stage at all? `cargo test` is already building the project. Is that in anyway different?

gabreal:
In other projects cargo build --release is used for buliding releases. Tests won't be run there and builds are optimized.

bkchr:
Yeah that makes sense, but we don't ship `parity-codec` releases :D

PR #26, opened 2019-01-03
Feature/btree map

JKingdom:
provide btree_map encode/decode for parity-codec, under std/wasm 
how to do:
btree_map<K, V> => Vec<(K, V)>
just use tuple vec method to encode/decode
have provided test case for encode/decode

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,14 +24,14 @@ name = "parity-codec-derive"
 version = "2.1.0"
 dependencies = [
  "parity-codec 2.1.5",
- "proc-macro2 0.4.19 (registry+https://github.com/rust-lang/crates.io-index)",
- "quote 0.6.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proc-macro2 0.4.20 (registry+https://github.com/rust-lang/crates.io-index)",
+ "quote 0.6.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "syn 0.14.9 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


Robbepop:
Hi,
thank you for your pull request!

We might want to merge your PR as soon as https://github.com/paritytech/parity-codec/pull/27 got merged.
Note that your implementation must be provided under the `"full"` crate feature.

JKingdom:
> Hi,
> thank you for your pull request!
> 
> Maybe we will merge your PR as soon as #27 got merged. (Still under discussion since we do not want to bloat this core library too much on the other side.)
> 
> Note that for the merge your implementation must then be provided under the `"full"` crate feature this is going to be introduced in the aforementioned #27.

Thank you very much. In fact in our parachain project, we use BTreeMap in `decl_storage!`, so that we need this update soon to support our project.
We use a wrapper struct to solve it now, but the metadata would export the wrapper struct name, which we hope to use `BTreeMap` self to be more unified in our project.

Though it's not very urgent, but still hope to solve it as soon as possible, thx :)

Robbepop:
With https://github.com/paritytech/parity-codec/pull/27 we might be able to merge this is you perform the required adjustments - i.e. put the implementing code under the new `full` crate feature.

Note that the decision to implement this new crate feature wasn't easy since in general consensus code (oftentimes code depending on `parity-codec`) should just be the logical parts of the greater system. This excludes shiny types like `String` from our advised use case. However, since this is Rust and since we cannot foresee any reasonable use cases we wanted to provide some way to go around this.

PR #27, opened 2019-01-08
Add no_std impl for some types

Robbepop:
Adds `no_std` impls for `Encode` and `Decode` for:
- `alloc::string::String`
- `alloc::borrow::Cow`
- `core::marker::PhantomData`

Also adds `Debug` impl in `no_std` for
- `Compact<T>`
- `OptionBool`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,19 @@ default = ["std"]
 derive = ["parity-codec-derive"]
 std = ["serde"]
 
+# WARNING: DO _NOT_ USE THIS FEATURE IF YOU ARE WORKING ON CONSENSUS CODE!*
+#
+# Provides implementations for more data structures than just Vec and Box.
+# Concretely it will provides parity-codec implementations for many types
+# that can be found in std and/or alloc (nightly).
+#
+# This feature was mainly introduced after it became clear that pDSL requires
+# it for the sake of usability of its users.
+#
+# * For rational about this please visit:
+# https://github.com/paritytech/parity-codec/pull/27#issuecomment-453031914
+full = []
+
 [workspace]
 members = [
 	"derive",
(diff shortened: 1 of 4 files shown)


bkchr:
https://github.com/paritytech/parity-codec/pull/1#issuecomment-422698476

Robbepop:
@bkchr I am failing to see the reason behind not wanting to expose fundamental types such as `String` and `PhantomData` (especially that one). So what is the reason behind this decision?

It would be a great life improvement for all crates depending on `parity-codec` and they will most certainly provide impls themselves anyway if they are in need of them. That's at least what I would do if I fail to see the reason behind this decision.

bkchr:
For the reason behind that, you need to ask @gavofyork. I don't know it.

Robbepop:
Okay thanks - will do that.

Would be very nice to know about the exact reasons behind this decision for all of the developers. :)

Robbepop:
Wanted to add that also serde provides a crate feature to enable impls for alloc collections:
https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml#L71

Robbepop:
For the record: I talked to @gavofyork and his reply to this question and the PR was the following:

> Consensus code shouldn’t be including human-readable stuff like String. It’s for pure logic, not for text manipulation.

And also ...

> I’m happy for you to add a (non-default) feature, though.

So I will rework this PR slightly to be guarded by a crate feature.
The name of the crate feature should be useful to tell people not using it for consensus code.

Cargo.toml
```diff
@@ -12,10 +12,21 @@ parity-codec-derive = { version = "2.0", default-features = false, optional = tr
...
-std = [
-	"serde"
-]
 derive = ["parity-codec-derive"]
+std = ["serde"]
+
+# WARNING: DO _NOT_ USE THIS FEATURE IF YOU ARE WORKING ON CONSENSUS CODE!*
+#
+# Provides implementations for more data structures than just Vec and Box.
+# Concretely it will provides parity-codec implementations for many types
```
gavofyork:
```suggestion
# Concretely it will provide parity-codec implementations for many types
```

PR #31, opened 2019-01-15
Resolve issue + extend Compact

thiolliere:
Resolve issue: 
macro modification, and adding of trait requirements are to make it compatible with substrate as it is now
* Encode might not be mandatory by modifying decl_module macros when it derive Encode for Call enum and some code
* From could be implemented using a trick (encode as ref and then decode)

Extend compact:
related to https://github.com/paritytech/substrate/issues/1405
to allow Permill to be written as Compact<Permill> we provide a trait to implement
then on substrate we will have do
```rust
impl codec::CompactAs for Permill {
       type As = u32;
       fn encode_as(&self) -> &u32 {
               &self.0
       }
       fn decode_from(x: u32) -> Permill {
               Permill(x)
       }
}

impl From<codec::Compact<Permill>> for Permill {
       fn from(x: codec::Compact<Permill>) -> Permill {
               x.0
       }
}
```

or we could provide some macro to help but seems OK

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,7 @@ dependencies = [
  "arrayvec 0.4.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "parity-codec-derive 2.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.80 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


thiolliere:
Test Added

Also I modified CompactAs so it requires From trait. This is needed to be HasCompact

PR #34, opened 2019-01-18
generalize #[codec(compact)]

thiolliere:
By using HasCompact trait we can generalize the code generation to derive compact.

Thus this allow us to write
```rust
struct TestCompactHasCompact<T: HasCompact> {
	#[codec(compact)]	
        // instead of #[codec(encoded_as = "<T as HasCompact>::Type")]
	bar: T,
}
```

this has also been tested with substrate test --all


thiolliere:
related to https://github.com/paritytech/substrate/issues/907

PR #40, opened 2019-01-23
Trait bounds for associated types

0x7CFE:
### The problem

Previously custom derive for `Encode` and `Decode` traits used straightforward approach to generic types: it just added `: (En|De)code` bound to every declared type parameter it could find in the type definition.

So, for a struct like:
```rust
#[derive(Encode, Decode)]
struct Struct<T> {
	field: T
}
```

Custom derive would generate impls like this:
```rust
impl<T> Encode for Struct<T> where T: Encode { ... }
```

In most cases, this works just fine. However, in others, you may want to specify type parameter that is not expected to be encodable at all. Instead, it's used to pass several types at once. This pattern is prevalent in Substrate and SRML where we usually pass `T: Trait` here and there and downstream types use this `T` to specialize their structures with associated types. Otherwise, it would be impractical to pass dozens of configuration types instead of a single `T`.

For example, in [UTXO prototype](https://github.com/paritytech/substrate-node-template/pull/17) I'd like to use the following enum:

```rust
#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
#[derive(Clone, Encode, Decode, Hash)]
pub enum LockStatus<T: Trait>
{
	Locked,
	LockedUntil(T::BlockNumber),
}
```

and define the storage as:
```rust
decl_storage! {
	trait Store for Module<T: Trait> as Utxo {
		// ...
		LockedOutputs: map H256 => Option<LockStatus<T>>;
	}
}
```
Note that `LockStatus` struct does not expect the whole `T` to be encodable or decodable, it just needs that `T` to extract `T::BlockNumber` type. Obviously, previous implementation of custom derive didn't allowed me to do that since it would blindly create the impl with incorrect trait bound:  
```rust 
impl<T> Encode for LockStatus<T> where T: Trait, T: Encode { ... }
```

### The solution

This PR tries to address the issue by bounding actual types that are used, not the type parameters. When doing this, it exploits two facts:
- Rust allows trivial or redundant trait bounds
- Rust has a powerful type system and inference rules

The first fact allows us to write bounds several times without any warning or compilation error. For example, if we have a struct that declares two fields using the same associated type:
```rust
#[derive(Encode, Decode)]
struct Struct<T: Trait> {
	field1: T::AssociatedType,
	field2: T::AssociatedType,
}
```

upon impl generation, we would add `T::AssociatedType: Encode` bound twice, but that's OK.

The second fact allows us to bound only the topmost type of the field, even if this type is in turn generic:

```rust
#[derive(Encode, Decode)]
struct Struct<T: Trait> {
	field1: Vec<T::Type1>,
	field2: Option<(T::Type1, T::Type2)>,
}
```

In this situation we add `Vec<T::Type1>: Encode` and `Option<(T::Type1, T::Type2)>: Encode` bounds, but Rust is powerful enough to understand that there exists an `impl Encode for Vec<T> where T: Encode`, so this would be effectively equal to `T::Type1: Encode`. The second field processed in the same way, but the reasoning chain would be one link longer.

P.S.: The same applies to `Decode` as well.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -25,7 +25,7 @@ extern crate quote;
 
 use proc_macro::TokenStream;
 use proc_macro2::Span;
-use syn::{DeriveInput, Generics, GenericParam, Ident};
+use syn::{DeriveInput, Generics, Ident};
 
 mod decode;
 mod encode;
@@ -36,10 +36,9 @@ const ENCODE_ERR: &str = "derive(Encode) failed";
 #[proc_macro_derive(Encode, attributes(codec))]
 pub fn encode_derive(input: TokenStream) -> TokenStream {
 	let mut input: DeriveInput = syn::parse(input).expect(ENCODE_ERR);
-	patch_generics(&mut input.generics, &input.data, parse_quote!(_parity_codec::Encode));
-	let name = &input.ident;
+	add_trait_bounds(&mut input.generics, &input.data, parse_quote!(_parity_codec::Encode));
 
-	// let generics = add_trait_bounds(input.generics, parse_quote!(_parity_codec::Encode));
+	let name = &input.ident;
 	let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
 
 	let self_ = quote!(self);
@@ -75,10 +74,9 @@ pub fn encode_derive(input: TokenStream) -> TokenStream {
 #[proc_macro_derive(Decode, attributes(codec))]
 pub fn decode_derive(input: TokenStream) -> TokenStream {
 	let mut input: DeriveInput = syn::parse(input).expect(ENCODE_ERR);
-	patch_generics(&mut input.generics, &input.data, parse_quote!(_parity_codec::Decode));
-	let name = &input.ident;
+	add_trait_bounds(&mut input.generics, &input.data, parse_quote!(_parity_codec::Decode));
 
-	// let generics = add_trait_bounds(input.generics, parse_quote!(_parity_codec::Decode));
+	let name = &input.ident;
 	let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
 
 	let input_ = quote!(input);
@@ -110,7 +108,7 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
 	generated.into()
 }
 
-fn patch_generics(generics: &mut Generics, data: &syn::Data, bound: syn::Path) {
+fn add_trait_bounds(generics: &mut Generics, data: &syn::Data, bound: syn::Path) {
 	let types = collect_types(&data);
 	if !generics.params.is_empty() && !types.is_empty() {
 
@@ -123,26 +121,11 @@ fn patch_generics(generics: &mut Generics, data: &syn::Data, bound: syn::Path) {
 		for ty in types {
 			where_clause.predicates.push(parse_quote!(#ty : #bound));
 		}
-		
-		println!("predicates {:#?}", where_clause.predicates);
 	}
 }
 
 fn collect_types(data: &syn::Data) -> Vec<syn::Type> {
 	use syn::*;
-	use syn::visit::Visit;
-
-	struct TypeVisitor;
-
-	impl<'ast> Visit<'ast> for TypeVisitor {
-		fn visit_type_path(&mut self, i: &'ast TypePath) {
-			// println!("visited type {:#?}", i);
-			self.visit_path(&i.path);
-		}
-	}
-
-	let mut visitor = TypeVisitor;
-	// visitor.visit_data(&data);
 
 	match *data {
 		Data::Struct(ref data) => match &data.fields {
@@ -168,13 +151,4 @@ fn collect_types(data: &syn::Data) -> Vec<syn::Type> {
 
 		Data::Union(_) => panic!("Union types are not supported."),
 	}
-}
-
-fn add_trait_bounds(mut generics: Generics, bounds: syn::TypeParamBound) -> Generics {
-	for param in &mut generics.params {
-		if let GenericParam::Type(ref mut type_param) = *param {
-			type_param.bounds.push(bounds.clone());
-		}
-	}
-	generics
 }
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -35,11 +35,12 @@ const ENCODE_ERR: &str = "derive(Encode) failed";
 
 #[proc_macro_derive(Encode, attributes(codec))]
 pub fn encode_derive(input: TokenStream) -> TokenStream {
-	let input: DeriveInput = syn::parse(input).expect(ENCODE_ERR);
+	let mut input: DeriveInput = syn::parse(input).expect(ENCODE_ERR);
+	patch_generics(&mut input.generics, &input.data, parse_quote!(_parity_codec::Encode));
 	let name = &input.ident;
 
-	let generics = add_trait_bounds(input.generics, parse_quote!(_parity_codec::Encode));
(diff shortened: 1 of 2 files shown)


derive/src/lib.rs
```diff
@@ -108,11 +108,47 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+				| Fields::Unnamed(FieldsUnnamed { unnamed: fields, .. }) => {
+					fields.iter().map(|f| f.ty.clone()).collect()
+				},
+
+				Fields::Unit => { vec![] },
+			}
+		})
+		.collect(),
+
+		Data::Union(_) => panic!("Union types are not supported."),
```
bkchr: Please no `panic`, make the function return a Result and at the point where you generate code, use https://docs.rs/syn/0.15.26/syn/parse/struct.Error.html#method.to_compile_error
This will show a nice compile error.

derive/src/lib.rs
```diff
@@ -108,11 +108,47 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+			match &variant.fields {
+				| Fields::Named(FieldsNamed { named: fields , .. })
+				| Fields::Unnamed(FieldsUnnamed { unnamed: fields, .. }) => {
+					fields.iter().map(|f| f.ty.clone()).collect()
+				},
+
+				Fields::Unit => { vec![] },
+			}
+		})
+		.collect(),
```
bkchr: Nit: This could be moved up one line.

tests/mod.rs
```diff
@@ -267,3 +267,27 @@ fn enum_compact_meta_attribute_works() {
 		}
 	}
 }
+
+#[test]
+fn associated_type_bounds() {
+	trait Trait {
+		type AssociatedType;
```
bkchr: Could you add a second associated type that does not implement `Encode`?

derive/src/lib.rs
0x7CFE: Thanks for the suggestion. However, that's how it was implemented before. Also, code in `encode` and `decode` modules panics too.

bkchr: Yeah, but from an usage perspective it is not nice. If we touch this code, we can improve it :)

0x7CFE: Ok, will change that code as well.

thiolliere:
if wanted I can finish this PR

bkchr:
Yeah that is a breaking change, but just because rust `std` has a broken derive, does not mean we need to follow this brokenness. 
@0x7CFE what is your status on this?

0x7CFE:
I'm trying to provide proper spanned compiler errors instead of panics. Unfortunately, I don't quite get how to extract span information from `DeriveInput` stuff. The idea is to highlight erroneous entry that could not be encoded. Examples are unions and enums with more than 256 items.

pepyakin:
@0x7CFE don't fields implement `Spanned`? If they do, you can do something like [this](https://github.com/paritytech/parity-codec/pull/40/files#diff-11de442f8c13daab760fd46822928e3dR160).

0x7CFE:
@bkchr, I've addressesed most of the issues.

Using this version I was able to build substrate and my UTXO prototype and it worked. Also, I used the new feature and it worked too:

```rust
#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
#[derive(Clone, Encode, Decode, Hash)]
pub enum LockStatus<T: Trait>
{
	Locked,
	LockedUntil(T::BlockNumber),
}
```

Remaining issues are mostly due to `panic!`s in existing code. I think I need to refactor them to `Result`s as well.

tests/mod.rs
```diff
@@ -267,3 +267,29 @@ fn enum_compact_meta_attribute_works() {
...
+	struct Struct<T: Trait, Type> {
+		field: (Vec<T::EncodableType>, Type),
+	}
+
+	#[derive(Debug, PartialEq)]
+	struct TraitImplementor;
+
+	impl Trait for TraitImplementor {
+		type EncodableType = u32;
+		type NonEncodableType = std::marker::PhantomData<Self>;
```
bkchr: We provide an implementation of `Encode` for `PhantomData`. So this is an encodable type. Just add a random struct that represents the `NonEncodableType`.

bkchr:
Which remaining issues?

derive/src/lib.rs
```diff
@@ -108,11 +118,49 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+	use syn::*;
+
+	let types = match *data {
+		Data::Struct(ref data) => match &data.fields {
+			| Fields::Named(FieldsNamed { named: fields , .. })
+			| Fields::Unnamed(FieldsUnnamed { unnamed: fields, .. }) => {
+				fields.iter().map(|f| f.ty.clone()).collect()
+			},
+
+			Fields::Unit => { vec![] },
```
bkchr:
```suggestion
			Fields::Unit => { Vec::new() },
```

derive/src/lib.rs
```diff
@@ -108,11 +118,49 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+		},
+
+		Data::Enum(ref data) => data.variants.iter().flat_map(|variant| {
+			match &variant.fields {
+				| Fields::Named(FieldsNamed { named: fields , .. })
+				| Fields::Unnamed(FieldsUnnamed { unnamed: fields, .. }) => {
+					fields.iter().map(|f| f.ty.clone()).collect()
+				},
+
+				Fields::Unit => { vec![] },
```
bkchr:
```suggestion
				Fields::Unit => { Vec::new() },
```

derive/src/lib.rs
```diff
@@ -108,11 +118,49 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+fn add_trait_bounds(generics: &mut Generics, data: &syn::Data, bound: syn::Path) -> Result<(), syn::Error> {
+	if generics.params.is_empty() {
+		return Ok(());
+	}
+
+	let types = collect_types(&data)?;
+	if !types.is_empty() {
+		let where_clause = generics.make_where_clause();
+
+		for ty in types {
```
bkchr:
```suggestion
		types.into_inter().for_each(|ty| where_clause.predicates.push(parse_quote!(#ty : #bound)));
```

derive/src/lib.rs
```diff
@@ -108,11 +118,49 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+			where_clause.predicates.push(parse_quote!(#ty : #bound));
```
bkchr:
```suggestion
```

derive/src/lib.rs
```diff
@@ -108,11 +118,49 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
 		}
```
bkchr:
```suggestion
```

derive/src/lib.rs
```diff
@@ -108,11 +118,47 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+fn add_trait_bounds(generics: &mut Generics, data: &syn::Data, bound: syn::Path) -> Result<(), syn::Error> {
+	if generics.params.is_empty() {
+		return Ok(());
+	}
+
+	let types = collect_types(&data)?;
+	if !types.is_empty() {
+		let where_clause = generics.make_where_clause();
+
+		types.into_inter().for_each(|ty| where_clause.predicates.push(parse_quote!(#ty : #bound)));
```
bkchr:
```suggestion
		types.into_iter().for_each(|ty| where_clause.predicates.push(parse_quote!(#ty : #bound)));
```

Sorry! My mistake!

PR #41, opened 2019-01-25
improve automatic bound

thiolliere:
related to https://github.com/paritytech/parity-codec/issues/38

EDIT: delete old way to do it

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -118,29 +118,50 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
 	generated.into()
 }
 
-fn add_trait_bounds(generics: &mut Generics, data: &syn::Data, bound: syn::Path) -> Result<(), syn::Error> {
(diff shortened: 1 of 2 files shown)


thiolliere:
Note: On the implementation part we could do less effort.

Serde try to be efficient with `associated_type_usage: Vec<&'ast syn::TypePath>` so when a type T::Type is used it expand in where T::Type: Bound instead of T: Bound.
This is not perfect as explained in the first comment but can be useful.

So we could have a simpler implementation without this feature.

Note: Also it has been tested with new test and on substrate

thiolliere:
OK also related to https://github.com/paritytech/parity-codec/pull/40 I have to take a look

thiolliere:
So this PR has been rework to use the new bound mecanism.

What is now does is quite straightforward instead of bounding Codec it filter out encoded_as and compact fields when bonding for codec.
And also it automatically bound HasCompact for compact field.

It has been tested with substrate

PR #42, opened 2019-02-06
compact using_encoded no vec allocation

thiolliere:
on using_encoded method compact and compactref use now an arrayvec instead of calling encode which allocate a vec.

(Another way to do that is to implement using_encoded instead of encode_to which would allow to not allocate the full array in some cases, but then encode_to might be a bit less efficient)

also as the code is the same for each impl I could have make use a macro. But I prefer not and think it is OK this way.

The test ensure the arrayvec size is fine.

related to https://github.com/paritytech/parity-codec/issues/29

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -109,6 +109,20 @@ impl<W: ::std::io::Write> Output for W {
 	}
 }
 
+struct ArrayVecWrapper<T: arrayvec::Array>(ArrayVec<T>);
+
+impl<T: arrayvec::Array<Item=u8>> Output for ArrayVecWrapper<T> {
+	fn write(&mut self, bytes: &[u8]) {
+		for byte in bytes {
+			self.push_byte(*byte);
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


src/codec.rs
```diff
@@ -109,6 +109,20 @@ impl<W: ::std::io::Write> Output for W {
 	}
 }
 
+struct ArrayVecWrapper<T: arrayvec::Array>(ArrayVec<T>);
+
+impl<T: arrayvec::Array<Item=u8>> Output for ArrayVecWrapper<T> {
```
bkchr: Why do we need this wrapper? Can we not directly implement `Output` on `ArrayVec`?

thiolliere: no because it will conflict with `impl<W: ::std::io::Write> Output for W` if ArrayVec implement Write in the future

thiolliere:
primitive types already implement using_encoded without allocation

PR #46, opened 2019-02-12
Switch to rust `2018` and use `proc-macro-crate`

bkchr:


src/lib.rs
```diff
@@ -23,13 +23,11 @@
...
-#[cfg(feature = "std")]
-extern crate core;
 
-#[cfg(feature = "std")]
-extern crate serde;
 
-extern crate arrayvec;
+
+
+
```
thiolliere: quite some space no ? but whatever actually

src/codec.rs
```diff
@@ -56,15 +56,15 @@ impl<'a> Input for &'a [u8] {
 #[cfg(feature = "std")]
 impl<R: ::std::io::Read> Input for R {
 	fn read(&mut self, into: &mut [u8]) -> usize {
-		match (self as &mut ::std::io::Read).read_exact(into) {
+		match (self as &mut dyn (::std::io::Read)).read_exact(into) {
```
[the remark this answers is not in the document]
bkchr: I don't know :D That was `cargo fix` :D

src/codec.rs
```diff
@@ -279,10 +279,10 @@ impl<T: 'static> HasCompact for T where
 
 // compact encoding:
 // 0b00 00 00 00 / 00 00 00 00 / 00 00 00 00 / 00 00 00 00
-//   xx xx xx 00															(0 ... 2**6 - 1)		(u8)
-//   yL yL yL 01 / yH yH yH yL												(2**6 ... 2**14 - 1)	(u8, u16)  low LH high
-//   zL zL zL 10 / zM zM zM zL / zM zM zM zM / zH zH zH zM					(2**14 ... 2**30 - 1)	(u16, u32)  low LMMH high
-//   nn nn nn 11 [ / zz zz zz zz ]{4 + n}									(2**30 ... 2**536 - 1)	(u32, u64, u128, U256, U512, U520) straight LE-encoded
+//   xx xx xx 00															(0 ... 2**6)		(u8)
```
thiolliere: then I would write ..= to be more explicit with new syntax

thiolliere: oups would be `..` now isn't it

thiolliere: oh it doesn't seems to be "up to equal not included" syntax ? in pattern matching

bkchr: `...` is different than `..=`. See https://doc.rust-lang.org/reference/expressions/range-expr.html

thiolliere: Yes but .. is different from ..=
but for pattern matching it is a range-pattern: and ... is equals to ..=

https://doc.rust-lang.org/reference/patterns.html#range-patterns

https://play.integer32.com/?version=stable&mode=debug&edition=2018

bkchr: Ahh! Did not thought about that and I also did not know about this change. But the comment is referring to the range and does not use pattern matching. At least in my opinion. Thus, I would not change it.

thiolliere: But then I would write `..` to be as rust range no ?

bkchr: Yes. Wtf. My brain is completely fucked up.^^

PR #49, opened 2019-02-28
Support recursive types

bkchr:


derive/src/trait_bounds.rs
```diff
@@ -0,0 +1,200 @@
...
+
+				Fields::Unit => { Vec::new() },
+			}
+		}).collect(),
+
+		Data::Union(_) => return Err(Error::new(Span::call_site(), "Union types are not supported.")),
+	};
+
+	Ok(types)
+}
```
0x7CFE: newline missing

tests/mod.rs
```diff
@@ -398,3 +395,27 @@ fn generic_trait() {
...
+fn recursive_variant_2_encode_works() {
+	#[derive(Debug, PartialEq, Encode, Decode, Default)]
+	struct Recursive<A, B, N> {
+		data: N,
+		other: Vec<Struct<A, B, Recursive<A, B, N>>>,
+	}
+
+	let val: Recursive<u32, i32, u32> = Recursive::default();
+	val.encode();
+}
```
0x7CFE: newline missing

derive/src/lib.rs
```diff
@@ -139,73 +148,4 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
-				Fields::Unit => { Vec::new() },
-			}
-		}).collect(),
-
-		Data::Union(_) => return Err(Error::new(Span::call_site(), "Union types are not supported.")),
-	};
-
-	Ok(types)
-}
+}
```
0x7CFE: newline missing

bkchr:
In the case of an alias we can not do anything... The user will need to switch to a custom implementation, but I think that this is a really really rare case.

PR #50, opened 2019-02-28
Release 3.1.0

bkchr:


thiolliere:
I thought #40 was a breaking change. but in very rare cases: https://github.com/paritytech/parity-codec/pull/40#issuecomment-460980518

thiolliere:
oh I'm wrong this release only codec not derive. So what we want is just to include BtreeMap and using_encoded improvment. If so I approve

bkchr:
Hmm it seems to be complicated. I saw my comment in the pr and was also arguing that this is a breaking change. Now, I'm not sure anymore.
So, we don't break anything in the public API of `parity-codec`, it could only happen that something "break" in the API of a project using `parity-codec` if they refactor code as given in the rustc issue.

So, I think we don't break the `parity-codec` api with this change.
I checked polkadot and substrate and both compile without any errors. (one warning in substrate, again a rustc bug..., but I fixed this in substrate)
And in the case of the new derive, I would say that we lower the required trait bounds.

thiolliere:
it seems to me they don't have to refactor to if they already rely on the wrong bound for some use.

But actually I'm not sure how any situation of it could exist. I would go for approve then.

PR #52, opened 2019-03-06
decode returns Result instead of Option

### LINKED ISSUE #33: Decode::decode should return a Result instead of an Option
So that we can know why we failed to decode something.

montekki:
A very WIP attempt to fix  #33

The main idea is to propagate infomative errors from ```Decode::decode```, for instance the following code:

```rust
extern crate parity_codec;
extern crate parity_codec_derive;

use parity_codec::Decode;
use parity_codec_derive::{Encode, Decode};

#[derive(Encode, Decode, Debug)]
struct Test2(u32);

#[derive(Encode, Decode, Debug)]
struct Test{
        a: u32,
        c: Test2,
}

#[derive(Encode, Decode, Debug)]
enum TestEnum {
        Var1,
        Var2(u32),
}

fn main() {
        let mut wrong: &[u8] = b"\x0f";

        match Test::decode(&mut wrong) {
                Ok(_) => (),
                Err(a) => println!("error {}", a),
        }
        match Test2::decode(&mut wrong) {
                Ok(_) => (),
                Err(a) => println!("error {}", a),
        }
        match TestEnum::decode(&mut wrong) {
                Err(a) => println!("error {}", a),
                _ => (),
        }
}

```

yeilds the following messages on

```
error Error decoding field Test.a
error Error decoding field Test2.0
error No such variant in enum TestEnum

```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ parity-codec-derive = { path = "derive", version = "3.1", default-features = fal
 
 [dev-dependencies]
 serde_derive = { version = "1.0" }
-parity-codec-derive = { path = "derive", version = "3.1", default-features = false }
+parity-codec-derive = { path = "derive", version = "3.1" }
 
 [features]
 default = ["std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ parity-codec-derive = { path = "derive", version = "3.1", default-features = fal
 
 [dev-dependencies]
 serde_derive = { version = "1.0" }
-parity-codec-derive = { path = "derive", version = "3.1" }
+parity-codec-derive = { path = "derive", version = "3.1", default-features = false }
 
 [features]
 default = ["std"]
--- a/derive/src/decode.rs
+++ b/derive/src/decode.rs
@@ -63,7 +63,7 @@ pub fn quote(data: &Data, type_name: &Ident, input: &TokenStream) -> TokenStream
 			quote! {
 				match #input.read_byte()? {
 					#( #recurse )*
-					x => Err(_parity_codec::Error::new(#err_msg)),
+					x => Err(_parity_codec::Error::from(#err_msg)),
 				}
 			}
 
@@ -83,7 +83,10 @@ fn create_decode_expr(field: &Field, name: &String, input: &TokenStream) -> Toke
 		).to_compile_error();
 	}
 
-	let err_msg = format!("Error decoding field {}", name);;
+	#[cfg(feature = "std")]
+	let err_msg = format!("Error decoding field {}", name);
+	#[cfg(not(feature = "std"))]
+	let err_msg = { &name[..0] };
 
 	if compact {
 		let field_type = &field.ty;
@@ -91,7 +94,7 @@ fn create_decode_expr(field: &Field, name: &String, input: &TokenStream) -> Toke
 			{
 				let res = <<#field_type as _parity_codec::HasCompact>::Type as _parity_codec::Decode>::decode(#input);
 				match res {
-					Err(_) => return Err(_parity_codec::Error::new(#err_msg)),
+					Err(_) => return Err(_parity_codec::Error::from(#err_msg)),
 					Ok(a) => a.into(),
 				}
 			}
@@ -101,7 +104,7 @@ fn create_decode_expr(field: &Field, name: &String, input: &TokenStream) -> Toke
 			{
 				let res = <#encoded_as as _parity_codec::Decode>::decode(#input);
 				match res {
-					Err(_) => return Err(_parity_codec::Error::new(#err_msg)),
+					Err(_) => return Err(_parity_codec::Error::from(#err_msg)),
 					Ok(a) => a.into(),
 				}
 			}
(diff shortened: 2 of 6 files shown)


src/codec.rs
```diff
@@ -16,51 +16,73 @@
...
 	borrow::{Cow, ToOwned},
 };
 
 use core::{mem, slice};
 use arrayvec::ArrayVec;
 use core::marker::PhantomData;
 
+#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(PartialEq)]
+pub struct Error(pub String);
```
bkchr: Using a `String` as internal representation is maybe not the best idea. On `no_std` we will get into trouble with this.
Could we switch to `&'static str`?

montekki: yeah, updated the PR

derive/src/decode.rs
```diff
@@ -62,7 +62,7 @@ pub fn quote(data: &Data, type_name: &Ident, input: &TokenStream) -> TokenStream
 			quote! {
 				match #input.read_byte()? {
 					#( #recurse )*
-					x => Err(_parity_codec::Error::new("No such variant in enum ".to_string() + &x.to_string())),
+					x => Err(_parity_codec::Error::new("No such variant in enum")),
```
bkchr: Maybe we should add the name of the enum at least, to make debugging easier.

derive/src/decode.rs
```diff
@@ -87,9 +87,9 @@ fn create_decode_expr(field: &Field, name: &proc_macro2::TokenStream, input: &To
 		quote_spanned! { field.span() =>
 			{
 				let res = <<#field_type as _parity_codec::HasCompact>::Type as _parity_codec::Decode>::decode(#input);
-				let fname = stringify!(#name).replace(" ", "");
+				let msg = stringify!(Error decoding field #name);
```
bkchr: Please put the msg above the if and do something like `let msg = format!("Error decoding field: {}", name);` and then just use `#msg` in `quote!`

derive/src/decode.rs
```diff
@@ -98,9 +98,9 @@ fn create_decode_expr(field: &Field, name: &proc_macro2::TokenStream, input: &To
 		quote_spanned! { field.span() =>
 			{
 				let res = <#encoded_as as _parity_codec::Decode>::decode(#input);
-				let fname = stringify!(#name).replace(" ", "");
+				let msg = stringify!(Error decoding field #name);
 				match res {
-					Err(e) => return Err(_parity_codec::Error::new("Error decoding field ".to_string() + &fname + ": " + &e.0)),
```
bkchr: You can reuse the same msg from above.

derive/src/decode.rs
```diff
@@ -109,9 +109,9 @@ fn create_decode_expr(field: &Field, name: &proc_macro2::TokenStream, input: &To
 		quote_spanned! { field.span() =>
 			{
 				let res = _parity_codec::Decode::decode(#input);
-				let fname = stringify!(#name).replace(" ", "");
+				let msg = stringify!(Error decoding field #name);
```
bkchr: Same.

src/codec.rs
```diff
@@ -31,17 +30,22 @@ use core::marker::PhantomData;
...
 		Error(s)
 	}
+
+	pub fn what(&self) -> &'static str {
+		self.0
+	}
 }
 
+#[cfg(any(feature = "std", feature = "full"))]
 impl ToString for Error {
```
bkchr: We don't need this implementation of `ToString`.

src/codec.rs
```diff
@@ -31,17 +30,22 @@ use core::marker::PhantomData;
 
 #[cfg_attr(feature = "std", derive(Debug))]
 #[derive(PartialEq)]
-pub struct Error(pub String);
+pub struct Error(&'static str);
```
bkchr: Please implement `std::error::Error` and `std::fmt::Display` as well.

src/codec.rs
```diff
@@ -28,39 +28,75 @@ use core::{mem, slice};
...
 use core::marker::PhantomData;
 
+#[cfg(feature = "std")]
+use std::fmt;
+
+#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(PartialEq)]
+pub struct Error(&'static str);
+
+impl Error {
```
bkchr: Could you implement `From<&'static str>` for `Error` please.

montekki: ok done

Cargo.toml
```diff
@@ -15,7 +15,7 @@ parity-codec-derive = { path = "derive", version = "3.1", default-features = fal
 
 [dev-dependencies]
 serde_derive = { version = "1.0" }
-parity-codec-derive = { path = "derive", version = "3.1", default-features = false }
+parity-codec-derive = { path = "derive", version = "3.1" }
```
bkchr: This needs to be reverted.

derive/src/decode.rs
```diff
@@ -83,17 +83,42 @@ fn create_decode_expr(field: &Field, input: &TokenStream) -> TokenStream {
 		).to_compile_error();
 	}
 
+	#[cfg(feature = "std")]
+	let err_msg = format!("Error decoding field {}", name);
```
bkchr: You don't need to distinguish here.

montekki: I am afraid that if we don't explicitly trim these strings, they will not be elimintad from the resulting code by DCE

PR #54, opened 2019-03-11
forward encoding for single field struct

thiolliere:
Fix https://github.com/paritytech/parity-codec/issues/43

### DIFF
--- a/derive/src/encode.rs
+++ b/derive/src/encode.rs
@@ -260,9 +260,10 @@ fn impl_encode(data: &Data, type_name: &Ident) -> TokenStream {
 
 pub fn quote(data: &Data, type_name: &Ident) -> TokenStream {
 	if let Some(implementation) = try_impl_encode_single_field_optimisation(data) {
-		return implementation;
+		implementation
+	} else {
+		impl_encode(data, type_name)
 	}
-	impl_encode(data, type_name)
 }
 
 pub fn stringify(id: u8) -> [u8; 2] {
--- a/derive/src/encode.rs
+++ b/derive/src/encode.rs
@@ -119,10 +119,7 @@ fn encode_fields<F>(
 	}
 }
 
-pub fn quote(data: &Data, type_name: &Ident) -> TokenStream {
-	let call_site = Span::call_site();
-
-	// optimisation for single field struct
+fn try_impl_encode_single_field_optimisation(data: &Data) -> Option<TokenStream> {
 	let ref closure = quote!(f);
 	let optimisation = match *data {
 		Data::Struct(ref data) => {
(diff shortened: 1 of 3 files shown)


derive/src/encode.rs
```diff
@@ -79,9 +119,51 @@ fn encode_fields<F>(
...
+		return quote! {
+			fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, #closure: F) -> R {
+				#optimisation
+			}
+		};
+	}
+
+	let self_ = quote!(self);
+	let ref dest = quote!(dest);
+	let encoding = match *data {
```
bkchr: Maybe we could move the code below into an own function as well.

derive/src/encode.rs
```diff
@@ -79,9 +119,51 @@ fn encode_fields<F>(
...
+pub fn quote(data: &Data, type_name: &Ident) -> TokenStream {
 	let call_site = Span::call_site();
-	match *data {
+
+	// optimisation for single field struct
+	let ref closure = quote!(f);
+	let optimisation = match *data {
+		Data::Struct(ref data) => {
+			match data.fields {
+				Fields::Named(ref fields) => if fields.named.len() == 1 {
```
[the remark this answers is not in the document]
thiolliere: this is so cool!

thiolliere:
seems good to me now, I cleaned quote_spanned in older code as well.

derive/src/encode.rs
```diff
@@ -177,8 +249,22 @@ pub fn quote(data: &Data, type_name: &Ident, self_: &TokenStream, dest: &TokenSt
...
+	quote! {
+		fn encode_to<EncOut: _parity_codec::Output>(&#self_, #dest: &mut EncOut) {
+			#encoding
+		}
 	}
 }
+
+pub fn quote(data: &Data, type_name: &Ident) -> TokenStream {
+	if let Some(implementation) = try_impl_encode_single_field_optimisation(data) {
+		return implementation;
```
bkchr:
```suggestion
		implementation
```

derive/src/encode.rs
```diff
@@ -177,8 +249,22 @@ pub fn quote(data: &Data, type_name: &Ident, self_: &TokenStream, dest: &TokenSt
...
+	}
```
bkchr:
```suggestion
	} else {
```

derive/src/encode.rs
```diff
@@ -177,8 +249,22 @@ pub fn quote(data: &Data, type_name: &Ident, self_: &TokenStream, dest: &TokenSt
...
+	impl_encode(data, type_name)
```
bkchr:
```suggestion
		impl_encode(data, type_name)
```

derive/src/encode.rs
```diff
@@ -177,8 +249,22 @@ pub fn quote(data: &Data, type_name: &Ident, self_: &TokenStream, dest: &TokenSt
...
+}
```
bkchr:
```suggestion
	}
}
```

Issue #47, opened 2019-02-21
Decode is not implemented for `u8`


xlc:
What is the reason `u8` is not here? even `i8` is included.
This means `Decode` is not implemented for `[u8]` as well

https://github.com/paritytech/parity-codec/blob/50ea4fc9c74a426720ff62ed166c9ced485cac89/src/codec.rs#L1006-L1007

Original issue: https://github.com/paritytech/substrate/pull/1804#issuecomment-465741529

gavofyork:
Because it would otherwise make the two encodings: `Vec<u8>` and `Vec<T: Codec>` clash.

xlc:
In that case can we have `Decode` implemented for `[u8]`? There is `Encode` for `[u8]`, it doesn't make much sense to not have `Decode`

bkchr:
You can not implement `Decode` for `[u8]`. How do you want to decode to a slice?

PR #56, opened 2019-03-13
Improve codec implement for u8 and array.

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

## Changes

Implement `Encode` and `Decode` for `u8`;
Remove `impl Encode for [u8]`, `impl Encode for Vec<u8>` and `impl Decode for Vec<u8>`.

## Details 

Why not add `Encode` and `Decode` for `u8`?  In #47, Gavin said it would make the two encodings: `Vec<u8>` and `Vec<T: Codec>` clash.

But we already have these:

- `impl<T: Encode> Encode for [T]`
- `impl<T: Encode> Encode for Vec<T>`
- `impl<T: Decode> Decode for Vec<T>`

If we add `Encode` and `Decode` implementations for `u8`, we won't need `impl Encode for [u8]`, `impl Encode for Vec<u8>` and `impl Decode for Vec<u8>`. 

I think this will be clearer. Maybe I got something wrong?

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -768,32 +768,6 @@ impl<T: Decode> Decode for Box<T> {
 	}
 }
 
-impl Encode for [u8] {
-	fn encode_to<W: Output>(&self, dest: &mut W) {
-		let len = self.len();
-		assert!(len <= u32::max_value() as usize, "Attempted to serialize a collection with too many elements.");
-		Compact(len as u32).encode_to(dest);
-		dest.write(self)
-	}
-}
-
-impl Encode for Vec<u8> {
-	fn encode_to<W: Output>(&self, dest: &mut W) {
-		self.as_slice().encode_to(dest)
-	}
-}
-
-impl Decode for Vec<u8> {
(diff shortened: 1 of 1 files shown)


thiolliere:
The implementation differs, for `[T]` the function write len in a compact way and then all item using:
```rust
		for item in self {
			item.encode_to(dest);
		}
```
on the other hand the implementation of `[u8]` write len in a compact way and then call
```rust
		dest.write(self)
```

there may be a performance issue there, also maybe it could be rewrite with a condition if T is typeid of u8 but at first at least implementation doesn't seems equivalent

koushiro:
@thiolliere Thanks for your reply. You are right, there is indeed a performance issue after the benchmark, and the performance is about 7-8 times slower on my computer. I will close the PR.

BTW, what do you think about adding more u8 array types in `impl_non_endians!(...)`, like in `impl_array!(...)`. I think this is more consistent for encoding and decoding array types.

bkchr:
I think it the array types are enough as you can also just cast an array to a slice and encode this.

As you wanted to close your pr anyway, I will do it for you.

PR #58, opened 2019-03-19
impl skip fields and variants

thiolliere:


thiolliere:
not sure you've seen it actually @bkchr

bkchr:
I had seen and forget it, thanks for the reminder. WIll take a look.

derive/src/trait_bounds.rs
```diff
@@ -149,29 +160,49 @@ pub fn add(
...
+) -> syn::Result<Vec<syn::Type>> {
 	use syn::*;
 
 	let types = match *data {
 		Data::Struct(ref data) => match &data.fields {
 			| Fields::Named(FieldsNamed { named: fields , .. })
 			| Fields::Unnamed(FieldsUnnamed { unnamed: fields, .. }) => {
 				fields.iter()
-					.filter(|f| type_filter(f))
+					.filter(|f| type_filter(f, false))
```
bkchr: Why don't we check for `skip` here?

derive/src/trait_bounds.rs
```diff
@@ -149,29 +160,49 @@ pub fn add(
...
-	crate::utils::get_enable_compact(field)
+fn needs_has_compact_bound(field: &syn::Field, variant_skip: bool) -> bool {
+	!variant_skip
+		&& crate::utils::get_enable_compact(field)
 }
 
-fn collect_types(data: &syn::Data, type_filter: fn(&syn::Field) -> bool) -> syn::Result<Vec<syn::Type>> {
+fn needs_default_bound(field: &syn::Field, variant_skip: bool) -> bool {
+	// types from skipped variants doesn't need default
+	!variant_skip
```
bkchr: I don't understand why each function takes `variant_skip` and does the same with it? Can not just `collect_types` do this?

thiolliere: Yes true

thiolliere:
issues addressed

tests/skip.rs
```diff
@@ -0,0 +1,61 @@
...
+	#[derive(PartialEq, Debug, Default)]
+	struct UncodecType;
+
+	#[derive(PartialEq, Debug)]
+	struct UncodecUndefaultType;
+
+	#[derive(PartialEq, Debug, Encode, Decode)]
+	enum Enum<T=UncodecType, S=UncodecUndefaultType> {
+		#[codec(skip)]
+		A(S),
```
bkchr: Shouldn't this fail to compile when you want to decode it? As `S` is not insatiable?

derive/src/trait_bounds.rs
thiolliere: we should not filter out skip fields here because collect_types is used to get skip fields to bound Default on them in Decode macro

tests/skip.rs
thiolliere: I don't think so, isn't that our usecase with enum Event ? we want some variant to be skipped in the encoded representation imo

bkchr: I don't want it to fail, but as `S` does not implement `Default`, `decode` should not be callable?

thiolliere:
variant filter added

PR #59, opened 2019-03-29
Simplify implementations and optimize some cases

tomusdrw:
1. Use `extend_from_slice` for Vector output (instead of `extend` which requires an iterator).
2. Use `extend` for `ArrayVecOutput` instead of manual loop.
3. Avoid reallocations in `encode()` by providing `size_hint` in `Encode`.
4. Correctly implement `Encode` for wrapper/pointer types by delegating ALL methods - otherwise some of the methods could use default, non-optimized implementations. Introduce `WrapperTypeEncode` trait for simplified implementation of this for `Deref` types.
5. Remove manual imlementations of `Encode` for `Compact`, rather delegate everything to `CompactRef`

I've added some benchmarks to make sure that nothing is seriously broken, but I think it would be more interesting to see if and how this affects the size/cost of generated runtime in substrate. 
I'll chat with @pepyakin on how to measure that.


src/codec.rs
```diff
@@ -901,30 +939,9 @@ impl<K: Decode + Ord, V: Decode> Decode for BTreeMap<K, V> {
...
 
 impl Encode for () {
-	fn encode_to<T: Output>(&self, _dest: &mut T) {
-	}
-
 	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
 		f(&[])
 	}
-
-	fn encode(&self) -> Vec<u8> {
```
bkchr: Why did you remove this?

src/codec.rs
```diff
@@ -901,30 +939,9 @@ impl<K: Decode + Ord, V: Decode> Decode for BTreeMap<K, V> {
 }
 
 impl Encode for () {
-	fn encode_to<T: Output>(&self, _dest: &mut T) {
```
bkchr: And this?

src/codec.rs
```diff
@@ -994,8 +1032,9 @@ mod inner_tuple_impl {
 
 /// Trait to allow conversion to a know endian representation when sensitive.
 /// Types implementing this trait must have a size > 0.
-// note: the copy bound and static lifetimes are necessary for safety of `Codec` blanket
-// implementation.
+///
+/// note: the copy bound and static lifetimes are necessary for safety of `Codec` blanket
```
bkchr: I would make `Note` a heading.

src/lib.rs
```diff
@@ -37,6 +39,7 @@ extern crate serde_derive;
 pub use parity_codec_derive::*;
 
 #[cfg(feature = "std")]
+/// Alloc shim for std environment.
```
bkchr: Should probably be `#[doc(hidden)]`.

src/codec.rs
```diff
@@ -170,9 +183,7 @@ struct ArrayVecWrapper<T: arrayvec::Array>(ArrayVec<T>);
 
 impl<T: arrayvec::Array<Item=u8>> Output for ArrayVecWrapper<T> {
 	fn write(&mut self, bytes: &[u8]) {
-		for byte in bytes {
-			self.push_byte(*byte);
-		}
+		self.0.extend(bytes.iter().map(|i| *i));
```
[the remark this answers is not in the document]
tomusdrw: That would actually require using `unsafe set_len` first, cause the slice that `ArrayVec` derefs to only contains initialised elements.

src/codec.rs
```diff
@@ -756,19 +822,11 @@ impl_array!(
...
-		self.as_ref().encode_to(dest)
-	}
-}
-
-impl<T: Decode> Decode for Box<T> {
-	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-		Ok(Box::new(T::decode(input)?))
+impl Encode for [u8] {
+	fn size_hint(&self) -> usize {
+		self.len() + mem::size_of::<u32>()
```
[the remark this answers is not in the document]
tomusdrw: That won't really work, since `size_hint` is not implemented for `Compact/CompactRef`.
I've only implemented it for objects that can potentially contain large (kBytes, MBytes) values that would benefit from this optimization to avoid cluttering the code.
Note it's also perfectly fine to overestimate a bit, the actual allocation is likely to be larger than this value anyway.

src/codec.rs
tomusdrw: I'm pretty sure it's going to be optimized out by the compiler anyway and just clutters the code. But I'm ok to bring it back if you think it's worth having.

tomusdrw:
@thiolliere Added a test for a type that implements `WrapperType*`.

thiolliere:
> @thiolliere Added a test for a type that implements `WrapperType*`.

I was thinking create a type implement deref and wrappertype on it. and test it works.
This is mostly for change on the trait to be easily caught

src/codec.rs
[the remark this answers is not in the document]
bkchr: Yeah, makes sense :)

src/codec.rs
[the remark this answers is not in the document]
tomusdrw: I've checked the internal implementation and it's way different. with `push_byte` the length is updated multiple times, while with `extend` it's only updated once. However I agree that a memcpy would be even better.
Maybe we can a safe alternative to `copy_from_slice` for `ArrayVec`? I'll look into how feasible that is and for now change to unsafe.

src/codec.rs
bkchr: I would really like to see to get `Vec::new` back. I'm not 100% sure that the compiler will optimize this away. Could be, but involves some function calls that he would need to optimize into a single `Vec::new()`.

src/codec.rs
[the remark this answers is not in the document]
bkchr: As we had the discussion yesterday, using `unsafe` is no direct code smell. If you add the required length checks, it should be okay :)

PR #65, opened 2019-04-02
Implement empty enum derive

bkchr:


bkchr:
Master is done, `codec` bump is a good point.

PR #67, opened 2019-04-05
Avoid some malleability in compact integers

cheme:
Avoids multiple possible encoding for a same integer value.
This might not be required in codec but as it is use by trie (compact<u32>) I would rather have it.


src/codec.rs
```diff
@@ -1498,4 +1595,14 @@ mod tests {
 		assert_eq!(MyWrapper(3u32.into()).encode(), result);
 		assert_eq!(MyWrapper::decode(&mut &*result).unwrap(), MyWrapper(3_u32.into()));
 	}
+
+	#[test]
+	fn malleability() {
```
bkchr: I would like to have a more descriptive name of this test.

src/codec.rs
```diff
@@ -1498,4 +1595,14 @@ mod tests {
...
+		let enc = ((5u16 << 2) | 0b01).encode();
+		assert!(Compact::<u16>::decode(&mut & enc[..]).is_err());
```
bkchr: Please check for the exact error message and not just if an error occurred.

And it would be nice to test the other `Compact` impls as well.

src/codec.rs
```diff
@@ -625,15 +625,22 @@ impl Decode for Compact<u16> {
 		let prefix = input.read_byte()?;
 		Ok(Compact(match prefix % 4 {
 			0 => prefix as u16 >> 2,
-			1 => u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? as u16 >> 2,
+			1 => {
+				let x = u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
+				if x > 0b00111111 && x < (0b00111111_11111111 + 1) {
```
```suggestion
				if x > 0b00111111 && x <= 0b00111111_11111111 {
```

src/codec.rs
```diff
@@ -644,12 +651,31 @@ impl Decode for Compact<u32> {
 		let prefix = input.read_byte()?;
 		Ok(Compact(match prefix % 4 {
 			0 => prefix as u32 >> 2,
-			1 => u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? as u32 >> 2,
-			2 => u32::decode(&mut PrefixInput{prefix: Some(prefix), input})? as u32 >> 2,
+			1 => {
+				let x = u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
+				if x > 0b00111111 && x < (0b00111111_11111111 + 1) {
```
```suggestion
				if x > 0b00111111 && x <= 0b00111111_11111111 {
```

src/codec.rs
```diff
@@ -644,12 +651,31 @@ impl Decode for Compact<u32> {
...
+					x as u32
+				} else {
+					return Err("out of range decoding Compact<u32>".into());
+				}
+			},
+			2 => {
+				let x = u32::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
+				if x > 0b00111111_11111111 && x < (u32::max_value() >> 2) + 1 {
```
```suggestion
				if x > 0b00111111_11111111 && x <= (u32::max_value() >> 2) {
```

src/codec.rs
```diff
@@ -664,19 +690,51 @@ impl Decode for Compact<u64> {
 		let prefix = input.read_byte()?;
 		Ok(Compact(match prefix % 4 {
 			0 => prefix as u64 >> 2,
-			1 => u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? as u64 >> 2,
-			2 => u32::decode(&mut PrefixInput{prefix: Some(prefix), input})? as u64 >> 2,
+			1 => {
+				let x = u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
+				if x > 0b00111111 && x < (0b00111111_11111111 + 1) {
```
```suggestion
				if x > 0b00111111 && x <= 0b00111111_11111111 {
```

src/codec.rs
```diff
@@ -664,19 +690,51 @@ impl Decode for Compact<u64> {
...
+					x as u64
+				} else {
+					return Err("out of range decoding Compact<u64>".into());
+				}
+			},
+			2 => {
+				let x = u32::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
+				if x > 0b00111111_11111111 && x < (u32::max_value() >> 2) + 1 {
```
```suggestion
				if x > 0b00111111_11111111 && x <= u32::max_value() >> 2 {
```

src/codec.rs
```diff
@@ -687,21 +745,60 @@ impl Decode for Compact<u128> {
...
+				let x = u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
+				if x > 0b00111111 && x < (0b00111111_11111111 + 1) {
+					x as u128
+				} else {
+					return Err("out of range decoding Compact<u128>".into());
+				}
+			},
+			2 => {
+				let x = u32::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
+				if x > 0b00111111_11111111 && x < (u32::max_value() >> 2) + 1 {
```
```suggestion
				if x > 0b00111111_11111111 && x <= u32::max_value() >> 2 {
```

src/codec.rs
```diff
@@ -687,21 +745,60 @@ impl Decode for Compact<u128> {
 		let prefix = input.read_byte()?;
 		Ok(Compact(match prefix % 4 {
 			0 => prefix as u128 >> 2,
-			1 => u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? as u128 >> 2,
-			2 => u32::decode(&mut PrefixInput{prefix: Some(prefix), input})? as u128 >> 2,
+			1 => {
+				let x = u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
+				if x > 0b00111111 && x < (0b00111111_11111111 + 1) {
```
```suggestion
				if x > 0b00111111 && x <= 0b00111111_11111111 {
```

src/codec.rs
```diff
@@ -609,12 +609,12 @@ impl Decode for Compact<u8> {
 			0 => prefix as u8 >> 2,
 			1 => {
 				let x = u16::decode(&mut PrefixInput{prefix: Some(prefix), input})? >> 2;
-				if x < 256 {
+				if x > 0b00111111 && x < 255 + 1 {
```
bkchr: Why always + 1 and not just `<=`?

src/codec.rs
```diff
@@ -1498,4 +1595,61 @@ mod tests {
...
+			)*
+		};
+		( $s:expr, $ty2:ty) => {
+			let mut dest = Vec::new();
+			dest.push(0b11 + (($s - 4) << 2) as u8);
+			for _ in 0..($s - 1) {
+				dest.push(u8::max_value());
+			}
+			dest.push(0);
+			assert!(Compact::<$ty2>::decode(&mut &dest[..]).is_err());
```
bkchr: Please check the exact error message, your test would be successful even if some other error would be returned.

src/codec.rs
cheme: :) that is an old habit, most likely totally useless now ( https://stackoverflow.com/questions/1029992/is-the-inequality-operator-faster-than-the-equality-operator ).

bkchr: Okay, I remember that `> 0` is fast than `0 <`. Or something similar ^^

src/codec.rs
```diff
@@ -664,19 +696,51 @@ impl Decode for Compact<u64> {
...
+					let x = u32::decode(input)?;
+					if x > u32::max_value() >> 2 {
+						x as u64
+					} else {
+						return Err(U64_OUT_OF_RANGE.into());
+					}
+				},
+				8 => {
+					let x = u64::decode(input)?;
+					if x > u64::max_value() >> 8 {
```
bkchr: I think this is incorrect, if we require 8 bytes, we have a value that requires a full `u64`.

src/codec.rs
```diff
@@ -664,19 +696,51 @@ impl Decode for Compact<u64> {
...
+				} else {
+					return Err(U64_OUT_OF_RANGE.into());
+				}
+			},
 			3|_ => match (prefix >> 2) + 4 {
-				4 => u32::decode(input)? as u64,
-				8 => u64::decode(input)?,
+				4 => {
+					let x = u32::decode(input)?;
+					if x > u32::max_value() >> 2 {
```
bkchr: This is also incorrect, we store the "length" information in an extra byte in front of the actual value.

src/codec.rs
bkchr: So the encoded value actually requires 9 bytes.

src/codec.rs
```diff
@@ -687,21 +751,60 @@ impl Decode for Compact<u128> {
...
+					return Err(U128_OUT_OF_RANGE.into());
+				}
+			},
 			3|_ => match (prefix >> 2) + 4 {
-				4 => u32::decode(input)? as u128,
-				8 => u64::decode(input)? as u128,
-				16 => u128::decode(input)?,
+				4 => {
+					let x = u32::decode(input)?;
+					if x > u32::max_value() >> 2 {
```
bkchr: Same here as for `u64`.

src/codec.rs
```diff
@@ -664,19 +696,51 @@ impl Decode for Compact<u64> {
...
+				},
 				x if x > 8 => return Err("unexpected prefix decoding Compact<u64>".into()),
 				bytes_needed => {
 					let mut res = 0;
 					for i in 0..bytes_needed {
 						res |= (input.read_byte()? as u64) << (i * 8);
 					}
-					res
-				}
+					if res > u64::max_value() >> (8 - bytes_needed + 1) * 8 {
```
```suggestion
					if res > u64::max_value() >> (8 - bytes_needed) * 8 {
```

src/codec.rs
```diff
@@ -687,21 +751,60 @@ impl Decode for Compact<u128> {
...
 				x if x > 16 => return Err("unexpected prefix decoding Compact<u128>".into()),
 				bytes_needed => {
 					let mut res = 0;
 					for i in 0..bytes_needed {
 						res |= (input.read_byte()? as u128) << (i * 8);
 					}
-					res
-				}
-			}
+					if res > u128::max_value() >> (16 - bytes_needed + 1) * 8 {
```
```suggestion
					if res > u128::max_value() >> (16 - bytes_needed) * 8 {
```

src/codec.rs
```diff
@@ -1498,4 +1601,62 @@ mod tests {
...
+
+
+	#[test]
+	fn should_avoid_overlapping_definition() {
+		check_bound!(0b01, u8, u16, [ (u8, U8_OUT_OF_RANGE), (u16, U16_OUT_OF_RANGE),
+			(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]);
+		check_bound!(0b10, u16, u32, [ (u16, U16_OUT_OF_RANGE),
+			(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]);
+		check_bound_u32!(
+			[(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]);
```
```suggestion
			[(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]
		);
```

src/codec.rs
```diff
@@ -1498,4 +1601,62 @@ mod tests {
...
+			assert_eq!(Compact::<$ty2>::decode(&mut &dest[..]),
+				Err($ty2_err.into()));
+		};
+	}
+
+
+	#[test]
+	fn should_avoid_overlapping_definition() {
+		check_bound!(0b01, u8, u16, [ (u8, U8_OUT_OF_RANGE), (u16, U16_OUT_OF_RANGE),
+			(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]);
```
```suggestion
			(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]
		);
```

src/codec.rs
```diff
@@ -1498,4 +1601,62 @@ mod tests {
...
+			dest.push(0);
+			assert_eq!(Compact::<$ty2>::decode(&mut &dest[..]),
+				Err($ty2_err.into()));
+		};
+	}
+
+
+	#[test]
+	fn should_avoid_overlapping_definition() {
+		check_bound!(0b01, u8, u16, [ (u8, U8_OUT_OF_RANGE), (u16, U16_OUT_OF_RANGE),
```
```suggestion
		check_bound!(
			0b01, u8, u16, [ (u8, U8_OUT_OF_RANGE), (u16, U16_OUT_OF_RANGE),
```

src/codec.rs
```diff
@@ -1498,4 +1601,62 @@ mod tests {
...
+			(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]);
+		check_bound!(0b10, u16, u32, [ (u16, U16_OUT_OF_RANGE),
+			(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]);
```
```suggestion
			(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]
		);
```

src/codec.rs
```diff
@@ -1498,4 +1601,62 @@ mod tests {
...
+				Err($ty2_err.into()));
+		};
+	}
+
+
+	#[test]
+	fn should_avoid_overlapping_definition() {
+		check_bound!(0b01, u8, u16, [ (u8, U8_OUT_OF_RANGE), (u16, U16_OUT_OF_RANGE),
+			(u32, U32_OUT_OF_RANGE), (u64, U64_OUT_OF_RANGE), (u128, U128_OUT_OF_RANGE)]);
+		check_bound!(0b10, u16, u32, [ (u16, U16_OUT_OF_RANGE),
```
```suggestion
		check_bound!(
			0b10, u16, u32, [ (u16, U16_OUT_OF_RANGE),
```

src/codec.rs
cheme: I am not sure I understand; this just means that on 8 byte an encoding happen iff it could not happen on 7.
So it happens if value is more than what could be encoded on 7 (as in the test) and less than what can be on 8 (we decode from 8 byte).
(when we match 8 here the actual encoding is 9 byte length but the read length from first byte is 8)

cheme: I did ran 
```rust
	#[test]
	fn compact_u64_test() {
		for a in [
			u64::max_value(), 
			u64::max_value() - 1,
			u64::max_value() << 8,
			(u64::max_value() << 8) - 1,
			u64::max_value() << 16,
			(u64::max_value() << 16) - 1,
		].into_iter() {
			let e = Compact::<u64>::encode(&Compact(*a));
			let d = Compact::<u64>::decode(&mut &e[..]).unwrap().0;
			assert_eq!(*a, d);
		}
	}

	#[test]
	fn compact_u128_test() {
		for a in [
			u64::max_value() as u128, 
			(u64::max_value() - 10) as u128, 
			u128::max_value(),
			u128::max_value() - 10,
		].into_iter() {
			let e = Compact::<u128>::encode(&Compact(*a));
			let d = Compact::<u128>::decode(&mut &e[..]).unwrap().0;
			assert_eq!(*a, d);
		}
	}
```
without a pb (but as I say I may not have understood what you meant)

PR #74, opened 2019-04-26
Removes `usize` in impl_endian!

satellitex:
`usize` is not a fixed size type.
So, It can cause dangerous runtime errors. (ref: https://github.com/paritytech/substrate/issues/2357#issuecomment-486939172)

I propose `usize` is not supported `Codec`. It makes checking the above error on Rust type system.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -1136,7 +1136,7 @@ macro_rules! impl_non_endians {
 	)* }
 }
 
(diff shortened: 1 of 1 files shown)


satellitex:
Ok! I changed.
But, it is dependent on many repositories. Should I indicate that there is no `usize` or `isize` Codec?

bkchr:
No, the next release is going to break the interface anyway, so no worries.

PR #75, opened 2019-04-27
Implement Encode/Decode for BTreeSet 

satellitex:
I add Encode/Decode for BTreeSet, because it is not implement.
And `btree_set_workds` checked.


src/codec.rs
```diff
@@ -943,6 +944,30 @@ impl<K: Decode + Ord, V: Decode> Decode for BTreeMap<K, V> {
 	}
 }
 
+impl<T: Encode + Ord> Encode for BTreeSet<T> {
+	fn encode_to<W: Output>(&self, dest: &mut W) {
+		let len = self.len();
+		assert!(len <= u32::max_value() as usize, "Attempted to serialize a collection with too many elements.");
+		(len as u32).encode_to(dest);
```
bkchr: Please use compact

satellitex: I Changed it.

bkchr:
Can you please merge lasted master. We don't use option as result of decode anymore.

satellitex:
@bkchr Many happened conflict. Can I resolve it?

bkchr:
I don't understand your question.

satellitex:
I execute `git merge origin/master` in this branch. 
And many conflicts have happened.

satellitex:
Ah, The problem is that I forked from parity-codec-v3. I will fork from the master and try again.

bkchr:
Please just cherry-pick your two commits into your branch.

bkchr:
@pepyakin I don't understand your comment, the original set already should not contain any duplicates.

cheme:
Not sure (not needed in most case), could be interesting to fail on unordered content when decoding (only allow set to be encoded/decoded as ordered content).

bkchr:
Hmm yeah, but also adds extra complexity/cost to the decoding.

satellitex:
@cheme I think that we should not allow using unordered contents sets.
Fortunately, Btree(Set|Map) meets that specification. But, Hash(Set|Map) is not.

cheme:
@satellitex yes but as @bkchr point out, with current std btreeset implementation I do not know of a costless way to check content is rightly ordered without manually checking the ordering in the decode loop (checking if a value is added twice is natively supported but that will just do half the job of avoiding malleability).
This could be made optional, but it then does add upon code complexity.

thiolliere:
Is unique Encoding/Decoding something we want to enforce ?

It seems to be an interesting property when we use it in the trie: two difference keys cannot be decoded as the same rust structure (or enum).

But I don't know how much it is interesting and important ?

satellitex:
@thiolliere 
It is important to ensure that encoded a set of elements is always the same binary.
As you point out, there may be differences in the trie calculation results at each node.

bkchr:
If someone sees any further issues or have other ideas for improvements, just open an issue :)

@cheme made decoding for the compact unique and the rest should already be unique.

PR #77, opened 2019-05-08
BTree ordering check

cheme:
Related to latest pr about btreeset/btreemap, it fails on encoding that are not build from the encoder (key encoded multiple time or unordered encoding of key).

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -1175,9 +1175,17 @@ impl<K: Decode + Ord, V: Decode> Decode for BTreeMap<K, V> {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 			let mut r: BTreeMap<K, V> = BTreeMap::new();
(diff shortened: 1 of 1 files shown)


cheme:
I don't see it as an attack vector, more an additional guarantee over the library. If someone use the library considering it got this property he can end up comparing hash of big encoded chunk to check for change or any possible scenario.
Currently it is only some additional computation code for a property that may probably not be needed by the user.
This could be isolated into a trait check implementation that is only active if a feature is used, then with proper documentation it could be valuable for parity-codec as a library, but you manage to keep the code small so I am not sure it is suitable (putting this specific code in its own module it may end up not being to ugly).

bkchr:
We have a stable encoding anyway, so you already could hash the output. I still don't see the advantage of this.

bkchr:
The structure will always be correct, as we use the normal `insert` function to build it. If we would do some tricks and move the data into the internal representation, I would understand it.

As long as this is no attack vector, I'm going to close this. If someone thinks so, please reopen with a description of this attack.

PR #81, opened 2019-05-14
Support BitVec, BitBox and BitSlice

DarkEld3r:
Naive implementation to start with. I suppose we also need `BitBox` and `BitSlice` support. Not sure if flexibility to use different [cursor](https://docs.rs/bitvec/0.11.1/bitvec/cursor/index.html) and inner types is needed.

Needed for https://github.com/paritytech/polkadot/issues/254.

### DIFF
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -14,72 +14,38 @@
 
 //! `BitVec` specific serialization.
 
-use bitvec::vec::BitVec;
-
-use std::mem;
+use bitvec::{vec::BitVec, bits::Bits, cursor::Cursor};
+use byte_slice_cast::{AsByteSlice, ToByteSlice, FromByteSlice, Error as FromByteSliceError};
 
 use crate::codec::{Encode, Decode, Input, Output, Compact, Error};
 
-impl Encode for BitVec {
-	fn encode_to<W: Output>(&self, dest: &mut W) {
-		let a1 = vec![1u16, 2, 3];
-		//dest.write(a1.as_slice());
-
-
-
-
-
-		let len = self.len();
-		assert!(len <= u32::max_value() as usize, "Attempted to serialize a collection with too many elements.");
-		Compact(len as u32).encode_to(dest);
-		dest.write(self.as_slice());
+impl From<FromByteSliceError> for Error {
+	fn from(_: FromByteSliceError) -> Error {
+		"failed to cast from byte slice".into()
 	}
 }
 
+impl<C: Cursor, T: Bits + ToByteSlice> Encode for BitVec<C, T> {
+	fn encode_to<W: Output>(&self, dest: &mut W) {
+		let len = self.len();
+		assert!(len <= u32::max_value() as usize, "Attempted to serialize a collection with too many elements.");
+		Compact(len as u32).encode_to(dest);
+		dest.write(self.as_slice().as_byte_slice());
+	}
+}
+
+// TODO: Make implementation generic.
 impl Decode for BitVec {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
-			let element_size = (mem::size_of::<u8>() * 8) as f64;
-			let len_in_bytes = (len as f64 / element_size).ceil() as usize;
-			let mut array = Vec::<u8>::new();
-			for _ in 0..len_in_bytes {
-				//array.push(Compact::<u8>::decode(input)?.0);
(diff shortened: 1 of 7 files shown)


src/codec.rs
```diff
@@ -1227,6 +1230,29 @@ impl Decode for () {
 	}
 }
 
+#[cfg(feature = "bit-vec")]
+impl Encode for BitVec {
+	fn encode_to<W: Output>(&self, dest: &mut W) {
+		let len = self.len();
+		assert!(len <= u32::max_value() as usize, "Attempted to serialize a collection with too many elements.");
+		Compact(len as u32).encode_to(dest);
```
bkchr: Why do we encode the length twice?

src/codec.rs
```diff
@@ -1227,6 +1230,29 @@ impl Decode for () {
...
+}
+
+#[cfg(feature = "bit-vec")]
+impl Decode for BitVec {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
+			let len = len as usize;
+			let mut v = Self::from_vec(Vec::<u8>::decode(input)?);
+			assert!(len <= v.len());
+			unsafe { v.set_len(len); }
```
bkchr: I think the length should already be correctly set?

src/codec.rs
```diff
@@ -1901,4 +1925,27 @@ mod tests {
 			check_bound_high!(i, [(u128, U128_OUT_OF_RANGE)]);
 		}
 	}
+
+	#[cfg(feature = "bit-vec")]
```
bkchr: Please enable the feature in CI while testing.

src/codec.rs
DarkEld3r: I agree that it is somewhat ugly, but these are different lengths. The first one is a number of bits in `BitVec` and the second one is the length of a slice in elements (bytes):
```rust
let vec = bitvec![0, 0];
assert_eq!(2, vec.len());
assert_eq!(1, vec.as_slice().len());
```

bkchr: Ahh okay, but as the length of slices is computable by the number of bits, you should just do 
`dest.write(self.as_slice())` to write without the second length.

DarkEld3r:
Updated and ready for review.

src/bit_vec.rs
```diff
@@ -0,0 +1,194 @@
...
+use std::mem;
+
+use bitvec::{vec::BitVec, bits::Bits, cursor::Cursor, slice::BitSlice, boxed::BitBox};
+use byte_slice_cast::{AsByteSlice, ToByteSlice, FromByteSlice, Error as FromByteSliceError};
+
+use crate::codec::{Encode, Decode, Input, Output, Compact, Error};
+
+impl From<FromByteSliceError> for Error {
+	fn from(_: FromByteSliceError) -> Error {
+		"failed to cast from byte slice".into()
```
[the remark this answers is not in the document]
DarkEld3r: Thanks, fixed.

src/bit_vec.rs
```diff
@@ -0,0 +1,197 @@
...
+
+use bitvec::{vec::BitVec, bits::Bits, cursor::Cursor, slice::BitSlice, boxed::BitBox};
+use byte_slice_cast::{AsByteSlice, ToByteSlice, FromByteSlice, Error as FromByteSliceError};
+
+use crate::codec::{Encode, Decode, Input, Output, Compact, Error};
+
+impl From<FromByteSliceError> for Error {
+	fn from(e: FromByteSliceError) -> Error {
+		match e {
+			FromByteSliceError::AlignmentMismatch{..} => "failed to cast from byte slice: alignment mismatch".into(),
```
thiolliere:
```suggestion
			FromByteSliceError::AlignmentMismatch {..} => "failed to cast from byte slice: alignment mismatch".into(),
```

src/bit_vec.rs
```diff
@@ -0,0 +1,197 @@
...
+			FromByteSliceError::LengthMismatch{..} => "failed to cast from byte slice: length mismatch".into(),
```
thiolliere:
```suggestion
			FromByteSliceError::LengthMismatch {..} => "failed to cast from byte slice: length mismatch".into(),
```

src/bit_vec.rs
DarkEld3r: If only we used `rustfmt`... :upside_down_face:

thiolliere: I think some IDE should be able to run rustfmt on a specific line range. but I coudln't manage to use it in vim :-/

PR #82, opened 2019-05-24
Introduce U8 wrapper encode usable with encoded_as

thiolliere:
this is a wip implementation, that can be finished if ppl think it is interesting.

at the end user will be able to use it as such
```
#[derive(Debug, PartialEq, Encode, Decode)]
struct A {
	#[codec(encoded_as = "U8")]
	pub i: u8,
}
```

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -1428,7 +1428,7 @@ impl From<U8> for u8 {
 }
 
 impl Decode for U8 {
-	fn decode<I: Input>(_value: &mut I) -> Option<Self> {
+	fn decode<I: Input>(_value: &mut I) -> Result<Self, Error> {
 		unimplemented!();
 	}
 }
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -1419,6 +1419,37 @@ impl_non_endians!(
 	[u8; 112], [u8; 128], bool
 );
 
+pub struct U8(u8);
+
+impl From<U8> for u8 {
+	fn from(x: U8) -> Self {
+		x.0
+	}
+}
+
+impl Decode for U8 {
+	fn decode<I: Input>(_value: &mut I) -> Option<Self> {
+		unimplemented!();
+	}
+}
+
+pub struct RefU8<'a>(&'a u8);
+
+impl<'a> From<&'a u8> for RefU8<'a> {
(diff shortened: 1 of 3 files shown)


src/codec.rs
```diff
@@ -1419,6 +1419,37 @@ impl_non_endians!(
...
+
+impl From<U8> for u8 {
+	fn from(x: U8) -> Self {
+		x.0
+	}
+}
+
+impl Decode for U8 {
+	fn decode<I: Input>(_value: &mut I) -> Option<Self> {
+		unimplemented!();
```
Robbepop: This should be easy to implement via `i8` I guess.

tests/mod.rs
```diff
@@ -33,6 +33,12 @@ struct Struct<A, B, C> {
 	pub c: C,
 }
 
+#[derive(Debug, PartialEq, Encode, Decode)]
+struct A {
+	#[codec(encoded_as = "parity_codec::U8")]
```
Robbepop: I think it should be possible to leave away the `"` to make it:
`#[codec(encoded_as = parity_codec::U8)]`
But still it has some problems due to macros not being hygienic as always.

Robbepop:
I just thought about another approach to solve this:

First we need a sealed trait and a special internal `parity_codec` type that implements it:
```rust
mod private {
    pub trait Sealed {}
    pub enum OptIn {}
    impl Sealed for OptInt {}
    impl Sealed for () {}
}
```
Then we extend `parity_codec::Encode` and `parity_codec::Decode` traits by another associated type:
```rust
trait Encode {
    type OptIn: private::Sealed + 'static = ();
    ...
}
```
We also implement `parity_codec::{Encode, Decode}` for `u8` just as we did for all other types with the only difference that we set `OptIn` type to our new `private::OptIn` type.
Then we can easily apply @thiolliere 's proposal using `TypeId` without restricting ourselves to `'static` bounds for `T` since `<T as Encode>::OptIn` is bound to be `'static`.
We are also safe that no other implementations can abuse `OptIn` from the outside since it is sealed and cannot be implemented from external crates.

Have not tested this yet, might eventually work.

thiolliere:
closing in favor of https://github.com/paritytech/parity-codec/pull/83

might be reopen if 83 fails to succeed

PR #85, opened 2019-05-31
Release 4.0.0 version

DarkEld3r:
Major version must be increased because `Encode`/`Decode` no longer implemented for `isize` and `usize`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,7 +25,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "parity-codec"
-version = "3.5.2"
+version = "4.0.0"
 dependencies = [
  "arrayvec 0.4.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "bitvec 0.11.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


thiolliere:
Wasn't usize and isize a bug ?

The purpose of this PR is to then update substrate master ? or v1.0 ?
because if it is to update substrate master maybe it is time now to release a 4.0 with all features of parity-codec master.

DarkEld3r:
@thiolliere Yes, I want to try to update update substrate master. Not sure about v.1.0. I think publishing `parity-codec` master as 4.0 is a good idea, but as far as I know we want more changes before that. I suppose @bkchr can elaborate.

thiolliere:
Yes ok, I don't know about pending changes, master most breaking change I think is just the signature of decode from option to result so it should be somewhat easy for user to update.

All in all that doesn't harm to publish new major version but actually it can be a bit time consuming to update all codebase: to use the new major in substrate we have to publish a new major version of trie, rohdodendron and few other I think. So would be cool to put maximum of things inside.

thiolliere:
Though if we want to get it for 1.0, it might be better not to include signature change so upgrading other lib is straightforward.

bkchr:
Master will be `parity-scale-codec` `1.0`.

bkchr:
@DarkEld3r please release `4.0`.

PR #91, opened 2019-06-03
impl more size hint

thiolliere:
related https://github.com/paritytech/parity-codec/issues/87

This PR is based on gui-impl-for-u8 for the moment, it introduce for size_hint implementation for different types.

question:
* Do we want to implement size hint when it cost iterating on all elements ? (for instance Vec, BTreeMap, BTreeSet)

depending on the answer I'll update test.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -524,6 +524,10 @@ impl<'a> Encode for CompactRef<'a, ()> {
 }
 
 impl<'a> Encode for CompactRef<'a, u8> {
+	fn size_hint(&self) -> usize {
+		Compact::compact_len(self.0)
+	}
+
 	fn encode_to<W: Output>(&self, dest: &mut W) {
 		match self.0 {
 			0..=0b0011_1111 => dest.push_byte(self.0 << 2),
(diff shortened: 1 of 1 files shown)


bkchr:
I think we should not loop, at the end it is more costly then just re-allocating.

thiolliere:
Yes agree at some point I was wondering if it would be worth introducing another size_hint which doesn't take any argument so a set of this could just ask for the maximum size 🤔

thiolliere:
done

bkchr:
I thought about the "maximum size" as well, but then we should drop `size_hint`. Because how would you decide which to call?

thiolliere:
we could keep both:
* maximum_size() used on set of values
* size_hint(&self) used on individual values,

size_hint(&self) (thought maybe renamed) would be default implementation to maximum_size()

bkchr:
Yeah sounds good, do you create an issue for this?

thiolliere:
https://github.com/paritytech/parity-codec/issues/93

bkchr:
ty

PR #94, opened 2019-06-04
Documentation for Parity SCALE Codec

gautamdhameja:


README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Implementation
+
+The codec is implemented using the following traits:
+
+### Encode
+
+The `Encode` trait is used for encoding of data into the SCALE format. The `Encode` trait contains the following functions:
+
+* `size_hint(&self) -> usize`: Get the capacity (in bytes) required for the encoded data. This is needed to make sure we are not double-allocating memory needed for the encoding. This just needs to be an estimate and does not have to be the exact number. If you don't know the exact size, give the maximum.
```
bkchr: Hmm. It's correct, but should be changed. 
This is **required** to be a cheap operation, so should not involve iterations etc. 

Your sentence sounds so demanding, that the user needs to make sure that no double allocation is done.

And also if the size is not known, even no good maximum. Just don't implement it :)

README.md
```diff
@@ -0,0 +1,109 @@
+# Parity SCALE Codec
+
+Rust implementation of the SCALE (Simple Concatenated Aggregate Little-Endian) data format for types used in the Parity Substrate framework.
+
+<!-- Inspired from Gav's overview written for Subtrate docs site -->
+SCALE is a light-weight format which allows encoding (and decoding) which makes it highly suitable for resource-constrained execution environments like blockchain runtimes and low-power, low-memory devices.
+
+It is important to note that the encoding context (knowledge of how the types and data structure looks like) needs to be known separately at both encoding and decoding ends. The encoded data does not include this contextual information.
```
JoshOrndorff:
```diff
- knowledge of how the types and data structure looks like
+ knowledge of what the types and data structures looks like
```
OR
```diff
- knowledge of how the types and data structure looks like
+ knowledge of how the types and data structures look
```

README.md
```diff
@@ -0,0 +1,109 @@
...
+* `encode(&self) -> Vec<u8>`: Encodes the type data and returns a slice.
+* `using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R`: Encodes the type data and executes a closure on the encoded value. Returns the result from the executed closure.
+
+**Note:** Implementations should override `using_encoded` for value types and `encode_to` for allocating types. `size_hint` should be implemented for all types. Wrapper types should override all methods.
+
+### Decode
+
+The `Decode` trait is used for deserialization/decoding of encoded data into the respective types.
+
+* `fn decode<I: Input>(value: &mut I) -> Result<Self, Error>`: Tries to decode the value from SCALE format to the type it is called on. Returns an `Err` if the decoding fails.
```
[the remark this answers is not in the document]
gautamdhameja: Mainly type mismatch - if the data being decoded is not of the expected type, the decoding would fail. I believe this is true for all decoding and deserializing operations and doesn't need to be mentioned explicitly.

PR #97, opened 2019-06-12
Derive `CompactAs` for single field structs

### LINKED ISSUE #96: Derive `CompactAs` for unit structs.
Should be rather trivial to write a derive for this.

ascjones:
Resolves #96 (assuming that @bkchr means "unit structs" = "single field structs")

- Derives `CompactAs` on named and unnamed structs with a single non-skipped field.
- Extracts common function to wrap impl block in "dummy const"

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -158,3 +158,62 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
 
 	generated.into()
 }
+
+#[proc_macro_derive(CompactAs, attributes(codec))]
+pub fn decode_compact_as(input: TokenStream) -> TokenStream {
+	let mut input: DeriveInput = match syn::parse(input) {
+		Ok(input) => input,
+		Err(e) => return e.to_compile_error().into(),
+	};
+
+	if let Err(e) = trait_bounds::add(
+		&input.ident,
+		&mut input.generics,
+		&input.data,
+		parse_quote!(_parity_scale_codec::CompactAs),
+		None,
+	) {
+		return e.to_compile_error().into();
+	}
+
+	let name = &input.ident;
+	let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+
+//	let decoding = decode::quote(&input.data, name, &input_);
+
+	let impl_block = quote! {
+		impl #impl_generics _parity_scale_codec::CompactAs for #name #ty_generics #where_clause {
+			type As = u32;
+			fn encode_as(&self) -> &u32 {
+				&self.0
+			}
+			fn decode_from(x: u32) -> #name #ty_generics {
+				#name(x)
+			}
+		}
+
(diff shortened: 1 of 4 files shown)


derive/src/lib.rs
```diff
@@ -140,21 +141,75 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+				},
+				Fields::Unnamed(ref fields) if fields.unnamed.iter().count() == 1 => {
+					let (id, field) = fields.unnamed.iter().enumerate().next().expect("Exactly one field");
+					let id = syn::Index::from(id);
+					(&field.ty, quote!(&self.#id), quote!(#name(x)))
+				},
+				_ => {
+					return Error::new(
+						Span::call_site(),
+						"Only structs with a single field can derive CompactAs"
```
thiolliere: Maybe we should also take into consideration the skip attribute here, and instead only struct with only a single not skipped field can derive CompactAs

bkchr: Yeah, @thiolliere has already written some code for the `using_encoded` optimizations.

derive/src/lib.rs
```diff
@@ -36,20 +36,37 @@ mod encode;
...
-				quote!( extern crate #ident as _parity_scale_codec; )
-			},
-			Err(e) => Error::new(Span::call_site(), &e).to_compile_error(),
-		}
-	}
+/// Wraps the impl block in a "dummy const"
+fn wrap_with_dummy_const(input: &DeriveInput, prefix: &str, impl_block: proc_macro2::TokenStream) -> TokenStream {
+	let parity_codec_crate =
+		// This "hack" is required for the tests.
+		if env::var("CARGO_PKG_NAME").unwrap() == "parity-scale-codec" {
```
bkchr: I would still like to see this as an extra function. 
Trying to minimize the size of a function is a good metric, I think

derive/src/lib.rs
```diff
@@ -140,21 +141,75 @@ pub fn decode_derive(input: TokenStream) -> TokenStream {
...
+			type As = #inner_ty;
+			fn encode_as(&self) -> &#inner_ty {
+				#inner_field
+			}
+			fn decode_from(x: #inner_ty) -> #name #ty_generics {
+				#constructor
+			}
+		}
+
+		impl #impl_generics From<_parity_scale_codec::Compact<#name #ty_generics>> for #name #ty_generics #where_clause {
```
bkchr: As you could read my mind :heart:

derive/src/lib.rs
ascjones: https://github.com/paritytech/parity-scale-codec/pull/97/commits/9e24127cda7d1756dd90e0758386d726d639231e

ascjones:
>  but I also would like to see a test that uses Compact<SomeType> where SomeType derives CompactAs.

Isn't use of `Compact<T>` transparent because of  `From` impls both ways?

bkchr:
You can also have a type `Compact<_>`. It does not is transparent.

PR #99, opened 2019-06-12
impl codec for collections

### LINKED ISSUE #76: Add Implementation std::collection::* Encode/Decoder.
I would like to implement Encode / Decode for some of the types defined in `std::collection`.
https://doc.rust-lang.org/std/collections/index.html

`Vec`, `VecDeque`, `LinkedList`, `BinaryHeap`, `BTreeMap`, `BTreeSet`.
(`Vec`, `BTreeMap` and `BTreeSet` is already done.)

Implement similarly for other collection types. 
We may create a macro whose arguments is a type name and insert method.

thiolliere:
Fix #76

BTree*, BinaryHeap and LinkedList are implemented the same way:
* encoding: length in compact ++ all elements
*decoding: get length + iter to get all elements

so there are now implemented using a macro to avoid writing the same code

VecDeque is optimized for Vec<u8>


src/codec.rs
```diff
@@ -1202,51 +1202,64 @@ impl<T: Encode + Decode> EncodeAppend for Vec<T> {
...
+		} else {
+			for item in self {
+				item.encode_to(dest);
+			}
 		}
 	}
 }
 
-impl<T: Decode + Ord> Decode for BTreeSet<T> {
+impl<T: Decode + Ord> Decode for VecDeque<T> {
```
```suggestion
impl<T: Decode> Decode for VecDeque<T> {
```

src/codec.rs
```diff
@@ -1202,51 +1202,64 @@ impl<T: Encode + Decode> EncodeAppend for Vec<T> {
...
-impl<T: Encode + Ord> Encode for BTreeSet<T> {
+impl<T: Encode + Ord> Encode for VecDeque<T> {
 	fn encode_to<W: Output>(&self, dest: &mut W) {
 		let len = self.len();
 		assert!(len <= u32::max_value() as usize, "Attempted to serialize a collection with too many elements.");
 		Compact(len as u32).encode_to(dest);
-		for i in self.iter() {
-			i.encode_to(dest);
+
+		if let IsU8::Yes= <T as Encode>::IS_U8 {
```
```suggestion
		if let IsU8::Yes = <T as Encode>::IS_U8 {
```

PR #100, opened 2019-06-13
Fixed links in readme.

gautamdhameja:


thiolliere:
could you update link in src/compact.rs as well ?

PR #101, opened 2019-06-13
`len` method for some collections.

Kianenigma:
- I wonder if it should also be added for string and tuple. String is not really used in substrate. Tuple might make a bit of sense.
- Just saw this PR https://github.com/paritytech/parity-scale-codec/pull/99 and the same could be added for the new types there --though we should probably unify the macros.)

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -225,9 +225,9 @@ pub trait EncodeAppend {
 
 /// Trait that allows the length of a collection to be read, without having
 /// to read and decode the entire elements.
-pub trait EncodeLength {
+pub trait DecodeLength {
 	/// Return the number of elements in `self_encoded`.
-	fn len(self_encoded: Vec<u8>) -> Result<u32, Error>;
+	fn len(self_encoded: &mut &[u8]) -> Result<u32, Error>;
 }
 
 /// Trait that allows zero-copy read of value-references from slices in LE format.
@@ -673,9 +673,9 @@ impl Decode for () {
 
 macro_rules! len_impl {
 	( $( $type:ident< $($g:ident),* > ),* ) => { $(
-		impl<$($g: Encode + Decode),*> EncodeLength for $type<$($g),*> {
-			fn len(self_encoded: Vec<u8>) -> Result<u32, Error> {
-				Ok(u32::from(Compact::<u32>::decode(&mut &self_encoded[..])?))
+		impl<$($g: Encode + Decode),*> DecodeLength for $type<$($g),*> {
+			fn len(self_encoded: &mut &[u8]) -> Result<u32, Error> {
+				Ok(u32::from(Compact::<u32>::decode(self_encoded)?))
 			}
 		}
 	)*}
@@ -1046,12 +1046,12 @@ mod tests {
 		assert_eq!(decoded, expected);
 	}
 
-	fn test_encode_length<T: Encode + Decode + EncodeLength>(thing: &T, len: usize) {
-		assert_eq!(<T as EncodeLength>::len(thing.encode()).unwrap(), len as u32);
+	fn test_encode_length<T: Encode + Decode + DecodeLength>(thing: &T, len: usize) {
+		assert_eq!(<T as DecodeLength>::len(&mut &thing.encode()[..]).unwrap(), len as u32);
 	}
 
 	#[test]
-	fn len_works_for_all_encode_types() {
+	fn len_works_for_all_decode_types() {
 		let vector = vec![10; 10];
 		let mut btree_map: BTreeMap<u32, u32> = BTreeMap::new();
 		btree_map.insert(1, 1);
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -34,6 +34,8 @@ use arrayvec::ArrayVec;
 #[cfg(feature = "std")]
 use std::fmt;
 
+use core::convert::TryFrom;
+
 #[cfg_attr(feature = "std", derive(Debug))]
 #[derive(PartialEq)]
 #[cfg(feature = "std")]
@@ -688,7 +690,8 @@ macro_rules! impl_len {
 	( $( $type:ident< $($g:ident),* > ),* ) => { $(
 		impl<$($g),*> DecodeLength for $type<$($g),*> {
 			fn len(self_encoded: &mut &[u8]) -> Result<usize, Error> {
-				Ok(u32::from(Compact::<u32>::decode(self_encoded)?) as usize)
+				usize::try_from(u32::from(Compact::<u32>::decode(self_encoded)?))
+					.map_err(|_| "can't convert".into())
 			}
 		}
 	)*}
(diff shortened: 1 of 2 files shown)


thiolliere:
yes String would make sense to me.

Tuple I'm not sure, if the feature is wanted we should extend with another trait like EncodeLenTuple1, EncodeLenTuple2 and implement them generically if their first (or second respectively) does implement EncodeLen.
Same we could do impl a trait for Option which would be EncodeLenSome which would return `Option<u32>`.
Even Result could have EncodeLenOk -> `Option<u32>` EncodeLenErr -> `Option<u32>`...
But I'm not sure of the interest.

src/codec.rs
```diff
@@ -223,6 +223,13 @@ pub trait EncodeAppend {
 	fn append(self_encoded: Vec<u8>, to_append: &[Self::Item]) -> Result<Vec<u8>, Error>;
 }
 
+/// Trait that allows the length of a collection to be read, without having
+/// to read and decode the entire elements.
+pub trait EncodeLength {
```
bkchr: I would like to call it `DecodeLength` or something similar, but the current name implies we want to encode a length.

src/codec.rs
```diff
@@ -223,6 +223,13 @@ pub trait EncodeAppend {
...
+	/// Return the number of elements in `self_encoded`.
+	fn len(self_encoded: Vec<u8>) -> Result<u32, Error>;
```
bkchr: Please be a nice rust programmer and just take a slice :D

bkchr:
#99 is the reason I said we want to have this for all collection types ;)

src/codec.rs
Kianenigma: Done but I more or less mimicked how `EncodeAppend` is treating the input. What's the rationale?

Kianenigma:
> #99 is the reason I said we want to have this for all collection types ;)

Then can we have that either merged to master or I merge it locally and close it (as it seems to be approved and CI is gree, anything else to do there @thiolliere?)

thiolliere:
about #99 there is not really things to merge actually, VecDeque and Vec and not implemented using `impl_codec_through_iterator` macro because they are optimized for T==u8.

so all in all the merge will just be adding those new collections to your macro.

whatever we merge first is fine

bkchr:
#99 is merged, so please update your pr :)

src/codec.rs
bkchr: `EncodeAppend` is another story. We try to reuse the given vector, as we append stuff to it directly. In this case, we just read a value. So it makes no sense to take the `Vec`.

src/codec.rs
```diff
@@ -664,6 +671,18 @@ impl Decode for () {
...
+	( $( $type:ident< $($g:ident),* > ),* ) => { $(
+		impl<$($g: Encode + Decode),*> EncodeLength for $type<$($g),*> {
+			fn len(self_encoded: Vec<u8>) -> Result<u32, Error> {
+				Ok(u32::from(Compact::<u32>::decode(&mut &self_encoded[..])?))
+			}
+		}
+	)*}
+}
+
+len_impl!(Vec<T>, BTreeSet<T>, BTreeMap<K, V>);
```
thiolliere:
```suggestion
len_impl!(Vec<T>, BTreeSet<T>, BTreeMap<K, V>, VecDeque<T>, BinaryHeap<T>, LinkedList<T>);
```

src/codec.rs
```diff
@@ -677,6 +684,19 @@ impl Decode for () {
 	}
 }
 
+macro_rules! impl_len {
+	( $( $type:ident< $($g:ident),* > ),* ) => { $(
+		impl<$($g: Encode + Decode),*> DecodeLength for $type<$($g),*> {
+			fn len(self_encoded: &mut &[u8]) -> Result<u32, Error> {
```
```suggestion
			fn len(self_encoded: &mut &[u8]) -> Result<usize, Error> {
```

src/codec.rs
```diff
@@ -677,6 +684,19 @@ impl Decode for () {
...
+				Ok(u32::from(Compact::<u32>::decode(self_encoded)?))
```
```suggestion
				Ok(usize::from(u32::from(Compact::<u32>::decode(self_encoded)?)))
```

Kianenigma:
> Some last stuff :D
> 
> Why `usize`? Rust normally exposes lenghts as `usize` and this probably makes other code easier (no `u32 as usize` all over the codebase etc).

I perfectly agree and started with that. BUT: don't we have the assumption here that lengths are always at most u32? hence the SCALE returns u32 and the caller is responsible for doing whatever it wants. Though I don't have a strong opinion bc I don't know where exactly SCALE is being used aside from substrate. 

Will address the rest soonish. Thanks

bkchr:
We encode it as `u32`, but this is rather an implementation detail. Returning `usize` as any other `len` function is more like a convention and makes it easier to compare it against other lengths.

src/codec.rs
```diff
@@ -677,6 +684,19 @@ impl Decode for () {
 	}
 }
 
+macro_rules! impl_len {
+	( $( $type:ident< $($g:ident),* > ),* ) => { $(
+		impl<$($g: Encode + Decode),*> DecodeLength for $type<$($g),*> {
```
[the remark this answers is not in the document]
Kianenigma: hmm I see that both are accepted but --while acknowledging that I know very limited of how scale is working-- it seemed more reasonable to express via the `impl`: Any container is only encodable + can have `DecodeLength` if its generic types are `Encode + Decode`. Hence `($g: Encode + Decode),*`.

src/codec.rs
Kianenigma: can't go from `u32 -> usize` afaik. `as usize` or `std::Convert::try_from`?

src/codec.rs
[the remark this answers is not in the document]
bkchr: If you really want to know the length of a non-encodable type, we won't block these people 😂
I understand your reasoning, but I think we should not force any bound that is not required.

src/codec.rs
bkchr: Use try_from.

src/codec.rs
```diff
@@ -677,6 +686,20 @@ impl Decode for () {
 	}
 }
 
+macro_rules! impl_len {
+	( $( $type:ident< $($g:ident),* > ),* ) => { $(
+		impl<$($g),*> DecodeLength for $type<$($g),*> {
+			fn len(self_encoded: &mut &[u8]) -> Result<usize, Error> {
+				usize::try_from(u32::from(Compact::<u32>::decode(self_encoded)?))
+					.map_err(|_| "can't convert".into())
```
bkchr: Sorry, but can be please get a more meaningful error message? :D

Kianenigma: Yeah I thought this was the rule (short messages) by seeing some of the other ones :p

Kianenigma:
```suggestion
					.map_err(|_| "Failed convert decded size into usize.".into())
```

PR #106, opened 2019-06-24
replace DecodeLength with ConstrainedDecode

rphmeier:
I think this is more general and will be better to integrate into SRML macros.

I'm thinking that it can integrate like this (within `decl_module!` invocation):

```rust
fn my_constrained_call(
    #[constrained(|| MaximumLength(Trait::MaxItems::get()))] items: Vec<Item>,
    #[constrained(|| build_custom_constraint())] value: MyStruct,
    unconstrained: u32, // no annotation means the macro automatically inserts `NoConstraint`
) { ... }
```

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -225,12 +225,27 @@ pub trait EncodeAppend {
 	fn append(self_encoded: Vec<u8>, to_append: &[Self::Item]) -> Result<Vec<u8>, Error>;
 }
 
-/// Trait that allows the length of a collection to be read, without having
-/// to read and decode the entire elements.
-pub trait DecodeLength {
-	/// Return the number of elements in `self_encoded`.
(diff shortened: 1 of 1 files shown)


thiolliere:
I don't understand the final point of this, in substrate let's say for the srml-contract:put_code call 
```
		pub fn put_code(
			origin,
			#[compact] gas_limit: T::Gas,
			code: Vec<u8>
		) -> Result {
```
there is no point to restrain code size because of decoding purpose no ?

bkchr:
I think as discussed in Riot, we can close this.

rphmeier:
It's not _necessary_ to be secure, but it is helpful to make sure decoding can fail fast.

bkchr:
I created an issue for this https://github.com/paritytech/parity-scale-codec/issues/112

PR #108, opened 2019-06-25
Fix preallocation

thiolliere:
The implementation of Input for io::Read was using read_exact, that seems wrong was it intented ?

So apart from that this introduce a maximum preallocation which is 16MiB, related to substrate network limit.

After merging I will push tag 3.5.2 and make a new release

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -18,8 +18,9 @@ use crate::alloc::vec::Vec;
 use crate::alloc::boxed::Box;
 use crate::alloc::collections::btree_map::BTreeMap;
 
-// This default has been choosen according to substrate network limit of 16MiB
-const MAXIMUM_PREALLOCATION_SIZE: usize = 16777216;
+/// When decoding input we don't allocate too much in advance so a small crafted input can't
+/// trigger a big allocation.
+const MAX_PREALLOCATION_SIZE: usize = 4*1024;
 
 #[cfg(any(feature = "std", feature = "full"))]
 use crate::alloc::{
@@ -785,7 +786,7 @@ impl Decode for Vec<u8> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 			let len = len as usize;
 
-			if len < MAXIMUM_PREALLOCATION_SIZE {
+			if len < MAX_PREALLOCATION_SIZE {
 				let mut vec = vec![0; len]; // len is ok here
 				if input.read(&mut vec[..]) != len {
 					None
@@ -795,26 +796,23 @@ impl Decode for Vec<u8> {
 			} else {
 				// if len is considered too much for preallocation then use dynamic allocation
 				let mut vec = Vec::new();
-				let buffer_len = MAXIMUM_PREALLOCATION_SIZE;
+				let mut remains = len;
+				let buffer_len = MAX_PREALLOCATION_SIZE;
 				let mut buffer = vec![0; buffer_len];
 				let mut total_len_read = 0;
 
-				loop {
-					let len_read = input.read(&mut buffer[..buffer_len]);
+				while remains != 0 {
+					let len_read = input.read(&mut buffer[..buffer_len.min(remains)]);
 
 					if len_read == 0 {
-						break
+						return None
 					}
 
-					total_len_read += len_read;
+					remains -= len_read;
 					vec.extend_from_slice(&buffer[..len_read]);
 				}
 
-				if total_len_read != len {
-					None
-				} else {
-					Some(vec)
-				}
+				Some(vec)
 			}
 		})
 	}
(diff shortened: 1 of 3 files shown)


thiolliere:
this fix is not perfect it allow to allocate in advance as such:

if you have `Vec<Vec<Vec<_>>>` then user can allocate MAX_PREALLOCATION_SIZE*3, this might be still a lot do we have some opinion on the value ?

rphmeier:
> if you have Vec<Vec<Vec<_>>> then user can allocate MAX_PREALLOCATION_SIZE*3, this might be still a lot do we have some opinion on the value ?

That's a good point. But it is at least limited based on the types we define...

thiolliere:
> > if you have Vec<Vec<Vec<_>>> then user can allocate MAX_PREALLOCATION_SIZE*3, this might be still a lot do we have some opinion on the value ?
> 
> That's a good point. But it is at least limited based on the types we define...

yes basically any nested struct with a vec field can preallocate MAX_PREALLOCATION_SIZE (+ fields before it)

src/codec.rs
```diff
@@ -781,11 +784,37 @@ impl Decode for Vec<u8> {
...
+			if len < MAXIMUM_PREALLOCATION_SIZE {
+				let mut vec = vec![0; len]; // len is ok here
+				if input.read(&mut vec[..len]) != len {
+					None
+				} else {
+					Some(vec)
+				}
 			} else {
-				Some(vec)
+				// if len is consider too much for preallocation then use dynamic allocation
```
andresilva:
```suggestion
				// if len is considered too much for preallocation then use dynamic allocation
```

thiolliere:
> This PR only deals with preallocation issues, but we are still allowing unbounded size decoding the difference is that now we'll try to do it with multiple allocations.

we are allowing `unbounded size decoding` but it is actually bounded by input size (which wasn't the case before)

src/codec.rs
```diff
@@ -781,11 +784,37 @@ impl Decode for Vec<u8> {
 	fn decode<I: Input>(input: &mut I) -> Option<Self> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 			let len = len as usize;
-			let mut vec = vec![0; len];
-			if input.read(&mut vec[..len]) != len {
-				None
+
+			if len < MAXIMUM_PREALLOCATION_SIZE {
+				let mut vec = vec![0; len]; // len is ok here
+				if input.read(&mut vec[..len]) != len {
```
```suggestion
				if input.read(&mut vec[..]) != len {
```

src/codec.rs
```diff
@@ -781,11 +784,37 @@ impl Decode for Vec<u8> {
...
 			} else {
-				Some(vec)
+				// if len is considered too much for preallocation then use dynamic allocation
+				let mut vec = Vec::new();
+				let buffer_len = MAXIMUM_PREALLOCATION_SIZE;
+				let mut buffer = vec![0; buffer_len];
+				let mut total_len_read = 0;
+
+				loop {
+					let len_read = input.read(&mut buffer[..buffer_len]);
```
bkchr: Okay, this algorithm is not correct. You need to do `len -= len_read` in each round. Always read `buffer[..min(buffer_len, len)]`.
I think you understand the rest.

thiolliere: oh yes, I still forget about this. done

thiolliere:
there is only BTreeMap in 3.5.1 and nothing is preallocated

src/codec.rs
```diff
@@ -781,10 +785,32 @@ impl Decode for Vec<u8> {
 	fn decode<I: Input>(input: &mut I) -> Option<Self> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 			let len = len as usize;
-			let mut vec = vec![0; len];
-			if input.read(&mut vec[..len]) != len {
-				None
+
+			if len < MAX_PREALLOCATION_SIZE {
+				let mut vec = vec![0; len]; // len is ok here
+				if input.read(&mut vec[..]) != len {
```
tomusdrw: I believe that's wrong. `std::io::Read::read` can just read less bytes in case they are not available yet (like when you read directly from network or a slow disk) - we should try in a loop until we get `0` like we do with dynamic allocation. I'd unify the two codes and just have a bool flag to control if we are allowed to go over current capacity.

bkchr: This is not `std::io::Read`.

bkchr: Or do you mean because we implement `std::io::Read` for `Input`. We should probably drop this implementation...
Even `0` does not mean that we read until the end of a file in certain situations.. And I clearly don't want to loop here.

bkchr: I created an issue for this. Ty.

bkchr: Okay, looked again into the implementation :sweat_smile: 
We use https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact for the `std::io::Read` impl. So, we are safe and don't need to do that.

thiolliere:
Hmm I changed it to read when I implemented the fix in this PR because of
the first wrong implementation of the fix. Hmm yes we probably want to put
it back to read_exact. I missed that

src/codec.rs
tomusdrw: Yeah, that was my point. Thanks!

thiolliere:
(I don't know our usage but if we are using read this way (on waiting
input) then this should be fixed and release to help audit ppl.)

PR #111, opened 2019-06-25
Fix div

thiolliere:
hmm how did I forget that.


thiolliere:
all test of substrate works with this

PR #117, opened 2019-07-03
Better API

thiolliere:
it doesn't really make sense to me to have a read_exact function that returns the len specified on input.

This make the API more clear.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -94,7 +94,7 @@ impl From<&'static str> for Error {
 
 /// Trait that allows reading of data into a slice.
 pub trait Input {
-	/// Read the exact number of bytes required to fill the specified buffer `into`.
+	/// Read the exact number of bytes required to fill the given buffer.
 	///
 	/// Note that this function is similar to `std::io::Read::read_exact` and not
 	/// `std::io::Read::read`.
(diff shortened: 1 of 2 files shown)


src/codec.rs
```diff
@@ -94,17 +94,11 @@ impl From<&'static str> for Error {
 
 /// Trait that allows reading of data into a slice.
 pub trait Input {
-	/// Read into the provided input slice. Returns the number of bytes read.
+	/// Read the exact number of bytes required to fill the specified buffer `into`.
```
```suggestion
	/// Read the exact number of bytes required to fill the given buffer.
```

src/codec.rs
```diff
@@ -116,14 +110,14 @@ pub trait Input {
 
 #[cfg(not(feature = "std"))]
 impl<'a> Input for &'a [u8] {
-	fn read(&mut self, into: &mut [u8]) -> Result<usize, Error> {
+	fn read(&mut self, into: &mut [u8]) -> Result<(), Error> {
 		if into.len() > self.len() {
-			return Err("".into());
+			return Err("input length smaller than given buffer".into());
```
```suggestion
			return Err("Not enough data to fill buffer".into());
```

src/codec.rs
```diff
@@ -116,14 +110,14 @@ pub trait Input {
...
 		}
-		let len = core::cmp::min(into.len(), self.len());
+		let len = into.len();
 		into[..len].copy_from_slice(&self[..len]);
```
```suggestion
		into.copy_from_slice(&self[..len]);
```

Issue #109, opened 2019-06-25
Improve preallocation by changing Input definition


thiolliere:
fixed on v4 at https://github.com/paritytech/parity-scale-codec/pull/114
fixed on v3 at https://github.com/paritytech/parity-scale-codec/pull/108

but we want to improve the fix by changing input definition see Basti comment https://github.com/paritytech/parity-scale-codec/issues/109#issuecomment-505583916

thiolliere:
for this to be implemented properly we need to get the size of the input, then we can allocate size of buffer according to input size, thus it would allow us to have the performance of allocating huge buffer and still not allow user to allocate huge buffer by crafting small inputs

bkchr:
We would not allocate the buffer based on the input size. We need to introduce something like `minimum_size()` that returns the minimal size a data type requires in encoded format. Then we can check that `requested_len * minimum_size() <= input_len`. If this is true, we can continue. Maybe this needs a little bit more thinking, but should go into the right direction.

thiolliere:
OK I made a bit of research:

I implemented the following additional associated function for the trait Input:
```rust
/// Trait that allows reading of data into a slice.
pub trait Input {
	/// Require the input to be at least the len specified.
	fn require_minimum_len(&mut self, len: usize) -> Result<(), Error>;

	/// Read the exact number of bytes required to fill the given buffer.
	///
	/// Note that this function is similar to `std::io::Read::read_exact` and not
	/// `std::io::Read::read`.
	fn read(&mut self, into: &mut [u8]) -> Result<(), Error>;

	/// Read a single byte from the input.
	fn read_byte(&mut self) -> Result<u8, Error> {
		let mut buf = [0u8];
		self.read(&mut buf[..])?;
		Ok(buf[0])
	}
}
```

thus ppl can implement this for a std::io::Read like this:
```rust
#[cfg(feature = "std")]
pub struct IoReader<R: std::io::Read> {
	buffer: Vec<u8>,
	reader: R,
}

#[cfg(feature = "std")]
impl<R: std::io::Read> Input for IoReader<R> {
	fn require_minimum_len(&mut self, len: usize) -> Result<(), Error> {
		if self.buffer.len() >= len {
			return Ok(())
		}

		let filled_len = self.buffer.len();
		self.buffer.resize(len, 0);
		self.reader.read_exact(&mut self.buffer[filled_len..])?;
		Ok(())
	}

[...]

Issue #112, opened 2019-06-26
Constrained decode


bkchr:
We maybe should provide an interface for constrained decoding. We currently don't have any ideas how this could look like.

See https://github.com/paritytech/parity-scale-codec/pull/106 for more information.

bkchr:
If someone wants to take this, I would propose to create a new crate that sits on top of `parity-scale-codec`.

thiolliere:
it seems to me contrained decode is just about adding some assert during the decoding process.

One idea would be a trait:
```rust
trait DecodeConstrained {
  type Constraints;
  fn decode_constrained(input, constraints: Constraint) {
    // depending on constraints do some assert during the decode
  }
}
```
one example would be:
```rust
impl DecodeConstrained for Vec<T> {
type Constrained = struct {
  max_limite: Option<u32>,
  on_new_element: Option<callback: Fn(&vec<T>) -> bool>
}
}
```
but that might be inefficient for decoding a vec then. But that can be handy to derive your own decode_constrained:
```rust
#[derive(DecodeConstrained)]
struct MyStruct {
  #[constraint("... callback that returns a bool depending on field")]
  field: signature,
  #[constraint("... callback that returns a bool depending on field and field2")]
  field2: signed_content,
  #[constraint(optional_constrained_field)]
  field3: Vec<T>, 
  #[constraint(struct Constrained { max_limit: Some(3), Default::default() })]
  field4: Vec<T>, 
  field5: Vec<T>,
}

// derives into:

struct MyStructConstraint {
  callback_field: Option<..>,
  callback_field2: Option<..>,
  constraint_field3: Option<<Vec<T> as ConstrainedDecode>::Constraints>,
}

// and during derivation of decode_constrained:
// it decodes field4 with the constrain max_limit=3
// and it decodes field5 just using decode and not decode_constrained
```

PR #119, opened 2019-07-04
Implement min_encoded_len and required_len

thiolliere:
Fix https://github.com/paritytech/parity-scale-codec/issues/109

Introduce `fn min_encoded_len` on `Decode` trait and `fn require_min_len` on `Input` trait.

Then when preallocating `Vec` or `array` in decode function we ensure that a valid result can actually be constructed with the input lenght. We do that by requiring that lenght of input of more than a minimum.

Thus someone cannot craft a false input which allocate tons of memory.

However if user expect its input to have certain property more than just being a valid struct it is not responsability of parity-scale-codec to consider those properties.
For example if user create: `struct A([u8; 1024])` and manually implement encode so that the array is encoded such as a `Vec`, then minimum encoded size of `[A; 32]` is 32 bytes but allocates 32*1024. Thus if this user consider that an input with mostly empty arrays is a wrong input he must implement Encode/Decode differently, this is not responsability of parity-scale-codec. Also such a situation might want to make use of constrained decoding principles proposed here: https://github.com/paritytech/parity-scale-codec/issues/112

Test has been implemented using a mock, I'm not sure it is the best strategy.

### TODO

* [x] implementation
* [x] update the README
* [x] do test minimum_len
* [ ] fuzzing of minimum_len as well ?
* [x] a bit more documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ The `Encode` trait is used for encoding of data into the SCALE format. The `Enco
 
 The `Decode` trait is used for deserialization/decoding of encoded data into the respective types.
 
+* `fn min_encoded_len() -> usize`: The minimum lenght a valid encoded value can have.
 * `fn decode<I: Input>(value: &mut I) -> Result<Self, Error>`: Tries to decode the value from SCALE format to the type it is called on. Returns an `Err` if the decoding fails.
 
 ### CompactAs
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ The `Encode` trait is used for encoding of data into the SCALE format. The `Enco
 
 The `Decode` trait is used for deserialization/decoding of encoded data into the respective types.
 
-* `fn min_encoded_len() -> usize`: The minimum lenght a valid encoded value can have.
+* `fn min_encoded_len() -> usize`: The minimum length a valid encoded value can have.
 * `fn decode<I: Input>(value: &mut I) -> Result<Self, Error>`: Tries to decode the value from SCALE format to the type it is called on. Returns an `Err` if the decoding fails.
 
 ### CompactAs
--- a/derive/src/decode.rs
+++ b/derive/src/decode.rs
@@ -72,8 +72,9 @@ pub fn quote(data: &Data, type_name: &Ident, input: &TokenStream) -> Result<Impl
 					},
 				};
 
+				let variant_index_len = 1usize;
 				let min_encoded_len = quote_spanned! { v.span() =>
-					1 + #impl_min_encoded_len
+					#variant_index_len + #impl_min_encoded_len
 				};
 
 				Ok(Impl { decode, min_encoded_len })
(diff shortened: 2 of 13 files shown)


tests/skip.rs
```diff
@@ -1,7 +1,21 @@
 #[macro_use]
 extern crate parity_scale_codec_derive;
 
-use parity_scale_codec::{Encode, Decode};
+use parity_scale_codec::{Encode, Decode, Error};
+
+/// Mock
+pub trait DecodeM: Decode {
```
bkchr: This should clearly be moved somewhere where it can be reused.^^

src/codec.rs
```diff
@@ -401,7 +464,12 @@ impl<T: Encode> Encode for Option<T> {
 	}
 }
 
+// TODO TODO: hmm so an Option<[u8; 1024]> can do some damage no ?
```
bkchr: No.

Kianenigma:
> Test has been implemented using a mock, I'm not sure it is the best strategy.

I think it is fine but also would be very good to see a test where something fails (or takes too long) with normal `decode` and works with `decode_m` (or anything else that explicitly shows which problem this PR solves).

README.md
```diff
@@ -28,6 +28,7 @@ The `Encode` trait is used for encoding of data into the SCALE format. The `Enco
 
 The `Decode` trait is used for deserialization/decoding of encoded data into the respective types.
 
+* `fn min_encoded_len() -> usize`: The minimum lenght a valid encoded value can have.
```
4meta5:
```suggestion
* `fn min_encoded_len() -> usize`: The minimum length a valid encoded value can have.
```

derive/src/decode.rs
```diff
@@ -15,82 +15,118 @@
...
+				let decode = quote_spanned! { v.span() =>
 					x if x == #index as u8 => {
-						#create
+						#impl_decode
 					},
-				}
+				};
+
+				let min_encoded_len = quote_spanned! { v.span() =>
+					1 + #impl_min_encoded_len
```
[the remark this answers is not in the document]
thiolliere: do you think it should be an exposed variable in parity-scale-codec ?
probably not, I implement it as a documented const in derive for now.

README.md
thiolliere: thanks

derive/src/decode.rs
```diff
@@ -15,82 +15,118 @@
...
-			Fields::Named(_) | Fields::Unnamed(_) => create_instance(
+			Fields::Named(_) | Fields::Unnamed(_) => fields_impl(
 				call_site,
 				quote! { #type_name },
 				input,
 				&data.fields,
 			),
 			Fields::Unit => {
-				quote_spanned! {call_site =>
+				let decode = quote_spanned! {call_site =>
```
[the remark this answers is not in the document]
thiolliere: I replaced with field span at some places and improve span globally, for some cases I'm not sure if it is interesting to have span or if it is irrelevant. all in all it doesn't harm.

thiolliere:
@bkchr comments addressed

@Kianenigma I add tests for invalid input.
I'm not sure about hardcoded value, what is important is that things are kept coherent between Encode and Decode.

PR #122, opened 2019-07-08
Move Input and Output traits to a separate crate codec-io

sorpaas:
We will eventually face the issue of dealing with multiple binary encoding formats (such as ssz). In those situations, it is apparent that serde's model (using one `Serialize` and `Deserialize` trait for everything) is not sufficient. For example, U256 needs to be serialized as string in json, while as bytearray in SCALE and ssz. This is not possible to be represented by basic Rust structures.

Instead, we need to define Encode/Decode traits for each format we want to encode (have `scale::Encode`, `ssz::Encode`, etc). Those definitions can share common traits `Input` and `Output`. In this PR, I moved those two traits into a separate crate `codec-io` to make it reusable.

~~Note that I also generalized `Error` -- in some other encoding libraries we may not want to represent Error as solely `&str` or unit type and want to actually derive the error and pass it to library user. This requires changing `Decode` trait in parity-codec to have an extra bound `From<I::Error>`, which is a breaking change.~~ (This is not needed any more. I tried a different design that alias `codec_io::Error` to `std::io::Error` in `std`.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "parity-scale-codec"
 description = "SCALE - Simple Concatenating Aggregated Little Endians"
-version = "1.0.0"
+version = "2.0.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "Apache-2.0"
 repository = "https://github.com/paritytech/parity-scale-codec"
@@ -11,13 +11,14 @@ edition = "2018"
 [dependencies]
 arrayvec = { version = "0.4", default-features = false, features = ["array-sizes-33-128", "array-sizes-129-255"] }
 serde = { version = "1.0", optional = true }
-parity-scale-codec-derive = { path = "derive", version = "1.0", default-features = false, optional = true }
+parity-scale-codec-derive = { path = "derive", version = "2.0", default-features = false, optional = true }
+codec-io = { version = "0.1", path = "io", default-features = false }
 bitvec = { version = "0.11", default-features = false, features = ["alloc"], optional = true }
 byte-slice-cast = { version = "0.3.1", optional = true }
 
 [dev-dependencies]
 serde_derive = { version = "1.0" }
-parity-scale-codec-derive = { path = "derive", version = "1.0", default-features = false }
+parity-scale-codec-derive = { path = "derive", version = "2.0", default-features = false }
 criterion = "0.2"
 bitvec = "0.11"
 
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "parity-scale-codec"
 description = "SCALE - Simple Concatenating Aggregated Little Endians"
-version = "1.0.0"
+version = "2.0.0"
```
bkchr: This version is not released, so please revert this.

derive/src/encode.rs
```diff
@@ -100,28 +100,28 @@ fn encode_fields<F>(
 			let field_type = &f.ty;
 			quote_spanned! {
 				f.span() => {
-					#dest.push(
+					_parity_scale_codec::Encode::encode_to(
 						&<<#field_type as _parity_scale_codec::HasCompact>::Type as
-							_parity_scale_codec::EncodeAsRef<'_, #field_type>>::RefType::from(#field)
+							_parity_scale_codec::EncodeAsRef<'_, #field_type>>::RefType::from(#field),
+						#dest
```
bkchr:
```suggestion
						#dest,
```

derive/src/encode.rs
```diff
@@ -100,28 +100,28 @@ fn encode_fields<F>(
...
 		} else if let Some(encoded_as) = encoded_as {
 			let field_type = &f.ty;
 			quote_spanned! {
 				f.span() => {
-					#dest.push(
+					_parity_scale_codec::Encode::encode_to(
 						&<#encoded_as as
-							_parity_scale_codec::EncodeAsRef<'_, #field_type>>::RefType::from(#field)
+							_parity_scale_codec::EncodeAsRef<'_, #field_type>>::RefType::from(#field),
+						#dest
```
bkchr:
```suggestion
						#dest,
```

io/src/lib.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright 2017, 2018 Parity Technologies
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies
```

io/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Error for slice-based input. Only used in `no_std` environments.
+#[cfg(not(feature = "std"))]
+#[derive(PartialEq, Eq, Clone)]
+pub enum SliceInputError {
+	/// Not enough data to fill the buffer.
+	NotEnoughData,
+}
+
+#[cfg(not(feature = "std"))]
```
bkchr: Why is this disabled on `std`?

sorpaas:
@bkchr `bytes` crate does not support `no_std` at this moment, or do you know if upstream plans to add support in the near future?

io/src/lib.rs
sorpaas: As far as I understand in `std` slice implements `io::Read`, so we cannot have this here.

sorpaas:
(I'm going to temporarily close this PR and embed this in ssz for now. I realized that there're still a few design that I haven't finalized.)

bkchr:
They are working on it: https://github.com/tokio-rs/bytes/pull/256

PR #125, opened 2019-07-12
Fuzzer cleanup

bkchr:


.gitlab-ci.yml
```diff
@@ -48,6 +48,23 @@ bench:rust:nightly:
...
+    - cd fuzzer && HFUZZ_RUN_ARGS="-t 5 -n 2 -v -N 10000 --exit_upon_crash" cargo hfuzz run codec-fuzzer
+  only:
+    - triggers
+    - tags
+    - master
+    - schedules
+    - web
+    - /^[0-9]+$/
+  tags:
+    - rust
```
TriplEight:
```suggestion
    - linux-docker
```

bkchr:
Closes: https://github.com/paritytech/parity-scale-codec/issues/116

src/lib.rs
```diff
@@ -90,108 +89,106 @@
...
 //! # fn main() { }
 //! ```
 //! ### Type with CompactAs
-//! 
+//!
 //! ```rust
-//! 
+//!
 //! use serde_derive::{Serialize, Deserialize};
-//! use parity_scale_codec_derive::{Encode, Decode};;
```
thiolliere: this break `cargo test --features bit-vec`
also those test must be added to CI

thiolliere: hmm I'm actually surprised, `cargo test --all` works but not `cargo test` maybe it is fine to consider that test must be launch with command as in gitlab-ci. which means which appropriates features

thiolliere: I mean at some point no CI is just succeding because of a weird cargo behavior. I will change ci to pass features derive explicitely

fuzzer/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+	};
+	(@INTERNAL
+		$data:ident;
+		$counter:expr;
+		{ $( $parsed:ty; $index:expr ),* }
+	) => {
+		let num = $counter;
+		$(
+			if $data[0] % num == $index {
+				let _ = <$parsed>::decode(&mut &$data[1..]);
```
thiolliere: would it make sense to add a test for min_encoded_len here ? like if result is success then we check min_encoded_len is coherent

src/lib.rs
thiolliere: but also tests shouldn't fail if the feature is not used, I'll move tests to with features tests only

fuzzer/src/main.rs
bkchr: You can add this.

thiolliere:
so now tests are made to be compile for both with and without derive feature, and also gitlab test for both.

But we might want to have test derive and test no derive requried, how can we do so ? @gabreal

bkchr:
Just edit the gitlab.yaml file and add the cargo commands.

thiolliere:
yes this is done, just the test seems not mandatory for merging PR, but they are appearing

bkchr:
Why do you think that they are not mandatory?

thiolliere:
oh I think I missread, ok great

thiolliere:
yes ok but no as now github is waiting for continuous-integration/gitlab-test:rust:stable but should wait for both:

continuous-integration/gitlab-test:rust:stable:derive
continuous-integration/gitlab-test:rust:stable:no_derive

gabreal:
just made ` continuous-integration/gitlab-test:rust:stable:no_derive ` mandatory. the name of the ci job `test:rust:stable` hasn't changed to no need to modify.

Issue #107, opened 2019-06-25
Introduce `decode_all` method that ensure that Input of decode is empty at end of call


thiolliere:
I would expect that in some places people don't check that there input is actually empty after decoding a structure. And that is ok in some places but might not be in some other.

For example: in srml/contract:

https://github.com/paritytech/substrate/blob/1c2eaa3bc6a5659fb70cfcc88037573921612064/srml/contracts/src/wasm/runtime.rs#L566-L583

```rust
	ext_dispatch_call(ctx, call_ptr: u32, call_len: u32) => {
		let call = {
			let call_buf = read_sandbox_memory(ctx, call_ptr, call_len)?;
			<<<E as Ext>::T as Trait>::Call>::decode(&mut &call_buf[..])
				.ok_or_else(|| sandbox::HostError)?
		};

		// Charge gas for dispatching this call.
		let fee = {
			let balance_fee = <<E as Ext>::T as Trait>::ComputeDispatchFee::compute_dispatch_fee(&call);
			approx_gas_for_balance::<<E as Ext>::T>(ctx.gas_meter.gas_price(), balance_fee)
		};
```

this is ok but I wouldn't expect reading the code that `call_len` can be different from `call.encode().len()`

how about introducing a decode_all method that check that just check that input is empty after having decoded value. this cost nothing just adding a provided method on the trait Decode

bkchr:
I'm not really sure this fits into the context of this codec, but as we otherwise would start to replicate the stuff all over the place, let's integrate it.
```
trait DecodeAll {
    fn decode_all(bla..);
}

impl DecodeAll for Decode {
    ...
}
```

I would like to see this as a different trait at least.

PR #126, opened 2019-07-12
Add and implement `DecodeAll`

bkchr:
Closes: https://github.com/paritytech/parity-scale-codec/issues/107

### DIFF
--- a/src/decode_all.rs
+++ b/src/decode_all.rs
@@ -1,0 +1,108 @@
+// Copyright 2017, 2018 Parity Technologies
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::{Error, Decode};
+
+/// The error message returned when `decode_all` fails.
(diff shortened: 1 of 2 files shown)


src/decode_all.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// The error message returned when `decode_all` fails.
+const DECODE_ALL_ERR_MSG: &str = "Input buffer has still data left after decoding!";
+
+/// Extension trait to [`Decode`] that ensures that the given input data is consumed completly while
+/// decoding.
+pub trait DecodeAll: Sized {
+	/// Decode `Self` and consume all of the given input data. If not all data is consumed, an error
+	/// is returned.
+	fn decode_all(input: &[u8]) -> Result<Self, Error>;
```
thiolliere: At some point I feel like it would be better to have an extension of the signature as well instead of a restriction:

We can create a trait 
```rust
TerminableInput: Input {
  fn terminated(&mut self) -> bool;
}
```
and having decode_all working on this input

bkchr: Hmm, yeah maybe.

bkchr: But maybe we can come up with a better name?

thiolliere: InputEnd, or InputEndable with a is_ended or ended method

bkchr: The thing is that almost every input is endable. Maybe we should just support any input and add the `is_eof` method to input.

thiolliere: yes it is a possibility but at the same time is_eof seems to be never needed for codec because of its design. Maybe someone could want not to have to implement this method, but I can't really get an example

bkchr: I thought that you never want to have decodeall on something different than a slice of u8. :D
But for input, you could just return `false` when you have an input that never ends.

thiolliere: that is probably true, and yes we can do that indeed.

Or something like InputExtended can be something also.

PR #129, opened 2019-07-17
no std works on stable + CI

thiolliere:
it couldn't compile for no std because of some std usage, so this is solved,

and also we now can compile for stable as alloc is stable.

Moreover CI is improved

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -48,6 +48,48 @@ test:rust:stable:no_derive:
   tags:
     - rust
 
+check:rust:stable:no_derive_no_std_full:
+  stage:                           test
+  script:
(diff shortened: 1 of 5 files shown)


thiolliere:
I also updated version to 1.0.1 as I already published 1.0.0.

thiolliere:
@gabreal so I extended here as well in order to check for more features.

at some point we will want to test for it but 

continuous-integration/gitlab-check:rust:stable:no_derive_full 
continuous-integration/gitlab-check:rust:stable:no_derive_no_std 
continuous-integration/gitlab-check:rust:stable:no_derive_no_std_full

(we will probably move them into test instead of check in the future.)

PR #130, opened 2019-07-18
impl debug for error in no_std + fix derive macro

thiolliere:
we don't implement debug for error in no_std, this is very not handy when we want to use expect for instance

also derive was making use of Result, which can conflict with some locally defined result.

I have hard time dealing with patch on wasm build in for substrate but all native seems to build and wasm seems only to struggle with the Debug not being implemented.

Update codec to 1.0.2 and derive to 1.0.1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,7 +82,7 @@ version = "0.1.0"
 dependencies = [
  "bitvec 0.14.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "honggfuzz 0.5.45 (registry+https://github.com/rust-lang/crates.io-index)",
- "parity-scale-codec 1.0.1",
+ "parity-scale-codec 1.0.2",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


thiolliere:
also as a note we have new confusing error message, rustc thinks WrapperTypeDecode is the appropriate way to implement Decode. But the full error message is ok actually

```rust
   --> /home/thiolliere/Developpement/substrate/core/sr-primitives/src/traits.rs:460:6
    |
460 | impl Hash for BlakeTwo256 {
    |      ^^^^ the trait `_IMPL_DECODE_FOR_ModuleId::_parity_scale_codec::WrapperTypeDecode` is not implemented for `primitive_types::H256`
    |
    = note: required because of the requirements on the impl of `_IMPL_DECODE_FOR_ModuleId::_parity_scale_codec::Decode` for `primitive_types::H256`

```

thiolliere:
in substrate there is DecodeFails, it can't really implement min_encoded_len.
It would want to return None,

At the same time I think this structure is quite useless (and used only in a test in substrate). So I think it is fine

PR #134, opened 2019-07-21
Add generic-array and vecarray support

sorpaas:
This adds generic-array and vecarray support in scale-codec.


src/codec.rs
```diff
@@ -912,6 +912,51 @@ macro_rules! impl_non_endians {
 impl_endians!(u16, u32, u64, u128, i16, i32, i64, i128);
 impl_non_endians!(u8 {IS_U8}, i8, bool);
 
+#[cfg(feature = "generic-array")]
+impl<T: Encode, L: generic_array::ArrayLength<T>> Encode for generic_array::GenericArray<T, L> {
```
bkchr: Could you please move this into separate files? Like bitvec.

src/generic_array.rs
```diff
@@ -0,0 +1,25 @@
...
+	fn encode_to<W: Output>(&self, dest: &mut W) {
+		for item in self.iter() {
+			item.encode_to(dest);
+		}
+	}
+}
+
+impl<T: Decode, L: generic_array::ArrayLength<T>> Decode for generic_array::GenericArray<T, L> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let mut r = Vec::new();
```
thiolliere: wouldn't it be better to have with_capacity here ? it doesn't harm as this capacity is on the type itself, nobody can craft it with wrong input.

src/lib.rs
```diff
@@ -242,6 +242,10 @@ mod joiner;
 mod keyedvec;
 #[cfg(feature = "bit-vec")]
 mod bit_vec;
+#[cfg(feature = "generic-array")]
```
thiolliere: "generic-array" is not a feature in Cargo.toml

src/vec_array.rs
```diff
@@ -0,0 +1,22 @@
...
+	fn encode_to<W: Output>(&self, dest: &mut W) {
+		for item in self.iter() {
+			item.encode_to(dest);
+		}
+	}
+}
+
+impl<T: Decode, L: typenum::Unsigned> Decode for vecarray::VecArray<T, L> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let mut r = Vec::new();
```
thiolliere: wouldn't it be better to have with_capacity here ? it doesn't harm as this capacity is on the type itself, nobody can craft it with wrong input.

thiolliere:
also add those features in CI

Cargo.toml
```diff
@@ -14,6 +14,9 @@ serde = { version = "1.0", optional = true }
 parity-scale-codec-derive = { path = "derive", version = "1.0", default-features = false, optional = true }
 bitvec = { version = "0.14.0", default-features = false, features = ["alloc"], optional = true }
 byte-slice-cast = { version = "0.3.1", optional = true }
+generic-array = { version = "0.12", optional = true }
+vecarray = { version = "0.1", optional = true }
+typenum = { version = "1.10", optional = true }
```
thiolliere: do we need also to add `default-features = false` ? same for byte-slice-cast actually, is this needed

src/lib.rs
bkchr: Dependencies are automatically added as features in Cargo.

Cargo.toml
sorpaas: Fixed! `generic-array` and `typenum` does not need `std` feature.

sorpaas: `byte-slice-cast` pulls in `std` by default and does not support `no_std`. So when we're building `bitvec` it just won't work for `no_std` environments.

thiolliere:
Looks good, do we want to implement optimisation for `T=u8` similarly to Encode of `[u8]` and Decode of `Vec<u8>` ?

src/generic_array.rs
```diff
@@ -0,0 +1,25 @@
+use crate::{Encode, Decode, Error, Output, Input};
```
bkchr: Missing copyright and no tests.

src/vec_array.rs
```diff
@@ -0,0 +1,22 @@
+use crate::{Encode, Decode, Error, Output, Input};
```
bkchr: Same.

src/generic_array.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright 2017, 2018 Parity Technologies
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies
```

src/vec_array.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright 2017, 2018 Parity Technologies
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies
```

Issue #131, opened 2019-07-19
`min_encoded_len` doesn't work for recursive type


thiolliere:
```rust
#[test]
fn stack_overflow() {
	#[derive(Encode, Decode)]
	pub enum Toto {
		T(Box<Toto>)
	}

	Toto::min_encoded_len();
}
```

thiolliere:
one idea would be that min_encoded_len has one parameter which is a set of typeid and when implementing for a recursive type if its type id is already in the set then it returns max_value.

thiolliere:
but this solution can be done in two way:
* every type register itself in the set (which is probably bad because it has some cost)
* potential recursive types register themself in the set (that means all abstract structure like Result, Box or some enums with only abstract variants).

Both solution seems bad actually

thiolliere:
(at some point `const fn` could be of some help as it is not stable)

thiolliere:
or we could do something like instead of having min_encoded_len returning the actual min_encoded_len it returns either a final result either a slice of callback, and those callbacks are discarded if there current computed result is more than the current minimum final result found.

But this seems overkill as well.

thiolliere:
another solution would be to change the strategy completely.

At some point what do we want to handle as Input ? we usually use [u8] so we could easily bound preallocation with input len. The decoded value can need more allocation than input len due to skip, compact for example, but this limit should be good enough for most usage.

So the idea would be that intead of having decode requiring input to be a certain length we could do the other way around: having decoding asking a bound from input and input could return (maybe optionnaly) its len or whatever bound it thinks is fine.

bkchr:
I think the last idea is the best.

thiolliere:
though the third as for now the same flaws as the Fix currently in v4 as to say https://github.com/paritytech/parity-scale-codec/pull/108#issuecomment-505513896

> this fix is not perfect it allow to allocate in advance as such:
if you have Vec<Vec<Vec<_>>> then user can allocate MAX_PREALLOCATION_SIZE*3

to avoid this I can currently think of:
```rust
/// Trait that allows zero-copy read of value-references from slices in LE format.
pub trait Decode: Sized {
...
	/// Attempt to deserialise the value from input.
	fn decode<I: Input>(value: &mut I, input_used: usize) -> Result<Self, Error>;
}

/// Trait that allows reading of data into a slice.
pub trait Input {
...
	fn min_len(&mut self) -> usize {
		0
	}
}
```
[...]

PR #135, opened 2019-07-22
Maximum preallocation in decode

thiolliere:
base on this https://github.com/paritytech/parity-scale-codec/issues/131#issuecomment-513715570

To ensure total preallocation doesn't overflow some maximum we need to make information moving from caller to callee in a way.

Two possible ways:
* caller send with decode what it actually pre-allocated for later datas. thus callee can bound its preallocation to a maximum minus this already pre-allocated size.
Implemented in this PR
* caller send to decode what is the maximum pre-allocation callee is allowed, callee will have to respect this number.
This would be done with having `decode(input, Some(maximum))` allowing user to specify a maxium to preallocate or no maximum if wanted. Actually this might be better, maybe trusted input would be without limit on preallocation.

Also what changes is:
* input is not longer implemented for Read because min_len is very not accurate, I can still do a wrapper though
* DEFAULT_MAX_PREALLOCATION_SIZE is exposed publicly so users can make use of it.

Question: Finally I think implementing second way with `decode(input, preallocation: Option<usize>)` might be better after all.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -44,7 +44,6 @@ use std::fmt;
 /// Such bound is necessary to avoid crafted input to allocate too much space
 pub const DEFAULT_MAX_DECODE_PREALLOCATION_SIZE: usize = 4 * 1024;
 
-
 /// Descriptive error type
 #[cfg(feature = "std")]
 #[derive(PartialEq, Debug)]
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -28,14 +28,22 @@ use crate::alloc::{
 };
 
 use core::{mem, slice, ops::Deref};
-use core::marker::PhantomData;
+use core::convert::TryFrom;
 use core::iter::FromIterator;
+use core::marker::PhantomData;
 use arrayvec::ArrayVec;
 
 #[cfg(feature = "std")]
 use std::fmt;
 
-use core::convert::TryFrom;
+/// Default bound for preallocation in decoding.
+///
+/// If length of input is more than the default then
+/// decode preallocation can be bound by input length.
+///
+/// Such bound is necessary to avoid crafted input to allocate too much space
+pub const DEFAULT_MAX_DECODE_PREALLOCATION_SIZE: usize = 4 * 1024;
+
 
 /// Descriptive error type
 #[cfg(feature = "std")]
@@ -107,9 +115,11 @@ pub trait Input {
 		self.read(&mut buf[..])?;
 		Ok(buf[0])
 	}
+
+	/// Minimum length of input, can be used to bound preallocation.
+	fn min_len(&mut self) -> usize;
 }
 
-#[cfg(not(feature = "std"))]
 impl<'a> Input for &'a [u8] {
 	fn read(&mut self, into: &mut [u8]) -> Result<(), Error> {
 		if into.len() > self.len() {
@@ -119,6 +129,10 @@ impl<'a> Input for &'a [u8] {
 		into.copy_from_slice(&self[..len]);
 		*self = &self[len..];
 		Ok(())
+	}
+
+	fn min_len(&mut self) -> usize {
+		self.len()
 	}
 }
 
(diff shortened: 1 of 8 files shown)


thiolliere:
also this API can be error prone as someone can easily forget to call decode_with_preallocated and bypassing previously allocated stuff. though the error seems not critical actually

arkpar:
What's the point? Is this to fight the case when declared size is actually bigger than the stream size?

I'd suggest just adding an optional function to the `Input` trait that would return remaining stream size and checking against that.

> input is not longer implemented for Read because min_len is very not accurate, I can still do a wrapper though

Not sure what `min_len` has to do with it, but we do use an implementation that reads blocks from a file and decodes them one by one in substrate.

thiolliere:
> What's the point? Is this to fight the case when declared size is actually bigger than the stream size?
> 
> I'd suggest just adding an optional function to the `Input` trait that would return remaining stream size and checking against that.

Yes it is for when declared size is bigger than stream size.

Maybe we can do simpler but here is the point:
We want pre-allocation (allocation for data before decoding them) to be limited by input size.
To do so input has now the method min_len, I thought I shouldn't ask for total len because some implementation wouldn't be able to give the full length of input, for instance Read would not.

Then the caller give what it actually pre-allocate so that the total pre-allocated is less than `Input::min_len().max(DEFAULT_MIN_PREALLOCATION)`
If I don't do that all structure that preallocate can do it up to the limit which means `Vec<Vec<Vec<u32>>>` can allocate `3*limit` on an input of size `3*Compact(u32::max_value).encode().len()`.

Maybe this nested stuff is actually OK, and so we can get rid of this preallocated parameter in decode_inner.
 
> > input is not longer implemented for Read because min_len is very not accurate, I can still do a wrapper though
> 
> Not sure what `min_len` has to do with it, but we do use an implementation that reads blocks from a file and decodes them one by one in substrate.

This is for optimisation: we can't get the length of input from the trait Read. Thus implementation of Read will return `Input::min_len() = 0` so the bound would be `preallocation <= DEFAULT_PREALLOCATION` instead of `preallocation <= DEFAULT_PREALLOCATION.max(actual length of input)`

thiolliere:
something we can do that is simple is:

* Input has a new method which is `min_len() -> usize { 0 }`
* implementation of Vec will allocate up to `Input::min_len().max(DEFAULT_PREALLOCATION)`.

The result of it would be nested structure can allocation more than `Input::min_len().max(DEFAULT_PREALLOCATION)` for instance `Vec<Vec<Vec<u32>>>` can allocation `3*limit` with a very small input.

arkpar:
`Read + Seek` can give you stream size. `Read` alone is not that important. I think it makes more sense to limit the input, rather than try and manage such an internal detail as preallocation size. We could just require the stream to be size limited instead.

thiolliere:
> Read + Seek can give you stream size. Read alone is not that important.

OK

> I think it makes more sense to limit the input, rather than try and manage such an internal detail as preallocation size. We could just require the stream to be size limited instead.

I don't understand, in substrate input are limited currently.

src/codec.rs
```diff
@@ -253,8 +258,16 @@ pub trait Decode: Sized {
 	#[doc(hidden)]
 	const IS_U8: IsU8 = IsU8::No;
 
+	/// Attempt to deserialise the value from input while having already preallocated space.
+	///
+	/// Use this when you want to decode while having already preallocated space for later data in
+	/// the input. This allow preallocated space to be bounded by input len or a default maximum.
+	fn decode_inner<I: Input>(input: &mut I, preallocated: usize) -> Result<Self, Error>;
```
bkchr: I don't understand this? What is preallocated size? Given your implementation this is just an upper bound.

I thought we would solve this by using the length of the input?

thiolliere: preallocated is the allocation the caller made for future data in input.

in a Vec I call decode of an element with `preallocated = (vec.capacity() - vec.len() - 1) * size_of::<T>()`

that means that inside the decoding of this element it will be aware the calle has already preallocated some space, thus if we want to guarantee `preallocated <= input_len` the decode knows that he only have `input_len - preallocated` available to pre allocate.

in decode of Vec<T> it is legal to pre-allocated up to:
```rust
let max_for_current_vec_preallocation = input.min_len().max(DEFAULT_MAX_DECODE_PREALLOCATION_SIZE)
	.saturating_sub(preallocated_by_caller);
```

thiolliere: note that this could have been done the other way around:

> caller send to decode what is the maximum pre-allocation callee is allowed, callee will have to respect this number.
This would be done with having decode(input, Some(maximum)) allowing user to specify a maxium to preallocate or no maximum if wanted. Actually this might be better, maybe trusted input would be without limit on preallocation.

But all in all that doesn't change that much, it will just be like:
`decode_inner(input, pre_allocated_limit: usize)`

thiolliere:
let's go in more easier implementation here https://github.com/paritytech/parity-scale-codec/pull/139

PR #139, opened 2019-07-24
Bound preallocation to input len

thiolliere:
* input is no longer implemented for Read but only for &[u8] and also `
struct IoReader<R: std::io::Read + std::io::Seek>(R);` because compiler say std could choose to implement Seek on `&[u8]`.
Also as I remember in substrate we also import block from stdin. I'm not sure stdin implement seek.

* input has a new method which is remaining_len (it is not named len to avoid shadowing len method on &[u8]) which returns remaining_len. This is quite a strong constraint added to input, instead we could just add min_remaining_len and implementor can return less than actual len.

* the only structure that preallocate is Vec thus it is the only one changed

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -156,8 +156,9 @@ impl From<std::io::Error> for Error {
 	}
 }
 
+/// Wrapper that implements Input for any `Read` and `Seek` type.
 #[cfg(feature = "std")]
-struct IoReader<R: std::io::Read + std::io::Seek>(R);
+struct IoReader<R: std::io::Read + std::io::Seek>(pub R);
 
 #[cfg(feature = "std")]
 impl<R: std::io::Read + std::io::Seek> Input for IoReader<R> {
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -575,13 +575,13 @@ impl<T: Decode> Decode for Vec<T> {
 
 			if let IsU8::Yes = <T as Decode>::IS_U8 {
 				let r = if len <= max_preallocation {
-					// if len ok for preallocation then preallocate and read the slice
+					// If len is ok for preallocation then preallocate and read the slice.
 					let mut r = vec![0; len];
 
 					input.read(&mut r[..len])?;
 					r
 				} else {
-					// if len is considered too much for preallocation then use dynamic allocation
+					// If len is considered too much for preallocation then use dynamic allocation.
 					let mut r = Vec::new();
 					let mut remains = len;
 					let buffer_len = max_preallocation;
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -95,6 +95,9 @@ impl From<&'static str> for Error {
 
 /// Trait that allows reading of data into a slice.
 pub trait Input {
+	/// Return remaining length of input.
+	fn remaining_len(&mut self) -> Result<usize, Error>;
+
 	/// Read the exact number of bytes required to fill the given buffer.
 	///
 	/// Note that this function is similar to `std::io::Read::read_exact` and not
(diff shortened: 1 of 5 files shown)


thiolliere:
in this implementation Vec<Vec<_>> can allocate input_len*2 with an input of len 2

src/compact.rs
```diff
@@ -46,6 +46,10 @@ struct PrefixInput<'a, T> {
 }
 
 impl<'a, T: 'a + Input> Input for PrefixInput<'a, T> {
+	fn remaining_len(&mut self) -> Result<usize, Error> {
+		Ok(self.input.remaining_len()?.saturating_add(1))
```
bkchr: Why add one here?

src/codec.rs
```diff
@@ -150,10 +156,23 @@ impl From<std::io::Error> for Error {
...
+		let len = self.0.seek(std::io::SeekFrom::End(0))?
+			.saturating_sub(self.0.seek(std::io::SeekFrom::Current(0))?);
+
+		Ok(len.try_into().unwrap_or(usize::max_value()))
+	}
+
 	fn read(&mut self, into: &mut [u8]) -> Result<(), Error> {
-		(self as &mut dyn std::io::Read).read_exact(into)?;
+		self.0.read_exact(into)?;
 		Ok(())
```
```suggestion
```

src/codec.rs
```diff
@@ -150,10 +156,23 @@ impl From<std::io::Error> for Error {
...
+
+		let len = self.0.seek(std::io::SeekFrom::End(0))?
+			.saturating_sub(self.0.seek(std::io::SeekFrom::Current(0))?);
+
+		Ok(len.try_into().unwrap_or(usize::max_value()))
+	}
+
 	fn read(&mut self, into: &mut [u8]) -> Result<(), Error> {
-		(self as &mut dyn std::io::Read).read_exact(into)?;
+		self.0.read_exact(into)?;
```
```suggestion
		self.0.read_exact(into).map_err(Into::into)
```

src/codec.rs
```diff
@@ -553,14 +572,37 @@ impl<T: Decode> Decode for Vec<T> {
...
 		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 			let len = len as usize;
-			if let IsU8::Yes = <T as Decode>::IS_U8 {
-				let mut r = vec![0; len];
+			let max_preallocation = input.remaining_len()?;
 
-				input.read(&mut r[..len])?;
-				let r = unsafe { core::mem::transmute::<Vec<u8>, Vec<T>>(r) };
+			if let IsU8::Yes = <T as Decode>::IS_U8 {
+				let r = if len <= max_preallocation {
```
bkchr: If we have an `u8` and the rest length does not match the length we have decoded, the input data is corrupt or someone is trying to fool us.

src/codec.rs
```diff
@@ -150,10 +156,23 @@ impl From<std::io::Error> for Error {
...
+#[cfg(feature = "std")]
+pub struct IoReader<R: std::io::Read + std::io::Seek>(pub R);
+
 #[cfg(feature = "std")]
-impl<R: std::io::Read> Input for R {
+impl<R: std::io::Read + std::io::Seek> Input for IoReader<R> {
+	fn remaining_len(&mut self) -> Result<usize, Error> {
+		use std::convert::TryInto;
+
+		let len = self.0.seek(std::io::SeekFrom::End(0))?
```
bkchr: This is not correct. You need to implement it as shown here: https://doc.rust-lang.org/src/std/io/mod.rs.html#1398

src/compact.rs
thiolliere: I forget prefix is optional :-/, fixed

src/codec.rs
thiolliere: oh yes

src/codec.rs
thiolliere: fixed

Issue #93, opened 2019-06-03
Introduce maximum_size() to implement size_hint for set of values


thiolliere:
When implementing size_hint for a vector or a set we don't want to iterate on all the values to actually get a hint.
One way to do so would be to have two method one taking a value `size_hint(&self)` and another one taking no value `maxium_size()`.

thus they would be used as such:

* maximum_size() used on set of values
* size_hint(&self) used on individual values,

size_hint(&self) (thought maybe renamed) would default implementation to maximum_size()

PR #140, opened 2019-07-25
Improve `EncodeAppend` api and clean up code

bkchr:


src/encode_append.rs
```diff
@@ -0,0 +1,168 @@
...
+		// already encoded data.
+		if encoded_len == encoded_new_len {
+			replace_len(&mut self_encoded);
+			append_new_elems(&mut self_encoded);
+
+			Ok(self_encoded)
+		} else {
+			let size = encoded_new_len + self_encoded.len() - encoded_len;
+
+			let mut res = Vec::with_capacity(size + input_len * mem::size_of::<T>());
```
thiolliere: mem::size_of::<T>() is different than encoded size of the instance of T. But this hint is good enough maybe and can be improved with https://github.com/paritytech/parity-scale-codec/issues/93

bkchr: The problem is, I can not iterate the same `iter` twice :(

PR #141, opened 2019-07-25
Add `EncodeLike`

bkchr:


src/encode_like.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl<T: Encode> EncodeLike<[T]> for BinaryHeap<T> {}
+
+impl<K: Encode, V: Encode> EncodeLike<[(K, V)]> for BTreeMap<K, V> {}
+
+#[cfg(feature = "std")]
+impl EncodeLike<str> for String {}
+
+#[cfg(feature = "std")]
+impl EncodeLike<String> for &'static str {}
```
thiolliere: why not 
```suggestion
impl<'a> EncodeLike<String> for &'a str {}
```

thiolliere:
it could also derive every thing that is derived with WrapperTypeEncode

bkchr:
I could not get it working as I wanted, so I will close this for the moment.

PR #143, opened 2019-07-26
Update byte-slice-cast version to support no_std

DarkEld3r:
Port #142 to the `parity-codec-v4` branch.


thiolliere:
do you need an updated version now ? I don't know but I usually bump version when releasing the new version. but I don't mind

thiolliere:
oh I though it was merged on master as well.

PR #146, opened 2019-07-29
New release

thiolliere:
add some EncodeLike,

bump verions. Once merge I release, tag and will update substrate deps

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "parity-scale-codec"
 description = "SCALE - Simple Concatenating Aggregated Little Endians"
-version = "1.0.2"
+version = "1.0.3"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "Apache-2.0"
 repository = "https://github.com/paritytech/parity-scale-codec"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -11,7 +11,7 @@ edition = "2018"
 [dependencies]
 arrayvec = { version = "0.4", default-features = false, features = ["array-sizes-33-128", "array-sizes-129-255"] }
 serde = { version = "1.0", optional = true }
-parity-scale-codec-derive = { path = "derive", version = "1.0", default-features = false, optional = true }
+parity-scale-codec-derive = { path = "derive", version = "^1.0.2", default-features = false, optional = true }
```
DarkEld3r: Just curious: are there some differences between `^1.0.2` and `1.0.2`?
Cargo book [says](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) the following:
> The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

thiolliere: oh ok indeed it seems not

PR #149, opened 2019-08-05
impl optional len

thiolliere:


src/codec.rs
```diff
@@ -96,8 +98,12 @@ impl From<&'static str> for Error {
 
 /// Trait that allows reading of data into a slice.
 pub trait Input {
-	/// Return remaining length of input.
-	fn remaining_len(&mut self) -> Result<usize, Error>;
+	/// If input has a length then return the remaining length of input.
+	/// Otherwise return None.
+	///
+	/// This length is used to constrained decoding allocation, thus returning None can slow down
+	/// performances.
```
```suggestion
```

src/codec.rs
```diff
@@ -96,8 +98,12 @@ impl From<&'static str> for Error {
 
 /// Trait that allows reading of data into a slice.
 pub trait Input {
-	/// Return remaining length of input.
-	fn remaining_len(&mut self) -> Result<usize, Error>;
+	/// If input has a length then return the remaining length of input.
```
```suggestion
	/// Should return the remaining length of the input data. If no information about the input length is available, `None` should be returned.
	///
	/// The length is used to constrain the preallocation while decoding. Returning a garbage length can open the doors for a denial of service attack to your application. 
	/// Otherwise, returning `None` can decrease the performance of your application.
```

src/codec.rs
```diff
@@ -14,6 +14,8 @@
 
 //! Serialisation.
 
+const MAX_PREALLOCATION: usize = 4*1024;
```
```suggestion
const MAX_PREALLOCATION: usize = 4 * 1024;
```

Ann please add a comment

src/codec.rs
```diff
@@ -179,6 +185,7 @@ impl<R: std::io::Read + std::io::Seek> Input for IoReader<R> {
 		len.saturating_sub(old_pos)
 			.try_into()
 			.map_err(|_| "Input cannot fit into usize length".into())
+			.map(|len| Some(len))
```
```suggestion
			.map(Some)
```

src/codec.rs
```diff
@@ -619,16 +626,47 @@ impl<T: Decode> Decode for Vec<T> {
...
+				// If there is input len and it cannot be pre-allocated then return directly.
+				if input_len.map(|l| l < len).unwrap_or(false) {
+					return Err("Not enough data to decode vector".into())
 				}
 
-				let mut r = vec![0; len];
-				input.read(&mut r)?;
+				// Note: we checked that if input_len is some then it can preallocated.
+				let r = if input_len.is_some() || len < MAX_PREALLOCATION {
+
```
```suggestion
```

src/codec.rs
```diff
@@ -619,16 +626,47 @@ impl<T: Decode> Decode for Vec<T> {
...
+					// Here we pre-allocate the whole buffer.
+					let mut r = vec![0; len];
+					input.read(&mut r)?;
+
+					r
+				} else {
+
```
```suggestion
```

src/compact.rs
```diff
@@ -46,8 +46,14 @@ struct PrefixInput<'a, T> {
 }
 
 impl<'a, T: 'a + Input> Input for PrefixInput<'a, T> {
-	fn remaining_len(&mut self) -> Result<usize, Error> {
-		Ok(self.input.remaining_len()?.saturating_add(self.prefix.iter().count()))
+	fn remaining_len(&mut self) -> Result<Option<usize>, Error> {
+		let len = if let Some(len) = self.input.remaining_len()? {
```
bkchr:
```suggestion
		let len = self.input.remaining_len()?.unwrap_or(0);
		Ok(Some(len.saturating_add(self.prefix.iter().count()))
```

src/codec.rs
```diff
@@ -619,16 +626,47 @@ impl<T: Decode> Decode for Vec<T> {
...
+					let mut r = vec![];
+
+					let mut remains = len;
+					let buffer_len = MAX_PREALLOCATION;
+					let mut buffer = vec![0; buffer_len];
+
+					while remains != 0 {
+						let len_read = buffer_len.min(remains);
+						input.read(&mut buffer[..len_read])?;
+						r.extend_from_slice(&buffer[..len_read]);
```
bkchr: If you would use `resize`, you would not need `buffer`.

src/compact.rs
thiolliere: hmm not really if input returns none then returning 0 Some(0) or Some(1) is wrong. indeed we don't use it in compact but still.

src/codec.rs
thiolliere: done

PR #152, opened 2019-08-24
Remove unsafe for endian and non-endian

thiolliere:
## Context

We found an issue when using in ink for decoding u128. Removing the unsafe solved it (I actually tested two times but I can't guarantee entirely I was sleepy yesterday).

## Done

This removed the unsafe for endian and non-endian types (i.e. boolean and integer). performance are probably not on par as it involves more copy now.

note: boolean decoding is implemented this way to ensure uniqueness of encoding.

## Comment

I'm not sure we want to merge it, though here it is.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -865,7 +865,7 @@ mod inner_tuple_impl {
 }
 
 macro_rules! impl_endians {
-	( $( $t:ty => $size:expr),* ) => { $(
+	( $( $t:ty ),* ) => { $(
 		impl EncodeLike for $t {}
 
 		impl Encode for $t {
@@ -881,7 +881,7 @@ macro_rules! impl_endians {
 
 		impl Decode for $t {
 			fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-				let mut buf = [0u8; $size];
+				let mut buf = [0u8; mem::size_of::<$t>()];
 				input.read(&mut buf)?;
 				Ok(<$t>::from_le_bytes(buf))
 			}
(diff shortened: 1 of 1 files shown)


thiolliere:
maybe something would be to use this branch in ink while we commit on how to solve this.

thiolliere:
It doesn't even solve the issue we find in ink.

bkchr:
Ahh, I wanted to comment on this! Just forgot about it. I think that your solution did not looked that bad and I would have accepted it.
I don't think that the performance would decrease with your solution.

What is the problem in ink?

thiolliere:
> I don't think that the performance would decrease with your solution.

Hmm indeed on how is from_le_bytes implemented performance can be same (and quite probably is).
```rust
				let mut buf = [0u8; $size];
				input.read(&mut buf)?;
				Ok(<$t>::from_le_bytes(buf))
```

> What is the problem in ink?

actually a rustc (or more probably llvm) compilation error for webassembly backend (rust issue 63918)

thiolliere:
what do you think about the boolean implementation, it is actually a breaking change but I think it is more like a bug fix.

bkchr:
> what do you think about the boolean implementation, it is actually a breaking change but I think it is more like a bug fix.

I looked at the code, but how does the encoding changes? I don't see it?

thiolliere:
previously if input was [2] or [3] then deocding a bool would return true, now it returns err.
now only [1] returns true and [2] returns Err.

Like I remembered in one PR for compact decoding it was mentionned that unicity of decoding is important for the trie. I think cheme told so.

```rust
use std::mem;
use std::slice;

fn main() {
    let input = [2];
	let mut val: bool = unsafe { mem::zeroed() };

	unsafe {
		let raw: &mut [u8] = slice::from_raw_parts_mut(
			&mut val as *mut bool as *mut u8,
			1
		);
		raw.copy_from_slice(&input[..]);
	}
	
	println!("{}", val)
}
```

bkchr:
Yeah okay, we only always encoded it as `1` or `0`. So, I think that your fix is right and does not break any running code :)

thiolliere:
also it was in the specification of scale-codec

thiolliere:
hmm actually it might break substrate let's investigate.

Issue #156, opened 2019-09-02
Limitation of EncodeLike


thiolliere:
Currently EncodeLike has some limitation:
```
the trait `encode_like::EncodeLike<u32>` is not implemented for `std::boxed::Box<&u32>`
```

indeed what it offers is being able to have one level of reduction for most types and also one level of reduction for every element of a tuple. Or rephrased:
* Box<T> does encode like T
* &T does encode like T
* but &Box<T> does not encode like T
* also for each element of a tuple (&u32, Box<u32>) does encode like (u32, u32) but (&Box<u32>, u32) does not.

At some point this is already helpful but if someone really wants to make use of such EncodeLike trait for edge cases the following pattern can be used:
```rust
[derive(Deref)]
struct Wrapper((Box<Box<u32>>, u32));

impl EncodeLike<(u32, u32)> for Wrapper {}
impl crate::codec::WrapperTypeEncode for Wrapper {}
```

I think at some point I should add this to the doc.

thiolliere:
also as a note I also thought of that but this seems quite not very handy:

```rust
pub trait ReduceLike<T> {}
impl<T: ReduceLike<TL>, TL> ReduceLike<TL> for &T {}
impl<A: ReduceLike<AL>, AL, B: ReduceLike<BL>, BL> ReduceLike<(AL, BL)> for (A, B) {}
impl<T: ReduceLike<TL>, TL> ReduceLike<TL> for Box<T> {}

impl ReduceLike<u32> for u32 {}

fn store_like<Like, Store: ReduceLike<Like>, By: ReduceLike<Like>>() {}

#[cfg(test)]
mod tests {
	#[test]
	fn it_works() {
		// crate::store_like::<_, (u32, u32), (&Box<&u32>, u32)>();
		crate::store_like::<_, (u32, Box<&u32>), (&Box<&u32>, u32)>();
	}
}
```

PR #157, opened 2019-09-02
Improve doc encodelike and implement some more.

thiolliere:
Fix https://github.com/paritytech/parity-scale-codec/issues/156

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -297,12 +297,16 @@ pub trait WrapperTypeEncode: Deref {}
 impl<T: ?Sized> WrapperTypeEncode for Box<T> {}
 impl<T: ?Sized + Encode> EncodeLike for Box<T> {}
 impl<T: Encode> EncodeLike<T> for Box<T> {}
+impl<T: Encode> EncodeLike<Box<T>> for T {}
 
-impl<'a, T: ?Sized> WrapperTypeEncode for &'a T {}
-impl<'a, T: Encode> EncodeLike for &'a T {}
+impl<T: ?Sized> WrapperTypeEncode for &T {}
+impl<T: Encode> EncodeLike<T> for &T {}
+impl<T: Encode> EncodeLike<&T> for T {}
 
-impl<'a, T: ?Sized> WrapperTypeEncode for &'a mut T {}
-impl<'a, T: Encode> EncodeLike for &'a mut T {}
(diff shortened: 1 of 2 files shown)


bkchr:
```
diff --git a/src/codec.rs b/src/codec.rs
index f67b1b8..9cecab7 100644
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -297,6 +297,7 @@ pub trait WrapperTypeEncode: Deref {}
 impl<T: ?Sized> WrapperTypeEncode for Box<T> {}
 impl<T: ?Sized + Encode> EncodeLike for Box<T> {}
 impl<T: Encode> EncodeLike<T> for Box<T> {}
+impl<T: Encode> EncodeLike<T> for Box<&T> {}
 
 impl<'a, T: ?Sized> WrapperTypeEncode for &'a T {}
 impl<'a, T: Encode> EncodeLike for &'a T {}
diff --git a/src/encode_like.rs b/src/encode_like.rs
index 9dcd4fc..6b74c37 100644
--- a/src/encode_like.rs
+++ b/src/encode_like.rs
@@ -83,6 +83,15 @@ mod tests {
                assert_eq!(slice_encoded, data_encoded);
        }
 
+       #[test]
+       fn test_box() {
+               let data = 10u32;
+               let encoded = data.encode();
+               let encoded2 = ComplexStuff::<Box<&u32>>::complex_method(&10u32);
+
+               assert_eq!(encoded, encoded2);
+       }
+
        #[test]
        fn interface_testing() {
                let value = 10u32;
```

It works?

thiolliere:
oh yes indeed then we can implement way more than just those I try to come up with something. it is just about iterating on lots of combination

bkchr:
I would more like mention in the documentation that we probably missed some combinations and accept prs to add these new combinations. And you could mention that this is not "design limitation", it is a flaw of rust that does not supports specialization..

thiolliere:
though the number of combination seems quite explosive like combinations such as
```rust
impl<A: Encode, B: Encode> EnodeLike<(A, B)> for Box<(A, &B)> {}
```
should be written. thus if we want to support an ok set of tuple the number of combination might be huge.

thiolliere:
also if a user defined two struct say that on encode like the other he has to implement all the combination of them inside box and tuple and vec and reference as well.

But might be an ok solution, do you have an idea of the number of implementation rustc can handle ?

PR #158, opened 2019-09-02
Add method to EncodeAppend to append with EncodeLike items

thiolliere:
implement EncodeLikeAppend which is EncodeAppend but using EncodeLike values.

It seems we can't derive EncodeAppend for EncodeLikeAppend automatically because some type can have EncodeLike not implemented and thus the `Iterator<EncodeLike<Item>>` cannot be converted into `Iterator<Item>`.

And if I derive only for types where `Item: EncodeLike<Item>` then it is a breaking change as EncodeAppend is no longer implemented for Item which doesn't have this bound.

I'm not sure if we should deprecate things ?
And naming is not perfect either.

### DIFF
--- a/src/encode_append.rs
+++ b/src/encode_append.rs
@@ -19,16 +19,20 @@ use crate::alloc::vec::Vec;
 
 /// Trait that allows to append items to an encoded representation without
 /// decoding all previous added items.
-///
-/// Note: EncodeLikeAppend should be prefered as it allows to append with any type that encode like
-/// `Item`.
 pub trait EncodeAppend {
 	/// The item that will be appended.
 	type Item: Encode;
 
-	/// Append all items in `iter` to the given `self_encoded` representation.
-	///
-	/// Except if `self_encoded` value is empty then it just insert the given input data.
+	/// Append all items in `iter` to the given `self_encoded` representation
+	/// or if `self_encoded` value is empty then insert the given input data.
+	#[deprecated(note = "Consider using `append_or_new` instead")]
+	fn append<'a, I: IntoIterator<Item=&'a Self::Item>>(
+		self_encoded: Vec<u8>,
+		iter: I,
+	) -> Result<Vec<u8>, Error> where Self::Item: 'a, I::IntoIter: ExactSizeIterator;
+
+	/// Append all items in `iter` to the given `self_encoded` representation
+	/// or if `self_encoded` value is empty then insert the given input data.
 	///
 	/// # Example
 	///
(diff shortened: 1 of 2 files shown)


bkchr:
I think we should add a method `append_new` to the `EncodeAppend` trait. `append` just calls the new `append_new`. As we are the only one using this, we hopefully don't break any downstream code. In a future breaking release, we would replace `append` with `append_new`. 

MAYBE, when we can make sure to not break Substrate/Polkadot, we could try to just replace `append` directly. For this we would need to check that substrate still builds with the new  `append` function.

thiolliere:
ya not possible to replace append like this:
```diff
@@ -41,10 +41,14 @@ pub trait EncodeAppend {
        /// // Add multiple element
        /// <Vec<u32> as EncodeAppend>::append(encoded, &[700u32, 800u32, 10u32]).expect("Adds new elements");
        /// ```
-       fn append<'a, I: IntoIterator<Item=&'a Self::Item>>(
+       fn append<'a, I, EncodeLikeItem>(
                self_encoded: Vec<u8>,
                iter: I,
-       ) -> Result<Vec<u8>, Error> where Self::Item: 'a, I::IntoIter: ExactSizeIterator;
+       ) -> Result<Vec<u8>, Error>
+       where
+               I: IntoIterator<Item=&'a EncodeLikeItem>,
+               EncodeLikeItem: 'a + EncodeLike<Self::Item>,
+               I::IntoIter: ExactSizeIterator;
 }
 ```

Because EncodeLike is not implemented for all type that implement Encode.

we can do this but it loose all interest as we couldn't happen with reference easily anymore:
```diff
@@ -41,10 +41,14 @@ pub trait EncodeAppend {
        /// // Add multiple element
        /// <Vec<u32> as EncodeAppend>::append(encoded, &[700u32, 800u32, 10u32]).expect("Adds new elements");
        /// ```
-       fn append<'a, I: IntoIterator<Item=&'a Self::Item>>(
+       fn append<I, EncodeLikeItem>(
                self_encoded: Vec<u8>,
                iter: I,
-       ) -> Result<Vec<u8>, Error> where Self::Item: 'a, I::IntoIter: ExactSizeIterator;
+       ) -> Result<Vec<u8>, Error>
+       where
+               I: IntoIterator<Item=EncodeLikeItem>,
+               EncodeLikeItem: EncodeLike<Self::Item>,
+               I::IntoIter: ExactSizeIterator;
 }
```

So I'll for another method, maybe name it append_with, I feel like append_new has some semantic meaning about creating when empty.

src/encode_append.rs
```diff
@@ -41,10 +48,37 @@ pub trait EncodeAppend {
...
+	) -> Result<Vec<u8>, Error> where Self::Item: 'a, I::IntoIter: ExactSizeIterator {
+		append_or_new_vec_with_any_item(self_encoded, iter)
+	}
+
+	fn append_or_new<'a, EncodeLikeItem, I>(
+		self_encoded: Vec<u8>,
+		iter: I,
+	) -> Result<Vec<u8>, Error>
+	where
+		I: IntoIterator<Item = &'a EncodeLikeItem>,
```
thiolliere: we do use a reference here because as we want to be able to append while not consuming value for instance `&([u32], u32)` implements `&EncodeLike<(Vec<u32>, u32)>` but not `EncodeLike<&(Vec<u32>, u32)>`

src/encode_append.rs
```diff
@@ -59,68 +93,70 @@ fn extract_length_data(data: &[u8], input_len: usize) -> Result<(u32, usize, usi
...
-			Ok(self_encoded)
-		} else {
-			let size = encoded_new_len + self_encoded.len() - encoded_len;
+		Ok(self_encoded)
+	} else {
+		let size = encoded_new_len + self_encoded.len() - encoded_len;
 
-			let mut res = Vec::with_capacity(size + input_len * mem::size_of::<T>());
-			unsafe { res.set_len(size); }
+		let mut res = Vec::with_capacity(size + input_len * mem::size_of::<Item>());
```
thiolliere: here is the only change of current behavior the hint size can be different but this is fine.

thiolliere:
also another note: I think the general design of EncodeAppend could have been better without all this where clause on the function. instead we could have put some generic parameter and put those where clause in the implementation of the trait instead. doing so would have made the use of this trait more easier, because at the moment we have to repeat all where clause everywhere we want to use this trait

thiolliere:
Also at some point I kind of prefer if the signature of the method was: `function(Option<encoded_value>, [item])` instead of just a vec which can be empty. Also at some point it seems that having the ability to encode to an empty value is not necessary.

But I'm ok with this current state, just if we want to rewrite a bit while introducing EncodeLike we could create a new trait with no where method and an Option of just not this feature.

bkchr:
Which where bound to you want to put at the implementation? You can not put a where bound for `I` into the implementation, that does not work.

And I don't see any reason for having `Option<Vec<u8>>`, what would be the benefit? Knowing that the value does not exists, which is the same as `Vec<u8>.len() == 0`?

thiolliere:
> Which where bound to you want to put at the implementation? You can not put a where bound for `I` into the implementation, that does not work.

Yes I was thinking something like this:
```rust
pub trait EncodeAppend3<I> {
	fn append_or_new(
		self_encoded: Vec<u8>,
		iter: I,
	) -> Result<Vec<u8>, Error>;
}

impl<'a, I, Item, T> EncodeAppend3<I> for Vec<T>
where
	T: Encode,
	I: IntoIterator<Item = &'a Item>,
	Item: 'a + EncodeLike<T>,
	I::IntoIter: ExactSizeIterator,
{
	fn append_or_new(
		self_encoded: Vec<u8>,
		iter: I,
	) -> Result<Vec<u8>, Error>
	{
		append_or_new_vec_with_any_item(self_encoded, iter)
	}
}

pub trait StorageValue<V: crate::Codec> {
	fn append<Iter>(items: Iter) -> Result<(), &'static str>
	where Self: EncodeAppend3<Iter>
	{
		unimplemented!();
	}
}
```

> And I don't see any reason for having `Option<Vec<u8>>`, what would be the benefit? Knowing that the value does not exists, which is the same as `Vec<u8>.len() == 0`?

Yes it is just to make the API more clear like self_encoded is not self_encoded it is actually self_encoded or empty_vec. but I'm ok

bkchr:
Hmm yeah, could work. If you want you can implement it. Then we would need to release a new breaking release.

Could be done with your other proposed improvements as well.

thiolliere:
close in favor of https://github.com/paritytech/parity-scale-codec/pull/161

PR #162, opened 2019-09-19
Update dependencies

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

# What I have changed

- Update dependencies
- Remove Cargo.lock (parity-scale-codec is a library)
- Reorder some imports (just my habit)


src/generic_array.rs
```diff
@@ -41,7 +41,8 @@ impl<T: Decode, L: generic_array::ArrayLength<T>> Decode for generic_array::Gene
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use generic_array::{GenericArray, arr, arr_impl};
+	use generic_array::typenum;
+	use generic_array::{GenericArray, arr};
```
thiolliere: those two line could be mergede

derive/src/encode.rs
```diff
@@ -14,14 +14,12 @@
...
-	punctuated::Punctuated,
-	spanned::Spanned,
-	token::Comma,
-	Error,
-};
+use proc_macro2::{Ident, Span, TokenStream};
+use syn::punctuated::Punctuated;
+use syn::spanned::Spanned;
+use syn::token::Comma;
+use syn::{Data, Field, Fields, Error};
```
thiolliere: Actually I think the grouped way is preferred. at least it seems to be the pattern used in substrate

Cargo.toml
```diff
@@ -12,16 +12,14 @@ edition = "2018"
 arrayvec = { version = "0.4", default-features = false, features = ["array-sizes-33-128", "array-sizes-129-255"] }
 serde = { version = "1.0", optional = true }
 parity-scale-codec-derive = { path = "derive", version = "^1.0.2", default-features = false, optional = true }
-bitvec = { version = "0.14.0", default-features = false, features = ["alloc"], optional = true }
+bitvec = { version = "0.15", default-features = false, features = ["alloc"], optional = true }
```
thiolliere: this is a breaking change as 0.14 -> 0.15 is breaking change.

Cargo.toml
```diff
@@ -12,16 +12,14 @@ edition = "2018"
...
 byte-slice-cast = { version = "0.3.2", default-features = false, optional = true }
-generic-array = { version = "0.12", optional = true }
+generic-array = { version = "0.13", optional = true }
```
thiolliere: this is also a breaking change

derive/src/decode.rs
```diff
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 use proc_macro2::{Span, TokenStream, Ident};
-use syn::{Data, Fields, Field, spanned::Spanned, Error};
+use syn::spanned::Spanned;
```
bkchr: Please revert.

derive/Cargo.toml
```diff
@@ -10,11 +10,7 @@ edition = "2018"
...
-proc-macro2 = "0.4"
-proc-macro-crate = "0.1.3"
-
-[features]
-default = ["std"]
-std = []
+syn = { version = "1.0", features = [ "full", "visit" ] }
+quote = "1.0"
+proc-macro2 = "1.0"
+proc-macro-crate = "0.1"
```
bkchr: Incorrect. There was a reason why we required `0.1.3`.

Cargo.toml
bkchr: You mean they changed the internal representation we use to serialize?

thiolliere: I havn't looked at this but probably not. My point was just that this change introduce a breaking change in parity-scale-codec.

because a major update of an exposed dependency is a breaking change. user of your libraries will have to make this major update for this dependency as well for their code not to break.

thiolliere: indeed that might be difficult to keep up with bitvec and generic-array if we have to release major release each time we want to update them.

But we can probably implement for both version if needed no ? like:
```toml
bitvec15 = { version = "0.15" }
bitvec16 = { version = "0.16" }
```

src/lib.rs
```diff
@@ -236,6 +236,12 @@
 #[doc(hidden)]
 pub extern crate alloc;
 
+#[cfg(feature = "bit-vec")]
+pub extern crate bitvec;
+
+#[cfg(feature = "generic-array")]
+pub extern crate generic_array;
```
thiolliere: I don't know what are the best practice in regards to reexport ?

Cargo.toml
thiolliere: still this

derive/Cargo.toml
thiolliere: and this

koushiro:
There is a problem, `bitvec` 0.14 with features `alloc` enabled cannot be compiled under the rust stable toolchain. We need to modify `.gitlab-ci.yml` or upgrade the `bitvec` version to 0.15.

thiolliere:
> There is a problem, `bitvec` 0.14 with features `alloc` enabled cannot be compiled under the rust stable toolchain. We need to modify `.gitlab-ci.yml` or upgrade the `bitvec` version to 0.15.

hmm so the test suite is not successful on master :-/

also I'm finally ok to bump bitvec to 0.15 as the only breaking change is minimum compiler version.

koushiro:
> > There is a problem, `bitvec` 0.14 with features `alloc` enabled cannot be compiled under the rust stable toolchain. We need to modify `.gitlab-ci.yml` or upgrade the `bitvec` version to 0.15.
> 
> hmm so the test suite is not successful on master :-/
> 
> also I'm finally ok to bump bitvec to 0.15 as the only breaking change is minimum compiler version.

Already bump `bitvec` to 0.15
All CI checks have been passed, PTAL :)

thiolliere:
Yes ok, changes are good but actually I'm still unsure about bitvec. like if we release a new minor version of parity-scale-codec with this. Then if a user use bitvec feature and does `cargo update -p parity-scale-codec` his code doesn't compile anymore.

but user should easily upgrade bitvec to 0.15 as it is not breaking anything expect minimal rustc. :-/

Robbepop:
What is the state for this PR? Can we merge with major bump, e.g. `1.1.0`? This is the last dependency of ink! that pulls in old `syn`, `quote` and `proc_macro2` dependencies.

PR #164, opened 2019-09-23
Remove the `drop(dest)` and make sure we don't generate a warning

bkchr:


thiolliere:
hmm actually this break substrate with:
```
error[E0453]: allow(unused) overruled by outer forbid(unused_imports)
  --> /home/thiolliere/Developpement/copy-substrate/core/consensus/babe/primitives/src/digest.rs:97:23
   |
97 | #[derive(Copy, Clone, Encode, Decode)]
   |                       ^^^^^^ overruled by previous forbid
   | 
  ::: /home/thiolliere/Developpement/copy-substrate/core/consensus/babe/primitives/src/lib.rs:19:56
   |
19 | #![forbid(unsafe_code, missing_docs, unused_variables, unused_imports)]
   |                                                        -------------- `forbid` level set here

```

PR #166, opened 2019-09-26
Remove allow unused

thiolliere:
I can't find why but inside a `const _ = { inside here }` no usued stuff gets reported. On both stable and nightly: see https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b9e33d172e399cce0ad64426c6f76d8d

I compile parity-scale-codec test and half compiled substrate no unused reported.


thiolliere:
I havn't tried with clippy though

thiolliere:
actually same for clippy if you run clippy in this playground no warning is reported for inside the const: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f46d4386b7eb7fb0201841d67084f04c

PR #188, opened 2020-03-09
Add duration to codec

pscott:
No issue addressed.

This PR adds encode / decode for `std::time::Duration` by using `as_millis()` and treating it as a `u64`.

Note: `as_millis()` returns a u128, but `from_millis(millis: u64)` takes a `u64`. I looked at [this issue](https://github.com/rust-lang/rust/issues/58580) and decided I would use `as u64` after calling `as_millis()`.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -15,7 +15,7 @@
 //! Serialisation.
 
 #[cfg(feature = "std")]
-use std::fmt;
+use std::{fmt, time::Duration};
 use core::{mem, ops::Deref, marker::PhantomData, iter::FromIterator, convert::TryFrom};
 
 use arrayvec::ArrayVec;
@@ -1099,6 +1099,24 @@ impl Decode for bool {
 	}
 }
 
+impl Encode for Duration {
+	fn size_hint(&self) -> usize {
+		mem::size_of::<u64>()
+	}
+
+	fn encode(&self) -> Vec<u8> {
+		let millis = self.as_millis() as u64;
+		millis.encode()
+	}
+}
+
+impl Decode for Duration {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let num = <u64>::decode(input)?;
+		Ok(Duration::from_millis(num))
+	}
+}
+
 #[cfg(test)]
 mod tests {
 	use super::*;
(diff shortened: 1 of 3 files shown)


src/codec.rs
```diff
@@ -1099,13 +1099,31 @@ impl Decode for bool {
...
+impl Encode for Duration {
+	fn size_hint(&self) -> usize {
+		mem::size_of::<u64>()
+	}
+
+	fn encode(&self) -> Vec<u8> {
+		let millis = self.as_millis() as u64;
+		millis.encode()
+	}
+}
```
thiolliere: Encoding of Duration must hold all information about Duration.
Duration is actually containing secs in u64 and nanos in u32 so you should convert to secs and subsecs_nanos to get all information

src/codec.rs
```diff
@@ -1099,13 +1099,32 @@ impl Decode for bool {
...
+		let secs = self.as_secs();
+		let nanos = self.subsec_nanos();
+		(secs, nanos).encode()
+	}
+}
+
+impl Decode for Duration {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let (secs, nanos) = <(u64, u32)>::decode(input)?;
+		Ok(Duration::from_secs(secs) + Duration::from_nanos(nanos as u64))
```
thiolliere: I think would be better to check that nanos is less than a 10^9 in order to make the encoding unique.

(Also if you check this then you can directly use Duration::new)

pscott: Good point!

src/codec.rs
```diff
@@ -1099,13 +1100,36 @@ impl Decode for bool {
...
+		let secs = self.as_secs();
+		let nanos = self.subsec_nanos();
+		(secs, nanos).encode()
+	}
+}
+
+impl Decode for Duration {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let (secs, nanos) = <(u64, u32)>::decode(input)?;
+		if nanos > A_BILLION {
```
thiolliere: if nanos is a billion it must be one second then
```suggestion
		if nanos >= A_BILLION {
```

src/codec.rs
```diff
@@ -1099,13 +1100,36 @@ impl Decode for bool {
...
+impl Decode for Duration {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let (secs, nanos) = <(u64, u32)>::decode(input)?;
+		if nanos >= A_BILLION {
+			Err("Number of nanoseconds should not be higher than 10^9.".into())
+		} else {
+			Ok(Duration::new(secs, nanos))
+		}
+	}
+}
```
thiolliere: oups I forget we have to implement encodelike itself.

```suggestion
}

impl EncodeLike for Duration {}
```

thiolliere:
actually we could also add to the fuzzer

PR #189, opened 2020-03-18
Handle invalid utf8 sequence

pscott:
## Issue
Currently, we're using `from_utf8_lossy` when decoding a `String`. This results in replacing invalid utf8 sequences by the `U+FFFD REPLACEMENT CHARACTER`. This itself introduces a weird behavior:
```rust
#[test]
fn string_decode_encode_test() {
    let mut bytes: &[u8] = &[20, 114, 167, 10, 20, 114];
    let expected = bytes.clone();
    let obj = <String>::decode(&mut bytes).unwrap();
    // obj is the decoded String coming from `bytes`. Since `bytes` contains an invalid utf8 sequence
    // obj now has inserted a replacement character instead in the third position of the array.
    let data: &[u8] = &obj.encode();
    // Comparing the original bytes (expected) to data will fail, because data now has the replacement character.
    assert_eq!(expected, data);
}
```

This PR addresses this issue and uses `from_utf8` instead of `from_utf8_lossy`. It will return an error in case of invalid utf8 sequence.
This PR also adds a testcase to make sure we keep the desired behavior.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -670,7 +670,7 @@ impl<T> Decode for PhantomData<T> {
 #[cfg(any(feature = "std", feature = "full"))]
 impl Decode for String {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-		Self::from_utf8(Vec::decode(input)?).map_err(|_| Error("Invalid utf8 characters"))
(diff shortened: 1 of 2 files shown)


bkchr:
Tests are failing.

pscott:
All good now!

PR #191, opened 2020-03-24
Extend Fuzzer with Encoding

pscott:
Currently, the fuzzer only tries to decode arbitrary bytes.
This PR:
1. Modifies the current fuzzing technique by:
    1. Decoding the arbitrary bytes
    2. Encoding the object obtained from step i
    3. Decoding the bytes from step ii
    4. Making sure that the objects from steps iii and i are the same, and that the bytes of step 2 and the original bytes are the same.
2. Adds a new macro that tests encoding by:
    1. Generating an arbitrary object (among the list of fuzzed types)
    2. Encoding the object from step i
    3. Decoding the bytes from step ii
    4. Comparing the object from step iii with the original arbitrary object

Encoding is tested in (1), however it is only encoding structures that have already been decoded (we might have a bias here).
Decoding is tested in (2), but only from bytes that we have personally encoded.
That is why both functions / macros are needed.

I had to remove `BitVec<BigEndian, u8>` and `Bitvec<BigEndian, u32>` from the (2) Encoding fuzzer, because I could not implement the `Arbitrary` trait for them.

Ideas to further enhance the fuzzer:
- Add `encode_append()` tests
- Add `BitVec<BigEndia, u8>`  and `BitVec<BigEndian, u32>` back to the Encode tests.
- If we wish to improve speed, separate (1) Decoding into two separate macros: one which sorts bytes (for comparing BinaryHeaps, and BTreeSet), and another one for the rest, that does not sort the bytes. Code would run faster, but this would add code complexity.

cc @kirushik

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ parity-scale-codec-derive = { path = "derive", version = "1.2.0", default-featur
 bitvec = { version = "0.15", default-features = false, features = ["alloc"], optional = true }
 byte-slice-cast = { version = "0.3.4", default-features = false, features = ["alloc"] }
 generic-array = { version = "0.13.2", optional = true }
+arbitrary = { version = "0.4.1", features = ["derive"] }
 
 [dev-dependencies]
 criterion = "0.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ parity-scale-codec-derive = { path = "derive", version = "1.2.0", default-featur
 bitvec = { version = "0.15", default-features = false, features = ["alloc"], optional = true }
 byte-slice-cast = { version = "0.3.4", default-features = false, features = ["alloc"] }
 generic-array = { version = "0.13.2", optional = true }
-arbitrary = { version = "0.4.1", features = ["derive"] }
+arbitrary = { version = "0.4.1", features = ["derive"], optional = true }
 
 [dev-dependencies]
 criterion = "0.3.0"
@@ -35,6 +35,7 @@ default = ["std"]
 derive = ["parity-scale-codec-derive"]
 std = ["serde", "bitvec/std", "byte-slice-cast/std"]
 bit-vec = ["bitvec"]
+fuzz = ["std", "arbitrary"]
 
 # WARNING: DO _NOT_ USE THIS FEATURE IF YOU ARE WORKING ON CONSENSUS CODE!*
 #
--- a/fuzzer/src/main.rs
+++ b/fuzzer/src/main.rs
@@ -12,6 +12,7 @@ pub struct MockStruct{
 }
 
 #[derive(Encode, Decode, Debug, Clone, Arbitrary)]
+/// Used for implementing the PartialEq trait for a BinaryHeap.
 struct BinaryHeapWrapper(BinaryHeap<u32>);
 
 impl PartialEq for BinaryHeapWrapper {
--- a/fuzzer/src/main.rs
+++ b/fuzzer/src/main.rs
@@ -16,6 +16,7 @@ struct BinaryHeapWrapper(BinaryHeap<u32>);
 
 impl PartialEq for BinaryHeapWrapper {
 	fn eq(&self, other: &BinaryHeapWrapper) -> bool {
+		// Sort needed because the Iterator returns the value in arbitrary order.
 		let a = self.0.iter().cloned().collect::<Vec<u32>>().sort();
 		let b = other.0.iter().cloned().collect::<Vec<u32>>().sort();
 		a == b
@@ -74,11 +75,15 @@ macro_rules! fuzz_decoder {
 	$(
 		if $data[0] % num == $index {
 			let mut d = &$data[1..];
-			let raw1 = d.clone();
+			// Sorting here is necessary since:
+			// "The order of the elements is not fixed, depends on the container, and cannot be relied on at decoding."
+			// (see https://substrate.dev/docs/en/conceptual/core/codec).
+			let raw1 = d.clone().sort();
 			let maybe_obj = <$parsed>::decode(&mut d);
 			if let Ok(obj) = maybe_obj {
 				let mut d2: &[u8] = &obj.encode();
-				let raw2 = d2.clone();
+				// Sorting here is necessary: see above comment.
+				let raw2 = d2.clone().sort();
 				let exp_obj = <$parsed>::decode(&mut d2);
 				match exp_obj {
 					Ok(obj2) => {
--- a/fuzzer/src/main.rs
+++ b/fuzzer/src/main.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet, VecDeque, LinkedList};
+use std::collections::{BTreeMap, BTreeSet, VecDeque, LinkedList, BinaryHeap};
 use std::time::Duration;
 
 use bitvec::{vec::BitVec, cursor::BigEndian};
@@ -8,6 +8,17 @@ use parity_scale_codec::{Encode, Decode, Compact};
 #[derive(Encode, Decode, PartialEq, Debug)]
 pub struct MockStruct{
 	vec_u: Vec<u8>
+}
+
+#[derive(Encode, Decode, Debug)]
+struct BinaryHeapWrapper(BinaryHeap<u32>);
+
+impl PartialEq for BinaryHeapWrapper {
+	fn eq(&self, other: &BinaryHeapWrapper) -> bool {
+		let a = self.0.iter().cloned().collect::<Vec<u32>>().sort();
+		let b = other.0.iter().cloned().collect::<Vec<u32>>().sort();
+		a == b
+	}
 }
 
 #[derive(Encode, Decode, PartialEq, Debug)]
@@ -124,7 +135,7 @@ fn fuzz_one_input(data: &[u8]){
 		BTreeMap<u8, u8>,
 		BTreeSet<u32>,
 		VecDeque<u8>,
-		// BinaryHeap<u32>,
+		BinaryHeapWrapper,
 		MockStruct,
 		MockEnum,
 		BitVec<BigEndian, u8>,
@@ -138,5 +149,3 @@ fn main() {
 		fuzz!(|data: &[u8]| { fuzz_one_input(data); });
 	}
 }
-
-
--- a/fuzzer/src/main.rs
+++ b/fuzzer/src/main.rs
@@ -76,7 +76,7 @@ macro_rules! fuzz_decoder {
 	$(
 		if $data[0] % num == $index {
 			let mut d = &$data[1..];
-			let mut raw1 = &d.clone();
+			let raw1 = &d.clone();
 			// Sorting here is necessary since:
 			// "The order of the elements is not fixed, depends on the container, and cannot be relied on at decoding."
 			// (see https://substrate.dev/docs/en/conceptual/core/codec).
--- a/fuzzer/src/main.rs
+++ b/fuzzer/src/main.rs
@@ -1,7 +1,7 @@
 use std::collections::{BTreeMap, BTreeSet, VecDeque, LinkedList, BinaryHeap};
 use std::time::Duration;
 
-use bitvec::{vec::BitVec, order::Msb0, bitvec, order::BitOrder, store::BitStore};
+use bitvec::{vec::BitVec, order::Msb0, order::BitOrder, store::BitStore};
 use honggfuzz::fuzz;
 use parity_scale_codec::{Encode, Decode, Compact};
 use honggfuzz::arbitrary::{Arbitrary, Unstructured, Result as ArbResult};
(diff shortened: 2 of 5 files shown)


fuzzer/src/main.rs
```diff
@@ -94,19 +142,75 @@ fn fuzz_one_input(data: &[u8]){
...
+			Option<Vec<u8>>,
+			Vec<u32>,
+			LinkedList<u8>,
+			BTreeMap<String, Vec<u8>>,
+			BTreeMap<u8, u8>,
+			BTreeSet<u32>,
+			VecDeque<u8>,
+			BinaryHeapWrapper,
+			MockStruct,
+			MockEnum,
```
pscott: I've removed the `BitVec<BigEndian, u8>` and `BitVec<BigEndian, u32>` types here because I could not implement Arbitrary for them

Cargo.toml
```diff
@@ -15,6 +15,7 @@ parity-scale-codec-derive = { path = "derive", version = "1.2.0", default-featur
 bitvec = { version = "0.15", default-features = false, features = ["alloc"], optional = true }
 byte-slice-cast = { version = "0.3.4", default-features = false, features = ["alloc"] }
 generic-array = { version = "0.13.2", optional = true }
+arbitrary = { version = "0.4.1", features = ["derive"] }
```
bkchr: This should be hidden behind a feature.

pscott:
So I've removed the [WIP] tag.
This PR would be ready for review, but I have one issue:
With the new (decode, encode, decode) flow I've added, we get a couple of "false-positives" coming from structures such as BTreeSet, and BTreeMap.
Indeed, running this code:
```rust
use parity_scale_codec::alloc::collections::BTreeMap;
use parity_scale_codec::{Decode, Encode};

fn main() {
	let mut bytes: &[u8] = &[4, 0, 16, 0, 72, 56, 212];
	let a = <BTreeMap<String, Vec<u8>>>::decode(&mut bytes).unwrap();
	let b = a.encode();
	println!("encoded: {:?}", b);
}
```
will output:  `[4, 0, 16, 0, 72, 56, 212]`
As you can see, the encoded version of the decoded object is different from the original bytes.
I'm trying to find a way to circumvent this problem (that only happens for BTreeSet and BTreeMap (and for MockEnum by extension)), but the only way out I see is:
- Split the "decoding" tests into two groups:
    1. With BTreeSet, BTreeMap and MockEnum, that will only try to decode arbitrary bytes.
    2. With the rest, that will run the full (decode, encode, decode) workflow

Thoughts @bkchr @thiolliere ?

thiolliere:
another way would be to compare raw_byte::decode() to raw_byte::decode().encode().decode()
but I'm not very sure if fuzzing this is very relevant

bkchr:
`BTreeSet` and `BTreeMap` sort. That is why the encoded version changes.

pscott:
> another way would be to compare raw_byte::decode() to raw_byte::decode().encode().decode()
> but I'm not very sure if fuzzing this is very relevant

Well one of the goal of my PR is to check for this particular case where we could decode a struct,  encode it, and get bytes that are different than the initial ones. I believe we should keep this flow, and simply make a special case for data structures that don't respect this (i.e `BTreeSet` and `BTreeMap`) :)



> `BTreeSet` and `BTreeMap` sort. That is why the encoded version changes.

Yes indeed, and that is why I wrote the comment asking whether we should split the fuzzing in two "groups" : those who sort, and those who don't :)

fuzzer/src/main.rs
bkchr: Put them in a `NewType` wrapper to implement the trait.

fuzzer/src/main.rs
```diff
@@ -1,16 +1,45 @@
...
+impl<O: 'static + BitOrder, T: 'static + BitStore + Arbitrary> Arbitrary for BitVecWrapper<O, T> {
+	fn arbitrary(u: &mut Unstructured<'_>) -> ArbResult<Self> {
+		let v = Vec::<T>::arbitrary(u)?;
+		Ok(BitVecWrapper(BitVec::<O, T>::from_vec(v)))
+	}
+}
+
+
+// #[derive(Encode, Decode, Debug, Clone, Arbitrary)]
+// pub struct BitcVecWrapper<BitVec>(BitVec);
```
DarkEld3r, line 27: Should this commented code be removed?

pscott: My bad code is not ready yet I'm switching back to draft mode :)

pscott:
https://github.com/paritytech/parity-scale-codec/pull/191/commits/952c6c8d525c03008ea8626e26c91af134747814 implements the idea I mentioned in the comments.
We now have three different decoding flow:
1. `round_trip` which decodes, then encodes the obtained object, and compares the bytes with the original bytes.
2. `round_trip_sorted` which decodes, then encodes the obtained objects, and compares the **sorted** bytes with the original **sorted** bytes. Useful for structs like BinaryHeap which have multiple valid encoded forms (of the same length).
3. `only_decode` which only decodes the arbitrary bytes. Useful for more complex structures (like `BTreeMap`, `BTreeSet`) (of different lengths).

fuzzer/src/main.rs
```diff
@@ -1,78 +1,216 @@
...
+}
+
+
+#[derive(Encode, Decode, Debug, Clone, Arbitrary)]
+/// Used for implementing the PartialEq trait for a BinaryHeap.
+struct BinaryHeapWrapper(BinaryHeap<u32>);
+
+impl PartialEq for BinaryHeapWrapper {
+	// Eq here is implemented by creating a vector of all values and then sorting it.
+	// TODO: A better approach would be to use `into_iter_sorted` and compare both iterator, once
```
pscott, line 32: Should I create an issue for this?

fuzzer/src/main.rs
```diff
@@ -1,78 +1,216 @@
...
 
-#[derive(Encode, Decode)]
+#[derive(Encode, Decode, Clone, PartialEq, Debug, Arbitrary)]
 pub struct MockStruct{
 	vec_u: Vec<u8>
 }
 
-#[derive(Encode, Decode)]
+#[derive(Encode, Decode, Clone, PartialEq, Debug)]
+/// Used for implementing the Arbitrary trait for a BitVec.
```
bkchr, line 15:
```suggestion
/// Used for implementing the Arbitrary trait for a BitVec.
#[derive(Encode, Decode, Clone, PartialEq, Debug)]
```

fuzzer/src/main.rs
```diff
@@ -1,78 +1,216 @@
...
+impl<O: 'static + BitOrder, T: 'static + BitStore + Arbitrary> Arbitrary for BitVecWrapper<O, T> {
+	fn arbitrary(u: &mut Unstructured<'_>) -> ArbResult<Self> {
+		let v = Vec::<T>::arbitrary(u)?;
+		Ok(BitVecWrapper(BitVec::<O, T>::from_vec(v)))
+	}
+}
+
+
+#[derive(Encode, Decode, Debug, Clone, Arbitrary)]
+/// Used for implementing the PartialEq trait for a BinaryHeap.
```
bkchr, line 27:
```suggestion
/// Used for implementing the PartialEq trait for a BinaryHeap.
#[derive(Encode, Decode, Debug, Clone, Arbitrary)]
```

fuzzer/src/main.rs
```diff
@@ -1,78 +1,216 @@
...
+struct BinaryHeapWrapper(BinaryHeap<u32>);
+
+impl PartialEq for BinaryHeapWrapper {
+	// Eq here is implemented by creating a vector of all values and then sorting it.
+	// TODO: A better approach would be to use `into_iter_sorted` and compare both iterator, once
+	// https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#method.into_iter_sorted is stable.
```
bkchr, line 33: This is a test and nobody cares how fast tests are^^
```suggestion
```

fuzzer/src/main.rs
```diff
@@ -1,78 +1,216 @@
...
+	fn eq(&self, other: &BinaryHeapWrapper) -> bool {
+		// Sort needed because the Iterator returns the value in arbitrary order.
+		let a = self.0.iter().cloned().collect::<Vec<u32>>().sort();
+		let b = other.0.iter().cloned().collect::<Vec<u32>>().sort();
+		a == b
```
bkchr, line 38:
```suggestion
		self.clone().into_sorted_vec() == other.clone().into_sorted_vec()
```

fuzzer/src/main.rs
```diff
@@ -1,78 +1,216 @@
...
+								println!("- Decoded data:");
+								println!("raw1_trunc = {:?}", raw1_trunc_to_obj_size);
+								println!("raw2 = {:?}", raw2);
+								println!("- Encoded objects:");
+								println!("obj1 = '{:?}'", obj);
+								println!("obj2 = '{:?}'", obj2);
+								println!("Type: {}", std::any::type_name::<$parsed>());
+								panic!("raw1 != raw2");
+							}
+						return
```
bkchr, line 195: Formatting.

fuzzer/src/main.rs
```diff
@@ -1,78 +1,216 @@
...
+						}
+					panic!("obj != obj2; obj={:?}, obj2={:?}", obj, obj2);
```
bkchr, line 197: Same

fuzzer/src/main.rs
```diff
@@ -1,78 +1,216 @@
...
+			let raw1 = &d.clone();
 
-		unreachable!()
+			let maybe_obj = <$parsed>::decode(&mut d);
+			if let Ok(obj) = maybe_obj {
+				let mut d2: &[u8] = &obj.encode();
+				let mut raw2 = Vec::from(d2);
+				// We are sorting here because we're in the "sorted" flow. Useful for container types
+				// which can have multiple valid encoded versions.
+				raw2.sort();
```
bkchr, line 175: You can not sort the raw data! That breaks the encoding!

fuzzer/src/main.rs
pscott: So given this test is used for fuzzing purposes, one could argue that actually the rate at which we can compare two binaryheap will directly impact the fuzzing speed. But maybe I'm missing something here :)

fuzzer/src/main.rs
pscott: Thank you :D

fuzzer/src/main.rs
bkchr: While you are right that the fuzzing rate could be improved, I don't expect any improvements over the solution I have posted below. We would still need to `clone()` and after briefly looking over the implementation, it seems that `into_iter_sorted` is essentially doing the same as `into_sorted_vec`. You could argue that you can stop earlier with an iterator, but yeah. I don't know. If we ever run this fuzzer and get into troubles we can improve.

pscott: Yup I guess the improvement would be minor. I had not seen `into_iter_sorted` when writing the code, so what you suggested is definitely better than my implementation. Thanks! :)

fuzzer/src/main.rs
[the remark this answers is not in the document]
pscott: It looks like you were having a look at some outdated code :D

fuzzer/src/main.rs
pscott: Yes it does indeed! This is what I tried to explain in the comments earlier: either we sort this (breaks encoding, but at least we can compare those bytes with the bytes we'll get after encoding the decoded.  OR, we can decide not to sort those, in which case we''l get false positives, so we should probably just not add those types to the decode-encode flow (and just try and decode them, period).

The first solution is a superset of the second solution (it decodes + encodes) so I would argue that we should use it whenever we can. Of course it is not perfect because we're not comparing the bytes directly but the sorted bytes, but hey it's better than nothing right? :p

fuzzer/src/main.rs
```diff
@@ -90,23 +223,91 @@ fn fuzz_one_input(data: &[u8]){
...
+	($( $type:ty, )*) => {
+		$(fuzz!(|data: $type| { fuzz_encode(data) });)*
+	};
+}
+
+fn fuzz_encode<T: Encode + Decode + Clone + PartialEq + std::fmt::Debug> (data: T) {
+	let original = data.clone();
+	let mut obj: &[u8] = &data.encode();
+	let decoded = <T>::decode(&mut obj);
+	if let Ok(object) = decoded {
```
bkchr, line 260: Again, please use a match.

fuzzer/src/main.rs
```diff
@@ -1,78 +1,211 @@
...
+		if $data[0] % num == $index {
+			let mut d = &$data[1..];
+			let raw1 = &d.clone();
 
-		unreachable!()
+			let maybe_obj = <$parsed>::decode(&mut d);
+			match maybe_obj {
+				Ok(obj) => {
+					let mut d2: &[u8] = &obj.encode();
+					let mut raw2 = Vec::from(d2);
```
bkchr, line 168:
```suggestion
					let mut d2 = obj.encode();
					let mut raw2 = d2.clone();
```

fuzzer/src/main.rs
```diff
@@ -1,78 +1,211 @@
...
+					// We are sorting here because we're in the "sorted" flow. Useful for container types
+					// which can have multiple valid encoded versions.
+					raw2.sort();
+					let exp_obj = <$parsed>::decode(&mut d2);
```
bkchr, line 172:
```suggestion
					let exp_obj = <$parsed>::decode(&mut &d2[..]);
```

fuzzer/src/main.rs
pscott: Argh haha yes I committed the change yesterday but forgot to push! :)

PR #194, opened 2020-03-30
Add .rustfmt.toml with hard_tabs set to true & run cargo fmt

Atul9:


bkchr:
Ty, but we currently don't use rustfmt

PR #198, opened 2020-04-02
publish new codec and codec-derive version

thiolliere:
polkadot, substrate and its dependency currently never implement CompactAs by hand, thus updating scale-codec won't break anything.

if this PR is accepted I will put the two tags and publish them on crate-io.

(note that codec-derive:1.3.0 won't be usable with codec:1.3.0 but only with codec:1.4.0)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "parity-scale-codec"
 description = "SCALE - Simple Concatenating Aggregated Little Endians"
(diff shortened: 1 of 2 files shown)


bkchr:
The problem is that there are maybe other users and we should follow semver IMHO.

bkchr:
I created a milestone and added some issues with stuff that should happen for 2.0. These issues should be fixed and after that we can release `2.0`.

PR #199, opened 2020-04-06
Add Encode and Decode impls for NonZero* types

Robbepop:
Adds `Encode` and `Decode` impls for

- `NonZeroI8`
- `NonZeroI16`
- `NonZeroI32`
- `NonZeroI64`
- `NonZeroI128`
- `NonZeroU8`
- `NonZeroU16`
- `NonZeroU32`
- `NonZeroU64`
- `NonZeroU128`

Does not use the `WrapperTypeEncode` and `WrapperTypeDecode` facilities since they require the `Deref` super trait which is not applicable to these types.

These impls have not been put behind a crate feature since they are very useful for performance critical applications (as ours) as they provide the compiler with optimization opportunities in some use cases.

As soon as specialization is a thing in Rust we could provide specializations for `Option<NonZero*>` as well.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -581,7 +581,7 @@ const _: () = {
 				impl Decode for $name {
 					fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 						let value = <$wrapped as Decode>::decode(input)?;
-						Self::new(value).ok_or_else(|| Error("cannot create non-zero number from 0"))
+						Self::new(value).ok_or_else(|| Error::from("cannot create non-zero number from 0"))
 					}
 				}
 			)*
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -552,6 +552,55 @@ impl<T: Decode> Decode for Option<T> {
 	}
 }
 
+const _: () = {
+	use core::num::{
+		NonZeroI8,
+		NonZeroI16,
+		NonZeroI32,
+		NonZeroI64,
+		NonZeroI128,
+		NonZeroU8,
+		NonZeroU16,
+		NonZeroU32,
+		NonZeroU64,
+		NonZeroU128,
+	};
+	macro_rules! impl_for_non_zero {
+		( $( ($name:ty, $wrapped:ty) ),* $(,)? ) => {
+			$(
+				impl Encode for $name {
+					fn size_hint(&self) -> usize {
+						<$wrapped as Encode>::size_hint(&self.get())
+					}
+
+					fn encode_to<W: Output>(&self, dest: &mut W) {
+						<$wrapped as Encode>::encode_to(&self.get(), dest)
+					}
+				}
+
+				impl Decode for $name {
+					fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+						let value = <$wrapped as Decode>::decode(input)?;
+						Self::new(value).ok_or_else(|| Error("cannot create non-zero number from 0"))
+					}
+				}
+			)*
+		}
+	}
+	impl_for_non_zero! {
+		(NonZeroI8, i8),
+		(NonZeroI16, i16),
+		(NonZeroI32, i32),
+		(NonZeroI64, i64),
+		(NonZeroI128, i128),
+		(NonZeroU8, u8),
+		(NonZeroU16, u16),
+		(NonZeroU32, u32),
+		(NonZeroU64, u64),
+		(NonZeroU128, u128),
+	}
+};
+
 macro_rules! impl_array {
 	( $( $n:expr, )* ) => {
 		$(
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -17,7 +17,18 @@
 #[cfg(feature = "std")]
 use std::fmt;
 use core::{mem, ops::Deref, marker::PhantomData, iter::FromIterator, convert::TryFrom, time::Duration};
-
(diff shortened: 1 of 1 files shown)


src/codec.rs
```diff
@@ -552,6 +552,55 @@ impl<T: Decode> Decode for Option<T> {
 	}
 }
 
+const _: () = {
```
bkchr, line 555: Please move the import to the top and I don't understand why you want to wrap it into an unnamed const.

src/codec.rs
```diff
@@ -552,6 +552,55 @@ impl<T: Decode> Decode for Option<T> {
...
+		NonZeroI64,
+		NonZeroI128,
+		NonZeroU8,
+		NonZeroU16,
+		NonZeroU32,
+		NonZeroU64,
+		NonZeroU128,
+	};
+	macro_rules! impl_for_non_zero {
+		( $( ($name:ty, $wrapped:ty) ),* $(,)? ) => {
```
bkchr, line 569: `wrapped` is not required.

src/codec.rs
```diff
@@ -552,6 +552,55 @@ impl<T: Decode> Decode for Option<T> {
...
+			$(
+				impl Encode for $name {
+					fn size_hint(&self) -> usize {
+						<$wrapped as Encode>::size_hint(&self.get())
```
bkchr, line 573:
```suggestion
						self.get().size_hint()
```

src/codec.rs
```diff
@@ -552,6 +552,55 @@ impl<T: Decode> Decode for Option<T> {
...
+					}
+
+					fn encode_to<W: Output>(&self, dest: &mut W) {
+						<$wrapped as Encode>::encode_to(&self.get(), dest)
```
bkchr, line 577:
```suggestion
						self.get().encode_to(dest)
```

src/codec.rs
```diff
@@ -552,6 +552,55 @@ impl<T: Decode> Decode for Option<T> {
...
+					}
+				}
+
+				impl Decode for $name {
+					fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+						let value = <$wrapped as Decode>::decode(input)?;
+						Self::new(value).ok_or_else(|| Error::from("cannot create non-zero number from 0"))
```
bkchr, line 584:
```suggestion
						Self::new(Decode::decode()?).ok_or_else(|| Error::from("cannot create non-zero number from 0"))
```

src/codec.rs
```diff
@@ -552,6 +552,55 @@ impl<T: Decode> Decode for Option<T> {
...
+		NonZeroU8,
+		NonZeroU16,
+		NonZeroU32,
+		NonZeroU64,
+		NonZeroU128,
+	};
+	macro_rules! impl_for_non_zero {
+		( $( ($name:ty, $wrapped:ty) ),* $(,)? ) => {
+			$(
+				impl Encode for $name {
```
bkchr, line 571: Please also implement the other `encode_` functions.

Robbepop:
@bkchr Okay, according to your review I see that you are not a fan of explicit code in macros. I will change it to be more implicit and make some other edits and re-request your review.

src/codec.rs
Robbepop: I did this to have the import statement for the `NonZero*` types close to their use site. I generally prefer this in code bases but will change it for this PR.

src/codec.rs
Robbepop: True, but also way less explicit. I prefer being explicit in macros.

bkchr:
I'm a fan of it, if it is required. 

However, I'm more a fan of short code ;)

src/codec.rs
Robbepop: What is the concrete use if the default implementation is as good? Is there a benchmark that proofs that this is worth the additional code?

src/codec.rs
Robbepop: I like explicit library and macro code because it is far more robust to bit rotting, e.g.:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=51f7bb4b11f26565b4fb420647901d95

This fails as soon as one of the primitive types receives a method `encode_to` or a trait with an `encode_to` method comes into play whereas the explicit version prevents this.

Robbepop:
> I'm a fan of it, if it is required.
> 
> However, I'm more a fan of short code ;)

But aren't you a fan of self-contained code?
The near use-site import makes the code self contained and the explicit type notion isolates it from its surroundings that could inflict it upon certain refactoring if written in implicit form as shown in my comments.

bkchr:
Self contained and isolated is for me a new file^^

src/codec.rs
Robbepop: done

src/codec.rs
Robbepop: done

src/codec.rs
Robbepop: applied suggestion in latest commit

src/codec.rs
Robbepop: applied suggestion in latest commit

src/codec.rs
Robbepop: done

src/codec.rs
Robbepop: done

Issue #200, opened 2020-04-07
#[codec(index = "200")] should not take a string


bkchr:
The macro should take the index as integer.

PR #206, opened 2020-04-08
Add check for attribute and modify index attribute

thiolliere:
Fix https://github.com/paritytech/parity-scale-codec/issues/200

Previously incorrect attributes were mostly not checked or were panicking the macro.

So attribute like `#[codec(index = "15")]` would simply be ignored because not following the syntax.

I first tried to change the code to make the attribute finder return a Result but because they are used for filtering it is very not convenient.

Instead I just created a global attribute checker which checks that the attributes are correct.

### DIFF
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -12,12 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Various internal utils.
+//!
+//! NOTE: attributes finder must be checked using check_attribute first, otherwise macro can panic.
+
 use std::str::FromStr;
 
 use proc_macro2::{TokenStream, Span};
 use syn::{
 	spanned::Spanned,
-	Meta, NestedMeta, Lit, Attribute, Variant, Field,
+	Meta, NestedMeta, Lit, Attribute, Variant, Field, DeriveInput, Fields, Data, FieldsUnnamed,
+	FieldsNamed, MetaNameValue
 };
 
 fn find_meta_item<'a, F, R, I>(itr: I, pred: F) -> Option<R> where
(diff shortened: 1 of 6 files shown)


derive/src/utils.rs
```diff
@@ -127,3 +136,120 @@ pub fn filter_skip_unnamed<'a>(fields: &'a syn::FieldsUnnamed) -> impl Iterator<
...
+	if attr.path.is_ident("codec") {
+		match attr.parse_meta()? {
+			Meta::List(ref meta_list) if meta_list.nested.len() == 1 => {
+				match meta_list.nested.first().unwrap() {
+					NestedMeta::Meta(Meta::Path(path))
+						if path.get_ident().map_or(false, |i| i == "skip") => Ok(()),
+
+					NestedMeta::Meta(Meta::NameValue(MetaNameValue { path, lit: Lit::Int(lit_int), .. }))
+						if path.get_ident().map_or(false, |i| i == "index")
+					=> lit_int.base10_parse::<u8>().map(|_| ()),
```
thiolliere, line 224: here if the error is raised it will output:
```
error: number too large to fit in target type
  --> tests/mod.rs:50:18
   |
50 |     #[codec(index = 15000)]
   |                     ^^^^^

```

bkchr: I think you should generate a custom error here. Saying that the maximum is 256 ;)

derive/src/utils.rs
```diff
@@ -127,3 +135,120 @@ pub fn filter_skip_unnamed<'a>(fields: &'a syn::FieldsUnnamed) -> impl Iterator<
...
+		match attr.parse_meta()? {
+			Meta::List(ref meta_list) if meta_list.nested.len() == 1 => {
+				match meta_list.nested.first().unwrap() {
+					NestedMeta::Meta(Meta::Path(path))
+						if path.get_ident().map_or(false, |i| i == "skip") => Ok(()),
+
+					NestedMeta::Meta(Meta::NameValue(MetaNameValue { path, lit: Lit::Int(lit_int), .. }))
+						if path.get_ident().map_or(false, |i| i == "index")
+					=> lit_int.base10_parse::<u8>().map(|_| ())
+						.map_err(|_| syn::Error::new(lit_int.span(), "Index must be in 0..255")),
```
thiolliere, line 223: ok now it shows (note that it is already checked to be a valid int literal)

```
error: Index must be in 0..255
  --> tests/mod.rs:50:18
   |
50 |     #[codec(index = 256)]
   |                     ^^^

```

derive/src/utils.rs
thiolliere: 👍 done

PR #212, opened 2020-06-03
IoReader suppose Seek impl to be of fixed len, substrate usage is otherwise

thiolliere:
In our code we expect remaining_len implementation to give the exact remaining_length (or an upper bound actually).

for instance:
https://github.com/paritytech/parity-scale-codec/blob/353db2cf1e417e1d0216e882dc0fb74c769f4a56/src/codec.rs#L777-L784

but IoReader use Seek trait to get the len, and Seek doesn't guarantee to return the final length, length of Seek implementation can change over time.

### Usage in substrate

in substrate, ioreader is used for a generic `Seek + Read` https://github.com/paritytech/substrate/blob/master/client/service/src/chain_ops.rs

(If I read correctly it is used inside a future context, I guess that means that Read will wait when reaching the end. Thus Seek implementation gives varying len of stream.)

### Additional notes

if wanted we could extend Input trait to return a custom max_preallocation.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -184,22 +184,7 @@ pub struct IoReader<R: std::io::Read + std::io::Seek>(pub R);
 #[cfg(feature = "std")]
 impl<R: std::io::Read + std::io::Seek> Input for IoReader<R> {
 	fn remaining_len(&mut self) -> Result<Option<usize>, Error> {
-		use std::convert::TryInto;
-		use std::io::SeekFrom;
-
-		let old_pos = self.0.seek(SeekFrom::Current(0))?;
-		let len = self.0.seek(SeekFrom::End(0))?;
-
-		// Avoid seeking a third time when we were already at the end of the
-		// stream. The branch is usually way cheaper than a seek operation.
-		if old_pos != len {
-			self.0.seek(SeekFrom::Start(old_pos))?;
-		}
-
-		len.saturating_sub(old_pos)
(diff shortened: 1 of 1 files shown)


thiolliere:
(Though maybe we should keep Seek trait for the implementation of skip)

thiolliere:
> If we drop this, we will loose the ability to get the remaining length even for normal Vec<u8>.

actually this implementation is for `IoReader<Read>`, I kept the manual implementation for `&[u8]` to keep the optimisation.

> However, maybe it is fine. Because if the user adds new data to the file between subsequent calls, the worst case should be that we probably return an error because some data is not present/ can not be decoded.
> 
> And files are not being used that much by us.

hmm but I'm still thinking that our usage in substrate is a varying length Seek, and when we stuck on Read the future is paused, though I'm very not familiar with this part of the code.

> We could also change the return value of remaining_len to an enum that tells the user if the returned value is a fixed upper bound etc.

yes either that or remaining_len returns `option<len>` and an additional method with default implementation
```rust
fn max_preallocation(&mut self) -> usize {
    self.remaining_len()?.unwrap_or(codec::MAX_PREALLOCATION)
}
```

and IoReader could have a field and do `IoReader::new_with_max_preallocation(...)`

bkchr:
Ahh, `IoReader`. Yeah sorry, I think than we can just drop it.

PR #214, opened 2020-06-19
Decode with recursion limit

arkpar:


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "parity-scale-codec"
 description = "SCALE - Simple Concatenating Aggregated Little Endians"
-version = "1.3.0"
+version = "1.3.1"
```
bkchr, line 4: We can not release this version. Master has breaking changes. If you want to release this. I propose that you backport this after merge to master. We have a tag for the last release.

src/codec.rs
```diff
@@ -131,6 +131,14 @@ pub trait Input {
...
 		Ok(buf[0])
 	}
+
+	/// Descend into nested structure when decoding
+	fn descend(&mut self) -> Result<(), Error> {
+		Ok(())
+	}
+
+	/// Ascend to previous structure level when decoding.
+	fn ascend(&mut self) { }
```
bkchr, line 141:
```suggestion
	fn ascend(&mut self) {}
```

src/depth_limit.rs
```diff
@@ -0,0 +1,108 @@
...
+	/// Decode `Self` with given maximum recursion depth.
+	/// is returned.
+	fn decode_with_depth_limit(limit: u32, input: &[u8]) -> Result<Self, Error>;
+	/// Decode `Self` and consume all of the given input data. If not all data is consumed, an error
+	/// is returned.
+	fn decode_all_with_depth_limit(limit: u32, input: &[u8]) -> Result<Self, Error>;
+}
+
+
+pub struct DepthTrackingInput<'a, I> {
```
bkchr, line 32: Missing docs.

src/codec.rs
```diff
@@ -131,6 +131,14 @@ pub trait Input {
 		self.read(&mut buf[..])?;
 		Ok(buf[0])
 	}
+
+	/// Descend into nested structure when decoding
+	fn descend(&mut self) -> Result<(), Error> {
```
bkchr, line 136: From the api docs and the usage I don't understand when we enter a nested structure.

If I have a struct as field of a struct, aren't we entering a new nested structure as well?

arkpar: It only makes sense to trace nested references as nested structures are not allowed in rust anyway. I'll update docs.

src/codec.rs
```diff
@@ -131,6 +131,17 @@ pub trait Input {
 		self.read(&mut buf[..])?;
 		Ok(buf[0])
 	}
+
+	/// Descend into nested reference when decoding.
```
bkchr, line 135:
```suggestion
	/// Descend into nested heap allocated structure while decoding.
```

src/codec.rs
```diff
@@ -131,6 +131,17 @@ pub trait Input {
...
+	/// This is called when decoding a new refence-based instance,
+	/// such as `Vec` or `Box`
```
bkchr, line 137:
```suggestion
	///
	/// This must be called when decoding a heap-based instance,
	/// such as `Vec` or `Box`.
```

src/codec.rs
```diff
@@ -131,6 +131,17 @@ pub trait Input {
...
+	fn descend(&mut self) -> Result<(), Error> {
```
bkchr, line 138:
```suggestion
	fn descend_heap(&mut self) -> Result<(), Error> {
```

Also not that good, but only `descend` isn't enough IMHO.

src/codec.rs
```diff
@@ -131,6 +131,17 @@ pub trait Input {
...
+		Ok(())
+	}
+
+	/// Ascend to previous structure level when decoding.
+	/// This is called when decoding reference-based type is finished.
+	fn ascend(&mut self) {}
```
bkchr, line 144: Same as above.

src/codec.rs
```diff
@@ -131,6 +131,17 @@ pub trait Input {
...
+
+	/// Descend into nested reference when decoding.
+	/// This is called when decoding a new refence-based instance,
+	/// such as `Vec` or `Box`
+	fn descend(&mut self) -> Result<(), Error> {
+		Ok(())
+	}
+
+	/// Ascend to previous structure level when decoding.
+	/// This is called when decoding reference-based type is finished.
```
bkchr, line 143:
```suggestion
	///
	/// This is called when decoding heap-based type is finished.
```

src/codec.rs
```diff
@@ -405,7 +416,10 @@ impl<T, X> Decode for X where
 	X: WrapperTypeDecode<Wrapped=T>,
 {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-		Ok(T::decode(input)?.into())
+		input.descend()?;
```
bkchr, line 419: `WrapperTypeDecode` is not only used for heap based structures.

src/depth_limit.rs
```diff
@@ -0,0 +1,108 @@
...
+		if input.input.is_empty() {
+			Ok(res)
+		} else {
+			Err(crate::decode_all::DECODE_ALL_ERR_MSG.into())
+		}
+	}
+
+	fn decode_with_depth_limit(limit: u32, input: &[u8]) -> Result<Self, Error> {
+		let mut input = DepthTrackingInput {
+			input: &mut &input[..],
```
KiChjang, line 85: This here creates a new byte slice that's separate from `input`, which would not advance the `input` index after decoding is done.

arkpar: That would require changing the signature to accept a mutable input. If such a use case is required a PR is welcome

PR #219, opened 2020-08-13
Remove unrequired bounds in HasCompact

thiolliere:
the trait HasCompact doesn't need to bound Clone, Eq, PartialEq, MaybeDebugSerde.

Also MaybeDebugSerde is not used by scale-codec in any way.

I think trait makes more sense know, also this is ok as master contains some few breaking change already.

### DIFF
--- a/src/compact.rs
+++ b/src/compact.rs
@@ -191,6 +191,20 @@ impl<T> core::fmt::Debug for Compact<T> where T: core::fmt::Debug {
 	}
 }
 
+#[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


bkchr:
Did you tested this in Substrate? I think there was a reason I added them :D

thiolliere:
(I revert the accidentally removal of serde implementation for compact, I'm now testing if this is breaking polkadot and substrate).

bkchr (approved):

bkchr:
> Did you tested this in Substrate? I think there was a reason I added them :D

Apparently this past me neither now me seem to know why past-past me added this xD

PR #220, opened 2020-08-26
Backport fix name conflict and release

thiolliere:
This is the fix for name conflicts for 1.3.4
(currently 1.3.5-release is same as 1.3.4-release, once this PR is merge I'll tag a new version for codec and codec-derive and publish both)

additional note: when we do:
```rust
match foo {
Foo::Variant { foo } => ...
}
```
the match conflicts if foo is defined as a const in scope
but when we do:
```rust
match foo {
Foo::Variant { ref foo } => ...
}
```
the conflicts result in rust error.

see https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=be416a1e0828637dfb31f18e9161af33

Currently in our code all the match is done with ref so we are fine, though maybe we should prefer doing like `foo: $unique_name` instead ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "parity-scale-codec"
 description = "SCALE - Simple Concatenating Aggregated Little Endians"
-version = "1.3.4"
+version = "1.3.5"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "Apache-2.0"
 repository = "https://github.com/paritytech/parity-scale-codec"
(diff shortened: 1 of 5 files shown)


thiolliere:
maybe we can also backport:
https://github.com/paritytech/parity-scale-codec/pull/199
https://github.com/paritytech/parity-scale-codec/pull/209
https://github.com/paritytech/parity-scale-codec/pull/211
https://github.com/paritytech/parity-scale-codec/pull/212
https://github.com/paritytech/parity-scale-codec/pull/215

I can do a backport PR for each

PR #222, opened 2020-08-26
Fix name conflict in derive macros

thiolliere:
This is the fix for name conflicts for master (no difference with the backport fix)

additional note: when we do:
```rust
match foo {
Foo::Variant { foo } => ...
}
```
the match conflicts if foo is defined as a const in scope
but when we do:
```rust
match foo {
Foo::Variant { ref foo } => ...
}
```
the conflicts result in rust error.

see https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=be416a1e0828637dfb31f18e9161af33

Currently in our code all the match is done with ref so we are fine, though maybe we should prefer doing like `foo: $unique_name` instead ?

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -195,13 +195,13 @@ pub fn decode_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream 
 	let name = &input.ident;
 	let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
 
-	let input_ = quote!(input);
+	let input_ = quote!(__codec_input_edqy);
 	let decoding = decode::quote(&input.data, name, &input_);
 
 	let impl_block = quote! {
 		impl #impl_generics _parity_scale_codec::Decode for #name #ty_generics #where_clause {
-			fn decode<DecIn: _parity_scale_codec::Input>(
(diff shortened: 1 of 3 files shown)


derive/src/lib.rs
```diff
@@ -195,13 +195,13 @@ pub fn decode_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream
...
 
-	let input_ = quote!(input);
+	let input_ = quote!(__codec_input_edqy);
 	let decoding = decode::quote(&input.data, name, &input_);
 
 	let impl_block = quote! {
 		impl #impl_generics _parity_scale_codec::Decode for #name #ty_generics #where_clause {
-			fn decode<DecIn: _parity_scale_codec::Input>(
-				#input_: &mut DecIn
+			fn decode<__CodecInputEdqy: _parity_scale_codec::Input>(
```
bkchr, line 203: Just one question, why always "edqy"? :D

bkchr (approved):

derive/src/lib.rs
thiolliere: I just generated 4 random characters, but thinking of it again it is overkill :-)

PR #223, opened 2020-09-24
Export `WrapperTypeDecode` from lib.rs

vmarkushin:


thiolliere (approved):

thiolliere:
Do you need a new release with this?

PERSONAL_NOTE: it would be useful to backport this PR to 1.* as well.

vmarkushin:
@thiolliere it's not necessary at the moment I made a similar implementation via `Decode`.

PR #226, opened 2020-10-08
Always implement Display for Error

stephaneyfx:
This makes it easier to write code that uses `Error` with or without `std` (e.g. it avoids having to write conditionally compiled code and allows using `Error` in places that require `Display`).

Thank you.


bkchr (approved):

stephaneyfx:
Thank you!

PR #231, opened 2020-12-21
Fix Vec::decode as reported by Brian and endianness

thiolliere:
*    fix alignment in `bitvec` decoding
*    fix endianness in codec and `bitvec`
*    fix miri reported error on `read_vec_u8` (renamed to `read_vec_from_u8s`) optimized decoding of some vec. Due of transmutation of `Vec<u8>` to `Vec<T>`.

This PR is meant to be backported.

Then a follow up PR will contains bump of bitvec and byte_slice_cast and some minor refactor.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,6 +54,13 @@ test-rust-stable:
     - time cargo test --verbose --all --features bit-vec,generic-array,derive
     - sccache -s
 
+test-rust-miri:
+  stage:                           test
+  <<:                              *docker-env
+  script:
+    - time cargo +nightly miri test --features bit-vec,generic-array,arbitrary --release
+    - sccache -s
+
 test-rust-stable-no_derive:
   stage:                           test
   <<:                              *docker-env
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,13 +54,6 @@ test-rust-stable:
     - time cargo test --verbose --all --features bit-vec,generic-array,derive
     - sccache -s
 
-test-rust-miri:
-  stage:                           test
-  <<:                              *docker-env
-  script:
-    - time cargo +nightly miri test --features bit-vec,generic-array,arbitrary --release
-    - sccache -s
-
 test-rust-stable-no_derive:
   stage:                           test
   <<:                              *docker-env
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -30,8 +30,6 @@ impl From<FromByteSliceError> for Error {
 				"failed to cast from byte slice: alignment mismatch".into(),
 			FromByteSliceError::LengthMismatch {..} =>
 				"failed to cast from byte slice: length mismatch".into(),
-			FromByteSliceError::CapacityMismatch {..} =>
-				"failed to cast from byte slice: capacity mismatch".into(),
 		}
 	}
 }
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -19,7 +19,7 @@ use core::mem;
 use bitvec::{vec::BitVec, store::BitStore, order::BitOrder, slice::BitSlice, boxed::BitBox};
 use byte_slice_cast::{AsByteSlice, ToByteSlice, FromByteSlice, Error as FromByteSliceError};
 
-use crate::codec::{Encode, Decode, Input, Output, Error, read_vec_u8};
+use crate::codec::{Encode, Decode, Input, Output, Error, read_vec_from_u8s};
 use crate::compact::Compact;
 use crate::EncodeLike;
 
@@ -62,7 +62,7 @@ impl<O: BitOrder, T: BitStore + FromByteSlice> Decode for BitVec<O, T> {
 			let bits = bits as usize;
 			let required_bytes = required_bytes::<T>(bits);
 
-			let vec = read_vec_u8(input, required_bytes)?;
+			let vec = read_vec_from_u8s(input, required_bytes)?;
 
 			let mut result = Self::from_slice(T::from_byte_slice(&vec)?);
 			assert!(bits <= result.len());
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -101,7 +101,7 @@ impl<O: BitOrder, T: BitStore + FromByteSlice> Decode for BitVec<O, T> {
 			unsafe {
 				let aligned_u8_ptr = aligned_vec.as_mut_ptr() as *mut u8;
 				for (i, v) in vec_u8.iter().enumerate() {
-					*aligned_u8_ptr.offset(i as isize) = *v;
+					*aligned_u8_ptr.add(i) = *v;
 				}
 			}
 
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -30,6 +30,8 @@ impl From<FromByteSliceError> for Error {
 				"failed to cast from byte slice: alignment mismatch".into(),
 			FromByteSliceError::LengthMismatch {..} =>
 				"failed to cast from byte slice: length mismatch".into(),
+			FromByteSliceError::CapacityMismatch {..} =>
+				"failed to cast from byte slice: capacity mismatch".into(),
 		}
 	}
 }
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -84,9 +84,9 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> EncodeLike for BitVec<O, T> {}
 
 /// # WARNING
 ///
-/// In bitvec v0.17.4 the only implementation of BitStore are u8, u16, u32, u64, usize.
-/// This implementation actually only support u8, u16, u32 and u64, encoding of with BitStore being
-/// usize is inconsistent between platforms.
+/// In bitvec v0.17.4 the only implementations of BitStore are u8, u16, u32, u64, and usize.
+/// This implementation actually only support u8, u16, u32 and u64, as encoding of usize
+/// is inconsistent between platforms.
 impl<O: BitOrder, T: BitStore + FromByteSlice> Decode for BitVec<O, T> {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(bits)| {
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -46,7 +46,7 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitSlice<O, T> {
 /// Reverse bytes of element for element of size `size_of_t`.
 ///
 /// E.g. if size is 2 `[1, 2, 3, 4]` is changed to `[2, 1, 4, 3]`.
-fn reverse_endian(vec_u8: &mut Vec<u8>, size_of_t: usize) {
+fn reverse_endian(vec_u8: &mut [u8], size_of_t: usize) {
 	for i in 0..vec_u8.len() / size_of_t {
 		for j in 0..size_of_t / 2 {
 			vec_u8.swap(i * size_of_t + j, i * size_of_t + (size_of_t - 1) - j);
@@ -72,7 +72,7 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
 
 		if cfg!(target_endian = "big") && mem::size_of::<T>() > 1 {
 			let mut vec_u8: Vec<u8> = byte_slice.into();
-			reverse_endian(&mut vec_u8, mem::size_of::<T>());
+			reverse_endian(&mut vec_u8[..], mem::size_of::<T>());
 			dest.write(&vec_u8);
 		} else {
 			dest.write(byte_slice);
@@ -96,7 +96,7 @@ impl<O: BitOrder, T: BitStore + FromByteSlice> Decode for BitVec<O, T> {
 			let mut vec_u8 = read_vec_from_u8s::<I, u8>(input, required_bytes)?;
 
 			if cfg!(target_endian = "big") && mem::size_of::<T>() > 1 {
-				reverse_endian(&mut vec_u8, mem::size_of::<T>());
+				reverse_endian(&mut vec_u8[..], mem::size_of::<T>());
 			}
 
 			let mut aligned_vec: Vec<T> = vec![0u8.into(); required_bytes / mem::size_of::<T>()];
@@ -263,19 +263,19 @@ mod tests {
 		let data = vec![1, 2, 3, 4, 5, 6, 7, 8];
 
 		let mut data_to_u8 = data.clone();
-		reverse_endian(&mut data_to_u8, mem::size_of::<u8>());
+		reverse_endian(&mut data_to_u8[..], mem::size_of::<u8>());
 		assert_eq!(data_to_u8, data);
 
 		let mut data_to_u16 = data.clone();
-		reverse_endian(&mut data_to_u16, mem::size_of::<u16>());
+		reverse_endian(&mut data_to_u16[..], mem::size_of::<u16>());
 		assert_eq!(data_to_u16, vec![2, 1, 4, 3, 6, 5, 8, 7]);
 
 		let mut data_to_u32 = data.clone();
-		reverse_endian(&mut data_to_u32, mem::size_of::<u32>());
+		reverse_endian(&mut data_to_u32[..], mem::size_of::<u32>());
 		assert_eq!(data_to_u32, vec![4, 3, 2, 1, 8, 7, 6, 5]);
 
 		let mut data_to_u64 = data.clone();
-		reverse_endian(&mut data_to_u64, mem::size_of::<u64>());
+		reverse_endian(&mut data_to_u64[..], mem::size_of::<u64>());
 		assert_eq!(data_to_u64, vec![8, 7, 6, 5, 4, 3, 2, 1]);
 	}
 }
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -60,14 +60,15 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
 		);
 		Compact(len as u32).encode_to(dest);
 
-		let mut vec_u8: Vec<u8> = self.as_slice().as_byte_slice().into();
+		let byte_slice: &[u8] = self.as_slice().as_byte_slice();
 
-		let size_of_t = mem::size_of::<T>();
-		if cfg!(target_endian = "big") && size_of_t != 1 {
-			reverse_endian(&mut vec_u8, size_of_t);
+		if cfg!(target_endian = "big") && mem::size_of::<T>() != 1 {
+			let mut vec_u8: Vec<u8> = byte_slice.into();
+			reverse_endian(&mut vec_u8, mem::size_of::<T>());
+			dest.write(&vec_u8);
+		} else {
+			dest.write(byte_slice);
 		}
-
-		dest.write(&vec_u8);
 	}
 }
 
@@ -82,9 +83,8 @@ impl<O: BitOrder, T: BitStore + FromByteSlice> Decode for BitVec<O, T> {
 
 			let mut vec_u8 = read_vec_from_u8s::<I, u8>(input, required_bytes)?;
 
-			let size_of_t = mem::size_of::<T>();
-			if cfg!(target_endian = "big") && size_of_t != 1 {
-				reverse_endian(&mut vec_u8, size_of_t);
+			if cfg!(target_endian = "big") && mem::size_of::<T>() != 1 {
+				reverse_endian(&mut vec_u8, mem::size_of::<T>());
 			}
 
 			let mut result = Self::from_slice(T::from_byte_slice(&vec_u8)?);
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -55,7 +55,7 @@ fn reverse_endian(vec_u8: &mut Vec<u8>, size_of_t: usize) {
 ///
 /// In bitvec v0.17.4 the only implementation of BitStore are u8, u16, u32, u64, usize.
 /// This implementation actually only support u8, u16, u32 and u64, encoding of with BitStore being
-/// usize is inconsistent between plateform.
+/// usize is inconsistent between platforms.
 impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
 	fn encode_to<W: Output>(&self, dest: &mut W) {
 		let len = self.len();
@@ -67,7 +67,7 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
 
 		let byte_slice: &[u8] = self.as_slice().as_byte_slice();
 
-		if cfg!(target_endian = "big") && mem::size_of::<T>() != 1 {
+		if cfg!(target_endian = "big") && mem::size_of::<T>() > 1 {
 			let mut vec_u8: Vec<u8> = byte_slice.into();
 			reverse_endian(&mut vec_u8, mem::size_of::<T>());
 			dest.write(&vec_u8);
@@ -83,7 +83,7 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> EncodeLike for BitVec<O, T> {}
 ///
 /// In bitvec v0.17.4 the only implementation of BitStore are u8, u16, u32, u64, usize.
 /// This implementation actually only support u8, u16, u32 and u64, encoding of with BitStore being
-/// usize is inconsistent between plateform.
+/// usize is inconsistent between platforms.
 impl<O: BitOrder, T: BitStore + FromByteSlice> Decode for BitVec<O, T> {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(bits)| {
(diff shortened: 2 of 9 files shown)


src/codec.rs
```diff
@@ -782,36 +782,71 @@ impl<T: Encode> Encode for [T] {
 	}
 }
 
-/// Read an `u8` vector from the given input.
-pub(crate) fn read_vec_u8<I: Input>(input: &mut I, len: usize) -> Result<Vec<u8>, Error> {
+/// Create a `Vec<T>` by casting directly from a buffer of read `u8`s
+///
+/// `T` encoding must be constant and same size as size of `T`, size of `T` must be less or equal
+/// to [`MAX_PREALLOCATION`].
```
thiolliere, line 788:
```suggestion
/// `T` must encodes to its byte representation, size of `T` must be less or equal to
/// [`MAX_PREALLOCATION`].
```

src/codec.rs
```diff
@@ -782,36 +782,71 @@ impl<T: Encode> Encode for [T] {
...
+			let items_len_filled = items.len();
+			let items_new_size = items_len_filled + items_len_read;
+
+			// This call is here to satisfy the existing
+			// vec_decode_right_capacity test.
+			// It's not obviously functionally necessary,
+			// nor is it obviously a problem.
+			items.reserve_exact(items_len_read);
+
+			items.resize_with(items_new_size, Default::default);
```
bkchr, line 839: Could you run the benchmarks of this against using `set_len` here?

src/codec.rs
```diff
@@ -782,36 +782,71 @@ impl<T: Encode> Encode for [T] {
...
+
+			let bytes_slice = items.as_mut_byte_slice();
+			let bytes_len_filled = items_len_filled * mem::size_of::<T>();
+			input.read(&mut bytes_slice[bytes_len_filled..])?;
+
+			items_remains = items_remains.checked_sub(items_len_read)
+				.expect("`items_len_read` is less than `items_remains`; qed");
```
bkchr, line 846:
```suggestion
			items_remains = items_remains.saturating_sub(items_len_read);
```

src/codec.rs
```diff
@@ -782,36 +782,71 @@ impl<T: Encode> Encode for [T] {
...
-			let len_read = MAX_PREALLOCATION.min(remains);
-			let len_filled = r.len();
-			r.resize(len_filled + len_read, 0);
-			input.read(&mut r[len_filled..])?;
-			remains -= len_read;
+		let mut items: Vec<T> = vec![];
+
+		let mut items_remains = items_len;
+
+		while items_remains != 0 {
```
bkchr, line 827:
```suggestion
		while items_remains > 0 {
```

bkchr (approved):
Looks good to me, besides the nitpicks/request.

bkchr:
> note: `bitvec` is not used used in polkadot as far as I know.

`bitvec` is used by Polkadot, but not yet on main net, if you mean this.

src/codec.rs
```diff
@@ -782,36 +782,71 @@ impl<T: Encode> Encode for [T] {
 	}
 }
 
-/// Read an `u8` vector from the given input.
-pub(crate) fn read_vec_u8<I: Input>(input: &mut I, len: usize) -> Result<Vec<u8>, Error> {
+/// Create a `Vec<T>` by casting directly from a buffer of read `u8`s
+///
+/// `T` must encodes to its byte representation, size of `T` must be less or equal to
```
bkchr, line 787: > `T` must encodes to its byte representation

What you mean by this?

src/codec.rs
```diff
@@ -782,36 +782,71 @@ impl<T: Encode> Encode for [T] {
...
+/// [`MAX_PREALLOCATION`].
```
thiolliere, line 788:
```suggestion
/// `T` encoding must be equal to size of `T`, size of `T` must be less or equal to
/// [`MAX_PREALLOCATION`].
```

src/codec.rs
thiolliere: I was thinking the binary representation of the type, but I changed to https://github.com/paritytech/parity-scale-codec/pull/231/files#r546887679

But this makes me think, when encoding/decoding `Vec<T>` we don't take care of the endian of the plateform. `Encode` calls `as_byte_slice` which has different behavior depending on endian of the plateform. Isn't this wrong ?

thiolliere: I guess we should do this optimisation only if `cfg(target_endian = little)`

bkchr: > When encoding/decoding `Vec<T>` we don't take care of the endian of the plateform. `Encode` calls `as_byte_slice` which has different behavior depending on endian of the plateform. Isn't this wrong ?

That is a very good point :see_no_evil:

src/codec.rs
bkchr: What do you think of 

```
The encoding of `T` must be equal to its binary representation,...
```
?

bkchr:
@thiolliere could you also add the miri test run to CI?

src/codec.rs
thiolliere: yes done as you suggest

src/codec.rs
thiolliere: I made a commit to fix this

src/codec.rs
thiolliere: there is quite a difference it goes roughly from 5,7 to 4,5

I changed for `set_len`

src/bit_vec.rs
```diff
@@ -50,21 +59,35 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
 
-			let vec = read_vec_u8(input, required_bytes)?;
+			let mut vec_u8 = read_vec_from_u8s::<I, u8>(input, required_bytes)?;
+
+			if cfg!(target_endian = "big") && mem::size_of::<T>() != 1 {
+				reverse_endian(&mut vec_u8, mem::size_of::<T>());
+			}
 
-			let mut result = Self::from_slice(T::from_byte_slice(&vec)?);
+			let mut result = Self::from_slice(T::from_byte_slice(&vec_u8)?);
```
thiolliere, line 90: there is still this `?` which an error to me, in case the alignment of vec_u8 is wrong the decoding will fail.
But I struggle at the moment to create a `Vec<T>`.

Maybe I can do:
```rust
let aligned_vec = Vec::<T>::new()
// resize and set_len
// unsafe transmute to &mut [u8]
// Copy from vec_u8 to aligned_vec

**Comment by thiolliere** [comment]
So there is still https://github.com/paritytech/parity-scale-codec/pull/231#discussion_r547303587
And I'm investigating some fuzzer stuff, but otherwise the PR looks OK.

this PR still meant to be backported. Then we can change update bitvec and byte_slice_cast. And update the signature of bitvec encode/decode implementation (as anyway types have changed a bit in bitvec).

File: src/bit_vec.rs (hunk 7 above)
Comment (thiolliere) [reply src/bit_vec.rs]: I commit an unsafe fix for it. all this part should be refactored with the upgrade on bitvec. Then we can bound `T: BitStore + Encode` or `T: BitStore + Decode` and implementation will be way better.

File: src/bit_vec.rs [hunk 8]
```diff
@@ -42,6 +43,19 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitSlice<O, T> {
...
+			vec_u8.swap(i * size_of_t + j, i * size_of_t + (size_of_t - 1) - j);
+		}
+	}
+}
+
+/// # WARNING
+///
+/// In bitvec v0.17.4 the only implementation of BitStore are u8, u16, u32, u64, usize.
+/// This implementation actually only support u8, u16, u32 and u64, encoding of with BitStore being
+/// usize is inconsistent between plateform.
```
thiolliere, line 58: hopefully bitvec will never release a new minor version of 0.17 with more implementation of BitStore.

.gitlab-ci.yml
```diff
@@ -54,6 +54,13 @@ test-rust-stable:
     - time cargo test --verbose --all --features bit-vec,generic-array,derive
     - sccache -s
 
+test-rust-miri:
+  stage:                           test
+  <<:                              *docker-env
+  script:
+    - time cargo +nightly miri test --features bit-vec,generic-array,arbitrary --release
```
thiolliere, line 61: I think the `cargo +nightly miri` requires miri component on the image,
I installed it with `rustup +nightly component add miri`

Would it be possible to have it in the image somehow ? @TriplEight

src/bit_vec.rs
```diff
@@ -168,6 +210,7 @@ mod tests {
 	}
 
 	#[test]
+	#[cfg_attr(miri, ignore)] // BitVec error due to outdated version of bitvec
```
thiolliere, line 213: the actuall error is: 
[full_error.txt](https://github.com/paritytech/parity-scale-codec/files/5731837/full_error.txt)

Once bitvec is migrated miri test failed at decoding because of wrong alignment. More precisely at line 67 `T::from_byte_slice(&vec)`
This is why I fixed it by copying from `Vec<u8>` to `Vec<T>`. This is not the fastest but at least it works.
Anyway I expect to have this refactored with upgrade of bitvec.

Some context on the alignment failure here https://github.com/brson/parity-scale-codec/commit/ad6c0aa00f8f503828a59a2f4a92cb8d5d7d280d#r45328246

src/codec.rs
```diff
@@ -635,8 +635,14 @@ macro_rules! impl_array {
...
 						( $ty:ty, $self:ident, $dest:ident ) => {{
-							let typed = unsafe { mem::transmute::<&[T], &[$ty]>(&$self[..]) };
-							$dest.write(<[$ty] as AsByteSlice<$ty>>::as_byte_slice(typed))
+							if cfg!(target_endian = "little") {
+								let typed = unsafe { mem::transmute::<&[T], &[$ty]>(&$self[..]) };
+								$dest.write(<[$ty] as AsByteSlice<$ty>>::as_byte_slice(typed))
+							} else {
+								for item in self.iter() {
+									item.encode_to(dest);
+								}
```
thiolliere, line 644: We can probably do a bit better than this by reading in little endian and then reversing bytes for big endian inplace. But I'm not sure of the speed improvment and I'm not sure of the need. Thus I kept it simple for now.

src/encode_append.rs
```diff
@@ -155,28 +155,31 @@ mod tests {
 	use crate::{Input, Encode, EncodeLike};
 	use std::collections::VecDeque;
 
+	const TEST_VALUE: u32 = {
+		#[cfg(not(miri))]
+		{ 1_000_000 }
+		#[cfg(miri)]
+		{ 1_000 }
```
thiolliere, line 162: otherwise miri test takes ages

thiolliere (commented):
I think the PR is finally ready so finally it does:
* fix alignment in bitvec decoding
* fix endianness in codec and bitvec
* fix miri reported error on `read_vec_u8` (renamed to `read_vec_from_u8s`) optimized decoding of some vec.

.gitlab-ci.yml
TriplEight: You are using `paritytech/ci-linux` CI image, which is main for substrate and polkadot. I know `miri` is a great tool and it would be cool to have it at least substrate CI. However, it's up to devs to decide.
The remaining downside of having `miri` installed in the image is it sometimes [lags behind](https://rust-lang.github.io/rustup-components-history/) the latest nightly. I'm not sure if it's _that_ important, but so far we've never had any components that do not follow the latest image. That being said, this is completely normal for `paritytech/sccache-ci-ubuntu`, `paritytech/contracts-ci-linux` and ``paritytech/ink-ci-linux` images.
I'd be really against having it installed on every push here and would definitely support your idea having it in the image, but let's ask @bkchr if they want `miri` test in substrate/polkadot? As an alternative, I could do a separate CI image for this project.

TriplEight: To illustrate that "lag behind" point: one cannot simply `rustup +nightly component add miri` in the pipeline and expect it to be installed every time. The CI image is updated nightly and sometimes that command will fail.
```
# rustup +nightly component add miri
error: component 'miri' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly
Sometimes not all components are available in any given nightly.
error: backtrace:
error:    0: error_chain::backtrace::imp::InternalBacktrace::new
   1: rustup::dist::manifestation::Manifestation::update
   2: rustup::toolchain::DistributableToolchain::add_component
   3: rustup::cli::rustup_mode::main
   4: rustup_init::main
   5: std::rt::lang_start_internal::{{closure}}::{{closure}}
             at /rustc/c7087fe00d2ba919df1d813c040a5d47e43b0fe7/src/libstd/rt.rs:52
      std::sys_common::backtrace::__rust_begin_short_backtrace
             at /rustc/c7087fe00d2ba919df1d813c040a5d47e43b0fe7/src/libstd/sys_common/backtrace.rs:130
   6: main
   7: __libc_start_main
   8: <unknown>
   ```
   Due to `miri` is not compatible with this nightly. As a result, you will have flaky pipelines.

**Comment by thiolliere** [comment]
I rerequest review because of the additional changes, but no pressure

Also I benchmarked the vec_decode_u8 between this and master:
* I don't see significative change when input is limited (`remaining_len` return some len).
* when input is not limited there is an improvment, because we use `set_len` now

File: src/bit_vec.rs [hunk 13]
```diff
@@ -42,6 +43,19 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitSlice<O, T> {
 	}
 }
 
+fn reverse_endian(vec_u8: &mut Vec<u8>, size_of_t: usize) {
```
bkchr, line 46: Some docs would be nice.

src/bit_vec.rs
```diff
@@ -50,21 +64,48 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
 impl<O: BitOrder, T: BitStore + FromByteSlice> Decode for BitVec<O, T> {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(bits)| {
 			let bits = bits as usize;
 			let required_bytes = required_bytes::<T>(bits);
 
-			let vec = read_vec_u8(input, required_bytes)?;
+			let mut vec_u8 = read_vec_from_u8s::<I, u8>(input, required_bytes)?;
+
+			if cfg!(target_endian = "big") && mem::size_of::<T>() != 1 {
```
bkchr, line 95:
```suggestion
			if cfg!(target_endian = "big") && mem::size_of::<T>() > 1 {
```

src/bit_vec.rs
```diff
@@ -50,21 +64,48 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
+
+			let mut aligned_vec: Vec<T> = vec![0u8.into(); required_bytes / mem::size_of::<T>()];
 
-			let mut result = Self::from_slice(T::from_byte_slice(&vec)?);
+			unsafe {
+				let aligned_u8_ptr = aligned_vec.as_mut_ptr() as *mut u8;
+				for (i, v) in vec_u8.iter().enumerate() {
+					*aligned_u8_ptr.add(i) = *v;
+				}
+			}
```
bkchr, line 106: Why don't we use the output of `read_vec_from_u8s` directly as input for `from_vec`?

src/bit_vec.rs
thiolliere: but then if user have implemented something generically using `BitStore` trait, then they will have to use our new trait no ?

src/bit_vec.rs
thiolliere: oh yes, I added some, does it need more ?

src/bit_vec.rs
thiolliere: so above we called `read_vec_from_u8s::<I, u8>`, we can't directly do `read_vec_from_u8s::<I, T>` (where T is `BitStore` thus one of u8, u16, ..., u64) because T doesn't implement `ToMutByteSlice`.

(it implements`Clone`, and we could get around the `Default` requirement in `read_vec_from_u8s`, so `ToMutByteSlice` is difficult without changing the signature `impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T>`).

So as a result of having used `u8` we might not have an aligned vec (at least under miri test condition). so here I just copy it to an aligned vec.

But anyway the mirit test is failing because of outdated bitvec. But I thought fixing the alignment would be better even if it involve an additional copy here.
Note that once we bump bitvec we can factorize it quite easily, removing this unsafe and decode a `Vec<T>` directly.

src/bit_vec.rs
bkchr: Ty

src/bit_vec.rs
bkchr: Makes sense! Ty for explaining it!

bkchr:
If we are here just blocked on miri, I propose that we integrate miri later when @TriplEight had time to create a docker image!

TriplEight:
@bkchr I was thinking maybe we'll need `miri` in substrate CI as well? So I'd add it in the main image and it will be lagging behind the latest nightly a bit? 
It's not a big deal, can also do a separate image for this repo, @thiolliere can you maybe list what tools do you not need from https://github.com/paritytech/scripts/tree/master/dockerfiles/ci-linux ?

bkchr:
> @bkchr I was thinking maybe we'll need `miri` in substrate CI as well? So I'd add it in the main image and it will be lagging behind the latest nightly a bit?
> It's not a big deal, can also do a separate image for this repo, @thiolliere can you maybe list what tools do you not need from https://github.com/paritytech/scripts/tree/master/dockerfiles/ci-linux ?

Also fine by me :D

thiolliere:
> @bkchr I was thinking maybe we'll need `miri` in substrate CI as well? So I'd add it in the main image and it will be lagging behind the latest nightly a bit?
> It's not a big deal, can also do a separate image for this repo, @thiolliere can you maybe list what tools do you not need from https://github.com/paritytech/scripts/tree/master/dockerfiles/ci-linux ?

you mean for the scale-codec repo or scale-codec + substrate ?

I only need the miri component, which is available only on rust nightlies toolchains.
To execute miri tests I think I need only this.
To execute parity-scale-codec CI I think I need:
* [?]  cargo-audit: I don't know, I think we don't need it for scale-codec repo, but maybe would be good to make use of it actually
* [?]  cargo-web: I don't think we need it also AFAICT
* [x]  sccache
* [ ]  wasm-pack
* [ ]  wasm-bindgen
* [?]  cargo-deny: I think we don't need this one, AFAICT
* [x]  wasm32-unknown-unknown toolchain
* [x]  miri component

src/bit_vec.rs
```diff
@@ -42,6 +43,22 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitSlice<O, T> {
...
+			vec_u8.swap(i * size_of_t + j, i * size_of_t + (size_of_t - 1) - j);
+		}
+	}
+}
+
+/// # WARNING
+///
+/// In bitvec v0.17.4 the only implementation of BitStore are u8, u16, u32, u64, usize.
+/// This implementation actually only support u8, u16, u32 and u64, encoding of with BitStore being
+/// usize is inconsistent between platforms.
```
dvdplm, line 61:
```suggestion
/// In bitvec v0.17.4 the only implementations of BitStore are u8, u16, u32, u64,and usize.
/// This implementation actually only support u8, u16, u32 and u64, as encoding of uszie
/// is inconsistent between platforms.
```

src/bit_vec.rs
```diff
@@ -50,21 +67,48 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
 	}
 }
 
 impl<O: BitOrder, T: BitStore + ToByteSlice> EncodeLike for BitVec<O, T> {}
 
+/// # WARNING
+///
+/// In bitvec v0.17.4 the only implementation of BitStore are u8, u16, u32, u64, usize.
+/// This implementation actually only support u8, u16, u32 and u64, encoding of with BitStore being
+/// usize is inconsistent between platforms.
```
dvdplm, line 89:
```suggestion
/// In bitvec v0.17.4 the only implementations of BitStore are u8, u16, u32, u64 and usize.
/// This implementation actually only support u8, u16, u32 and u64, as encoding of usize
/// is inconsistent between platforms.
```

src/bit_vec.rs
```diff
@@ -50,21 +67,48 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
 impl<O: BitOrder, T: BitStore + FromByteSlice> Decode for BitVec<O, T> {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(bits)| {
 			let bits = bits as usize;
 			let required_bytes = required_bytes::<T>(bits);
 
-			let vec = read_vec_u8(input, required_bytes)?;
+			let mut vec_u8 = read_vec_from_u8s::<I, u8>(input, required_bytes)?;
+
```
cheme, line 97: Would moving 'reverse_endian' on aligned vec be better (did we test miri with a big endian target, wondering if reverse_endian can trigger the warning?)?

dvdplm (commented):
a few documentation nitpicks

cheme (commented):
LGTM, added a few question (please do not read them as request).
I wonder do we have a way to test in a BE envt with CI?

src/codec.rs
```diff
@@ -827,39 +870,39 @@ impl<T: Decode> Decode for Vec<T> {
 		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 			let len = len as usize;
 
+			fn decode_unoptimized<I: Input, T: Decode>(
```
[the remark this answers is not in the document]
thiolliere: with_type_info will execute some specific code for when `T` is one of `U8,I8,U16,I16, U32, I32, U64, I64, U128, I128`. But here should execute the transmute when `T` is something else than one of these types, thus decode_unoptimized doesn't include the transmute

src/codec.rs
```diff
@@ -782,36 +794,67 @@ impl<T: Encode> Encode for [T] {
 	}
 }
 
-/// Read an `u8` vector from the given input.
-pub(crate) fn read_vec_u8<I: Input>(input: &mut I, len: usize) -> Result<Vec<u8>, Error> {
+/// Create a `Vec<T>` by casting directly from a buffer of read `u8`s
+///
+/// The encoding of `T` must be equal to its binary representation, and size of `T` must be less or
+/// equal to [`MAX_PREALLOCATION`].
+pub(crate) fn read_vec_from_u8s<I, T>(input: &mut I, items_len: usize) -> Result<Vec<T>, Error>
```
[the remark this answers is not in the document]
thiolliere: yes I think we can do it, we can probably improve some code in BE, but actually I wanted to do as least effort as possible to support BE platform for now, as I don't know any ppl who need it.

I also introduced reverse_endian in bitvec only because it was needed as `T` didn't even bound `Decode`. FYI in the follow-up PR I remove reverse_endian.

src/codec.rs
```diff
@@ -936,13 +979,19 @@ impl<T: Encode> Encode for VecDeque<T> {
...
+						core::mem::transmute::<(&[T], &[T]), (&[$ty], &[$ty])>(slices)
+					};
+
+					$dest.write(<[$ty] as AsByteSlice<$ty>>::as_byte_slice(typed.0));
+					$dest.write(<[$ty] as AsByteSlice<$ty>>::as_byte_slice(typed.1));
+				} else {
+					for item in $self {
+						item.encode_to($dest);
+					}
+				}
```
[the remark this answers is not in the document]
thiolliere: Yes we could I think we can, maybe we can do in later PR if ppl ask for it.

src/bit_vec.rs
thiolliere: reverse_endian is just changing bytes in a `&mut [u8]` so I think it is fine here.
I havn't run tests in BE actually :-|

src/codec.rs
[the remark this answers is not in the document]
cheme: makes sense thanks.

src/bit_vec.rs
```diff
@@ -42,6 +43,22 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitSlice<O, T> {
 	}
 }
 
+/// Reverse bytes of element for element of size `size_of_t`.
+///
+/// E.g. if size is 2 `[1, 2, 3, 4]` is changed to `[2, 1, 4, 3]`.
+fn reverse_endian(vec_u8: &mut Vec<u8>, size_of_t: usize) {
```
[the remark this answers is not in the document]
thiolliere: yes done in https://github.com/paritytech/parity-scale-codec/pull/231/commits/cb5128681062835e7f751703d8cc845cc04d0a70

src/bit_vec.rs
```diff
@@ -42,6 +43,22 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitSlice<O, T> {
...
+	for i in 0..vec_u8.len() / size_of_t {
+		for j in 0..size_of_t / 2 {
+			vec_u8.swap(i * size_of_t + j, i * size_of_t + (size_of_t - 1) - j);
```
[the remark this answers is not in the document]
thiolliere: actually I find the way I write more understandable to me, but I guess this is personal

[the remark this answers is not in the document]
cheme: I am fine with it :)

src/codec.rs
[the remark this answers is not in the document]
cheme: Is there a follow up pr? (or in the future)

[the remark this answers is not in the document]
thiolliere: in the future, but should be soon ready

cheme (approved):

src/codec.rs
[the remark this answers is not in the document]
thiolliere: it is here, https://github.com/paritytech/parity-scale-codec/pull/236

Thus the removal of reverse endian, (and probably bitvec performance regression on BE)

PR #233, opened 2020-12-31
Docs and cleanup for utils

dvdplm:
Document public items and rename a few utility methods. Simplifies the signature of `collect_types` a little bit to remove the variant filter (was always the same) and related changes to the type filters.

### DIFF
--- a/derive/src/decode.rs
+++ b/derive/src/decode.rs
@@ -35,7 +35,7 @@ pub fn quote(data: &Data, type_name: &Ident, input: &TokenStream) -> TokenStream
 			},
 		},
 		Data::Enum(ref data) => {
-			let data_variants = || data.variants.iter().filter(|variant| crate::utils::get_skip(&variant.attrs).is_none());
+			let data_variants = || data.variants.iter().filter(|variant| !utils::should_skip(&variant.attrs));
 
 			if data_variants().count() > 256 {
 				return Error::new(
@@ -76,8 +76,8 @@ pub fn quote(data: &Data, type_name: &Ident, input: &TokenStream) -> TokenStream
 
 fn create_decode_expr(field: &Field, name: &str, input: &TokenStream) -> TokenStream {
 	let encoded_as = utils::get_encoded_as_type(field);
-	let compact = utils::get_enable_compact(field);
-	let skip = utils::get_skip(&field.attrs).is_some();
+	let compact = utils::is_compact(field);
+	let skip = utils::should_skip(&field.attrs);
 
 	let res = quote!(__codec_res_edqy);
 
--- a/derive/src/decode.rs
+++ b/derive/src/decode.rs
@@ -46,7 +46,7 @@ pub fn quote(data: &Data, type_name: &Ident, input: &TokenStream) -> TokenStream
 
 			let recurse = data_variants().enumerate().map(|(i, v)| {
 				let name = &v.ident;
-				let index = utils::index(v, i);
+				let index = utils::get_variant_index(v, i);
 
 				let create = create_instance(
 					quote! { #type_name :: #name },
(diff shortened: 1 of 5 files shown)


derive/src/decode.rs
```diff
@@ -46,7 +46,7 @@ pub fn quote(data: &Data, type_name: &Ident, input: &TokenStream) -> TokenStream
 
 			let recurse = data_variants().enumerate().map(|(i, v)| {
 				let name = &v.ident;
-				let index = utils::index(v, i);
+				let index = utils::get_variant_index(v, i);
```
thiolliere, line 49: as you are cleaning up, isn't `variant_index` more rust idiomatic than `get_variant_index` ?
Anyway i don't mind either.

derive/src/utils.rs
```diff
@@ -42,8 +43,11 @@ fn find_meta_item<'a, F, R, I>(itr: I, pred: F) -> Option<R> where
 	}).next()
 }
 
-pub fn index(v: &Variant, i: usize) -> TokenStream {
-	// look for an index in attributes
+/// Look for a `#[scale(index = $int)]` attribute on a variant. If no attribute
+/// is found, fall back to the discriminant or just the variant index.
+/// Returns a `TokenStream`.
```
thiolliere, line 48: I usually prefer not to write anything, if we don't add any information on the type itself
```suggestion
```

derive/src/utils.rs
```diff
@@ -85,9 +90,10 @@ pub fn get_encoded_as_type(field_entry: &Field) -> Option<TokenStream> {
 	})
 }
 
-pub fn get_enable_compact(field_entry: &Field) -> bool {
-	// look for `encode(compact)` in the attributes
-	find_meta_item(field_entry.attrs.iter(), |meta| {
+/// Look for a `#[codec(compact)]` outer attribute on the given `Field`.
+/// Returns true or false.
```
thiolliere, line 94: same as above
```suggestion
```

derive/src/utils.rs
```diff
@@ -98,9 +104,9 @@ pub fn get_enable_compact(field_entry: &Field) -> bool {
 	}).is_some()
 }
 
-// return span of skip if found
-pub fn get_skip(attrs: &[Attribute]) -> Option<Span> {
-	// look for `skip` in the attributes
+/// Look for a `#[codec(skip)]` in the given attributes.
+/// Returns true or false.
```
thiolliere, line 108: same as above
```suggestion
```

derive/src/utils.rs
```diff
@@ -109,11 +115,12 @@ pub fn get_skip(attrs: &[Attribute]) -> Option<Span> {
...
 
 		None
-	})
+	}).is_some()
 }
 
-/// Returns if the `dumb_trait_bound` attribute is given in `attrs`.
-pub fn get_dumb_trait_bound(attrs: &[Attribute]) -> bool {
+/// Look for a `#[codec(dumb_trait_bound)]`in the given attributes.
+/// Returns true or false.
```
thiolliere, line 122: same as above
```suggestion
```

thiolliere (approved):
looks good to me, I as intended don't see any logic change.

derive/src/decode.rs
dvdplm: Good pt, changing.

derive/src/utils.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Parity Technologies
+// Copyright 2020 Parity Technologies
```
bkchr, line 1:
```suggestion
// Copyright 2018-2020 Parity Technologies
```

bkchr (approved):

PR #236, opened 2021-01-06
Bump bitvec, byte slice cast, and refactor a bit

thiolliere:
status: Reviewable

I'm a bit unsure yet, if it is a good idea to implement Codec on `BitVec<_, T>` when `T` (`BitStore`) implements Codec.
We have the method `required_elements` (previously `required_bytes`) which computes for a type `T` the number of element `T` needed to store N `bits`. But I wonder if some unexpected implementation of BitStore would break assumption of `required_elements`.

I tried to find a method in BitStore to have the equivalent of `required_elements` but couldn't fine for now.

One possibility could be to implement for specific BitStore types: i.e. `u8, u16, u32, u64, Cell<u8>, ...` so that no unexpected implementation can happen. (this is trivial to do, but maybe annoying for user not to have it automatically implemented)

EDIT: I see we can use `<T as BitMemory>::BITS` to get the size of a segment. So I refactored `required_elements` to make use of it.

### DIFF
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -63,7 +63,7 @@ impl<O: BitOrder, T: BitStore + Decode> Decode for BitVec<O, T> {
 
 			let mut result = Self::try_from_vec(vec)
 				.map_err(|_| {
-					Error::from("UNEXPECTED ERROR: `bits` is less than
+					Error::from("UNEXPECTED ERROR: `bits` is less or equal to
 					`ARCH32BIT_BITSLICE_MAX_BITS`; So BitVec must be able to handle the number of
 					segment needed for `bits` to be represented; qed")
 				})?;
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -30,8 +30,9 @@ impl<O: BitOrder, T: BitStore + Encode> Encode for BitSlice<O, T> {
 		);
 		Compact(len as u32).encode_to(dest);
 
-		// NOTE: from `BitSlice::as_slice`: "The returned slice handle views all elements touched
-		// by self"
+		// NOTE: doc of `BitSlice::as_slice`:
+		// > The returned slice handle views all elements touched by self
+		//
 		// Thus we are sure the slice doesn't contain unused elements at the end.
 		let slice = self.as_slice();
 
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -69,7 +69,7 @@ impl<O: BitOrder, T: BitStore + Decode> Decode for BitVec<O, T> {
 				})?;
 
 			assert!(bits as usize <= result.len());
-			unsafe { result.set_len(bits as usize); }
+			result.truncate(bits as usize);
 			Ok(result)
 		})
 	}
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -15,51 +15,12 @@
 //! `BitVec` specific serialization.
 
 use core::mem;
-use crate::alloc::vec::Vec;
-
 use bitvec::{vec::BitVec, store::BitStore, order::BitOrder, slice::BitSlice, boxed::BitBox};
-use byte_slice_cast::{AsByteSlice, ToByteSlice, FromByteSlice, Error as FromByteSliceError};
-
-use crate::codec::{Encode, Decode, Input, Output, Error, read_vec_from_u8s};
+use crate::codec::{Encode, Decode, Input, Output, Error, decode_vec_with_len, encode_slice_no_len};
 use crate::compact::Compact;
 use crate::EncodeLike;
 
-impl From<FromByteSliceError> for Error {
-	fn from(e: FromByteSliceError) -> Error {
-		match e {
-			FromByteSliceError::AlignmentMismatch {..} =>
-				"failed to cast from byte slice: alignment mismatch".into(),
-			FromByteSliceError::LengthMismatch {..} =>
-				"failed to cast from byte slice: length mismatch".into(),
-			FromByteSliceError::CapacityMismatch {..} =>
-				"failed to cast from byte slice: capacity mismatch".into(),
-		}
-	}
-}
-
-impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitSlice<O, T> {
-	fn encode_to<W: Output>(&self, dest: &mut W) {
-		self.to_vec().encode_to(dest)
-	}
-}
-
-/// Reverse bytes of element for element of size `size_of_t`.
-///
-/// E.g. if size is 2 `[1, 2, 3, 4]` is changed to `[2, 1, 4, 3]`.
-fn reverse_endian(vec_u8: &mut [u8], size_of_t: usize) {
-	for i in 0..vec_u8.len() / size_of_t {
-		for j in 0..size_of_t / 2 {
-			vec_u8.swap(i * size_of_t + j, i * size_of_t + (size_of_t - 1) - j);
-		}
-	}
-}
-
-/// # WARNING
-///
-/// In bitvec v0.17.4 the only implementations of BitStore are u8, u16, u32, u64, and usize.
-/// This implementation actually only support u8, u16, u32 and u64, as encoding of uszie
-/// is inconsistent between platforms.
-impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
+impl<O: BitOrder, T: BitStore + Encode> Encode for BitSlice<O, T> {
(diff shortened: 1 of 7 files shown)


src/bit_vec.rs
```diff
@@ -68,72 +30,75 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
-				for (i, v) in vec_u8.iter().enumerate() {
-					*aligned_u8_ptr.add(i) = *v;
-				}
+impl<O: BitOrder, T: BitStore + Decode> Decode for BitVec<O, T> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		<Compact<u32>>::decode(input).and_then(move |Compact(bits)| {
+			// Otherwise it is impossible to store it on 32bit machine.
+			if bits as usize > ARCH32BIT_BITSLICE_MAX_BITS {
+				return Err("Attempt to decode a bitvec with too many bits".into());
 			}
```
thiolliere, line 60: This is a constraint to ensure the behavior is same between 32bit and 64bit machine.

src/bit_vec.rs
```diff
@@ -68,72 +30,75 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
 		);
 		Compact(len as u32).encode_to(dest);
 
-		let byte_slice: &[u8] = self.as_slice().as_byte_slice();
+		// NOTE: doc of `BitSlice::as_slice`:
+		// > The returned slice handle views all elements touched by self
+		//
+		// Thus we are sure the slice doesn't contain unused elements at the end.
+		let slice = self.as_slice();
```
thiolliere, line 37: Here we could also check that len is less than `ARCH32BIT_BITSLICE_MAX_BITS` otherwise it won't be able to decode.

But between panicking and having an invalid data. I think it is fine to have an invalid data. Anyway having such a big BitSlice is likely to be an error by itself.

src/bit_vec.rs
```diff
@@ -68,72 +30,75 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
-			unsafe { result.set_len(bits); }
+			let required_elements = required_elements::<T>(bits)? as usize;
+			let vec = decode_vec_with_len(input, required_elements)?;
+
+			let mut result = Self::try_from_vec(vec)
+				.map_err(|_| {
+					Error::from("UNEXPECTED ERROR: `bits` is less than
+					`ARCH32BIT_BITSLICE_MAX_BITS`; So BitVec must be able to handle the number of
+					segment needed for `bits` to be represented; qed")
+				})?;
```
thiolliere, line 69: I'm not sure how to phrase it.
Basically we have ensure `bits` is less than `ARCH32BIT_BITSLICE_MAX_BITS`.
So the number of segment needed for `bits` should be less or equal to the number of segment needed for `ARCH32BIT_BITSLICE_MAX_BITS`. which must be less or equal to the maximum number of segment allowed.

src/bit_vec.rs
```diff
@@ -68,72 +30,75 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
+/// Calculates the number of element `T` required to store given amount of `bits` as if they were
+/// stored in `BitVec<_, T>`
+///
+/// Returns an error if the number of bits + number of bits in element overflow u32 capacity.
+/// NOTE: this should never happen if `bits` is already checked to be less than
+/// `BitStore::MAX_BITS`.
+fn required_elements<T: BitStore>(bits: u32) -> Result<u32, Error> {
+	let element_bits = T::Mem::BITS as u32;
+	let error = Error::from("Attempt to decode bitvec with too many bits");
+	Ok((bits.checked_add(element_bits).ok_or_else(|| error)?  - 1) / element_bits)
```
thiolliere, line 101: This checked_add is just extra safety but I'm ok to remove it with just a comment.

src/codec.rs
```diff
@@ -601,6 +601,90 @@ macro_rules! impl_for_non_zero {
 		)*
 	}
 }
+
+/// Encode the slice without prepending the len.
+///
+/// This is equivalent to encoding all the element one by one, but it is optimized for some types.
+pub(crate) fn encode_slice_no_len<T: Encode, W: Output>(slice: &[T], dest: &mut W) {
```
thiolliere, line 608: the code of this function is just copy pasted from encode of arrays which must be same as encode of `&[T]`.

src/codec.rs
```diff
@@ -601,6 +601,90 @@ macro_rules! impl_for_non_zero {
...
+			}
+		},
+	}
+}
+
+/// Encode the slice without prepending the len.
+///
+/// This is equivalent to encoding all the element one by one, but it is optimized in some
+/// situation.
+pub(crate) fn decode_vec_with_len<T: Decode, I: Input>(
```
thiolliere, line 647: This code is just copy pasted from decode of `Vec<T>` so it can be used by bitvec.

cheme (approved):
LGTM.

src/bit_vec.rs
thiolliere: I do this limitation because the size of BitVec is limited, see panic paragraph in https://docs.rs/bitvec/0.20.1/bitvec/vec/struct.BitVec.html#method.from_vec
it is not related to the BitStore type.

But I guess I'm overdoing here, and we could just fail to decode on 32bit plateform when bitvec doesn't handle it maybe. I wanted to make parity-scale-codec having the exact same behavior regardless if plateform is 32bit or 64bit

src/bit_vec.rs
thiolliere: but I wanted to insist that it is an internal error. Maybe it is actually better to panic with `expect("...; qed")`

src/bit_vec.rs
```diff
@@ -177,34 +142,33 @@ mod tests {
...
+
+		assert_eq!(Ok(0), required_elements::<u64>(0));
+		assert_eq!(Ok(1), required_elements::<u64>(1));
+		assert_eq!(Ok(1), required_elements::<u64>(63));
+		assert_eq!(Ok(1), required_elements::<u64>(64));
+		assert_eq!(Ok(2), required_elements::<u64>(65));
 	}
 
 	#[test]
-	#[cfg_attr(miri, ignore)] // BitVec error due to outdated version of bitvec
```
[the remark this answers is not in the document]
thiolliere: yes but compiled with `release` it take one or two minutes I think no ?

src/bit_vec.rs
cheme: As long as the user understood it is a limitation from the bitvec crate, it is fine.

src/bit_vec.rs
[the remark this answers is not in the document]
cheme: didn't try with release, that is a good point (still one or two minutes is not really fast, but I may not run it locally everyday so I won't complain).

src/codec.rs
```diff
@@ -601,6 +601,90 @@ macro_rules! impl_for_non_zero {
...
+			for item in slice.iter() {
+				item.encode_to(dest);
+			}
+		},
+	}
+}
+
+/// Encode the slice without prepending the len.
+///
+/// This is equivalent to encoding all the element one by one, but it is optimized in some
```
bkchr, line 645:
```suggestion
/// This is equivalent to decode all elements one by one, but it is optimized in some
```

bkchr (approved):

src/codec.rs
```diff
@@ -601,6 +601,90 @@ macro_rules! impl_for_non_zero {
...
+		encode_to(slice, dest),
+		{
+			for item in slice.iter() {
+				item.encode_to(dest);
+			}
+		},
+	}
+}
+
+/// Encode the slice without prepending the len.
```
thiolliere, line 643:
```suggestion
/// Decode the slice (without prepended the len).
```

src/bit_vec.rs
```diff
@@ -68,72 +30,75 @@ impl<O: BitOrder, T: BitStore + ToByteSlice> Encode for BitVec<O, T> {
...
-
-			let mut result = Self::from_vec(aligned_vec);
-			assert!(bits <= result.len());
-			unsafe { result.set_len(bits); }
+			let required_elements = required_elements::<T>(bits)? as usize;
+			let vec = decode_vec_with_len(input, required_elements)?;
+
+			let mut result = Self::try_from_vec(vec)
+				.map_err(|_| {
+					Error::from("UNEXPECTED ERROR: `bits` is less than
```
thiolliere, line 66:
```suggestion
					Error::from("UNEXPECTED ERROR: `bits` is less or equal to
```

PR #239, opened 2021-01-13
make `Output` trait object and encode_to allowed on not sized `Output`

thiolliere:
### PR description:

This won't allow us to have `Output` not trait object in the future.

I'm not sure how much needed this is. But having a trait object Output seems valuable.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -219,6 +219,15 @@ pub trait Output {
 	fn push_byte(&mut self, byte: u8) {
 		self.write(&[byte]);
 	}
+
+	/// Write encoding of given value to the output.
+	///
+	/// NOTE: this is only available when `Self` is `Sized`, thus it is not available in
+	/// `Encode::encode_to`. If you need it when `Self` is not `Sized` consider using
+	/// `Encode::encode_to` directly.
+	fn push<V: Encode + ?Sized>(&mut self, value: &V) where Self: Sized {
+		value.encode_to(self);
+	}
 }
 
 #[cfg(not(feature = "std"))]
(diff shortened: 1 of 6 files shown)


src/codec.rs
```diff
@@ -221,7 +221,11 @@ pub trait Output: Sized {
 	}
 
 	/// Write encoding of given value to the output.
-	fn push<V: Encode + ?Sized>(&mut self, value: &V) {
+	///
+	/// NOTE: this is only available when `Self` is `Sized`, thus it is not available in
+	/// `Encode::encode_to`. If you need it when `Self` is not `Sized` consider using
+	/// `Encode::encode_to` directly.
+	fn push<V: Encode + ?Sized>(&mut self, value: &V) where Self: Sized {
```
bkchr, line 228: Isn't the trick here to just take `value: &dyn Encode` to make this available on the trait? From my side we could also drop this function. 

Or there is another trick, we can move this into some own trait and impl this trait for `dyn Output` :D

bkchr (approved):

src/codec.rs
thiolliere: dyn Encode doesn't work as Encode cannot be made into trait object

But implementing a trait PushOutput to dyn Output works, but then it is no longer implemented on not-dyn Output (I think),
I guess we can have both,

Anyway I removed it because it brings no help and might confuse some people

thiolliere:
I'll prepare an understandable Changelog or release note for 2.0, so that ppl can upgrade without too much difficulty, in this context removing `push` is better to me than keeping no helpful function.

Issue #241, opened 2021-01-18
Introduce function `fn fix_size() -> Option<usize>` on Decode.


thiolliere:
Maybe we could introduce on decode:
```
/// Return some if the type encoding has a fix length.
fn fix_size() -> Option<usize>
```
Then when implementing Decode::skip for `Vec<T>` or `Option<T>`, we could make use of this information to skip a number of elements efficiently.

Currently decode for `Vec<T>` is optimized scale_info types (u8, i8, u16, ... i128). This would allow to optimize skip for more types.

One idea is to just have introduce it with a default implementation, even if we don't use it yet for skip optimization we can do later if we have important usecases:
```
/// Return some if the type encoding has a fix length.
fn fix_size() -> Option<usize> { None }
```

Issue #201, opened 2020-04-07
Add `skip()`


bkchr:
The `Decode` trait should get a method `skip(input: &mut Input)` to skip an encoded object.

thiolliere:
we do agree that skipping an encoded object shouldn't check that the object is valid.

So if someone works on untrusted data, skipping some part of the data to get the wanted part can succeed even if the data will actually fail to decode.

bkchr:
We don't really check if an object is valid?

But yeah, normally we just skip the amount of expected bytes.

thiolliere:
Maybe we should also introduce a new function:
```rust
/// Return the length of the type if fixed.
fn length() -> Option<usize> { None }
```

this would be used to optimized implementation of `Vec<T>::skip` and some other types
```rust
if let Some(size) = T::length {
  // decode compact and skip size*compact
}
```

thiolliere:
skip is now part of Decode trait. I think we can close this issue in favor of https://github.com/paritytech/parity-scale-codec/issues/244
The idea is to release 2.0 soon so it can goes into next substrate release

PR #243, opened 2021-01-21
Improve Decode trait, without improving implementations

thiolliere:
related https://github.com/paritytech/parity-scale-codec/issues/241
related https://github.com/paritytech/parity-scale-codec/issues/201

This introduces 2 new methods on Decode trait, with default implementations.

Their optimization for the "primitive" types will be done in later PR, and maybe later minor version.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -314,11 +314,11 @@ pub trait Decode: Sized {
 		Self::decode(input).map(|_| ())
 	}
 
-	/// Return the fix length of the encoded type.
+	/// Return the fix size of the encoded type, if any.
 	///
-	/// If it returns some length then it is the length of any encoded value of the type.
-	/// Otherwise the type may or may not have fixed encoded length.
-	fn fix_encoded_len() -> Option<usize> {
+	/// If it returns some size then it is the size of any encoded value of the type.
+	/// Otherwise the type may or may not have fixed encoded size.
+	fn fix_encoded_size() -> Option<usize> {
 		None
 	}
 }
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -317,7 +317,7 @@ pub trait Decode: Sized {
 		Self::decode(input).map(|_| ())
 	}
 
-	/// Allows to return the fixed encoded size of the type.
+	/// Allows to return the fixed codec size of the type.
 	///
 	/// If it returns `Some(size)` then any possible value of this
 	/// type has the given size (in bytes) when being encoded.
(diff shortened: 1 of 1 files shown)


src/codec.rs
```diff
@@ -307,7 +307,20 @@ pub trait Decode: Sized {
...
+		Self::decode(input).map(|_| ())
+	}
+
+	/// Return the fix size of the encoded type, if any.
+	///
+	/// If it returns some size then it is the size of any encoded value of the type.
+	/// Otherwise the type encoding may or may not have fixed size.
+	fn fix_encoded_size() -> Option<usize> {
+		None
+	}
```
bkchr, line 323:
```suggestion
	/// Returns `Some(_)` iff the type has a fixed encoded size.
	///
	/// This means any possible encoded value of this type has 
	/// the same size when being encoded.
	fn encoded_fixed_size() -> Option<usize> {
		None
	}
```

bkchr (approved):

src/codec.rs
thiolliere: `iff` means it is an equivalence https://en.wikipedia.org/wiki/If_and_only_if
But here (at least because of the default implementation) some type can have a fixed encoded size and return None.

So the doc should say: `if it returns some then it is the fixed encoded size` or maybe but I'm not sure: `Returns some only if the type has fixed encoded size`.
I'm struggling to find a good way to express it

thiolliere: I gave it a try in https://github.com/paritytech/parity-scale-codec/pull/243/commits/3f3a43298cf849820dc48b14646acbb579d5c55e

src/codec.rs
```diff
@@ -307,7 +307,25 @@ pub trait Decode: Sized {
...
+	fn skip<I: Input>(input: &mut I) -> Result<(), Error> {
+		Self::decode(input).map(|_| ())
+	}
+
+	/// Allows to return the fixed codec size of the type.
+	///
+	/// If it returns `Some(size)` then any possible value of this
+	/// type has the given size (in bytes) when being encoded.
+	///
+	/// NOTE: A type with a fixed encoded size may return `None`.
```
bkchr, line 325:
```suggestion
	/// Returns the fixed encoded size of the type.
	///
	/// If it returns `Some(size)` then any possible value of this
	/// type has the given size (in bytes) when being encoded.
```

bkchr (approved):
Some last nits :P

src/codec.rs
dvdplm, line 325:
```suggestion
	/// Returns the fixed encoded size of the type.
	///
	/// If it returns `Some(size)` then all possible values of this
	/// type have the given size (in bytes) when encoded.
	///
	/// NOTE: A type with a fixed encoded size may return `None`.
```

thiolliere: merged through David suggestion

PR #245, opened 2021-01-23
add EncodedSize that calculates the encoded size more efficiently.

laizy:
used when the encoded bytes is not interested.
This could replace `Encode::using_encoded(|v| v.len())` used in substrate to avoid unnecessary memory allocation and data copy.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -290,6 +290,27 @@ pub trait Encode {
 	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
 		f(&self.encode())
 	}
+
+	/// Calculates the encoded size, used when the encoded bytes is not interested.
+	fn encoded_size(&self) -> usize {
+		let mut size_tracker = SizeTracker { writen: 0 };
+		self.encode_to(&mut size_tracker);
+		size_tracker.writen
+	}
+}
+
+struct SizeTracker {
+	writen: usize,
+}
+
+impl Output for SizeTracker {
+	fn write(&mut self, bytes: &[u8]) {
+		self.writen += bytes.len();
+	}
+
+	fn push_byte(&mut self, _byte: u8) {
+		self.writen += 1;
+	}
 }
 
 /// Trait that allows the length of a collection to be read, without having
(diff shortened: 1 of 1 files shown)


thiolliere:
At some point we can put this function in the trait `Encode`, so that default implementation can be optimized, isn't it better ?

laizy:
@thiolliere , yes, it will be more user friendly and accessable if implemented in `Encode` directly, I can change that in this pr if you want.

thiolliere:
Yes IMHO that would be better, just having it in the trait and with the default implementation, no need to optimize it for primitive types (at least not yet).

src/codec.rs
```diff
@@ -290,6 +290,27 @@ pub trait Encode {
 	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
 		f(&self.encode())
 	}
+
+	/// Calculates the encoded size, used when the encoded bytes is not interested.
```
thiolliere, line 294:
```suggestion
	/// Calculates the encoded size, used when the encoded bytes are not interesting.
```

src/codec.rs
```diff
@@ -290,6 +290,27 @@ pub trait Encode {
...
+	fn encoded_size(&self) -> usize {
+		let mut size_tracker = SizeTracker { writen: 0 };
+		self.encode_to(&mut size_tracker);
+		size_tracker.writen
+	}
+}
+
+struct SizeTracker {
+	writen: usize,
```
thiolliere, line 303:
```suggestion
	written: usize,
```

src/codec.rs
```diff
@@ -290,6 +290,27 @@ pub trait Encode {
 	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
 		f(&self.encode())
 	}
+
+	/// Calculates the encoded size, used when the encoded bytes is not interested.
+	fn encoded_size(&self) -> usize {
+		let mut size_tracker = SizeTracker { writen: 0 };
+		self.encode_to(&mut size_tracker);
+		size_tracker.writen
```
thiolliere, line 298:
```suggestion
		size_tracker.written
```

src/codec.rs
```diff
@@ -290,6 +290,27 @@ pub trait Encode {
...
+	}
+}
+
+struct SizeTracker {
+	writen: usize,
+}
+
+impl Output for SizeTracker {
+	fn write(&mut self, bytes: &[u8]) {
+		self.writen += bytes.len();
```
thiolliere, line 308:
```suggestion
		self.written += bytes.len();
```

src/codec.rs
```diff
@@ -290,6 +290,27 @@ pub trait Encode {
...
+	}
+
+	fn push_byte(&mut self, _byte: u8) {
+		self.writen += 1;
```
thiolliere, line 312:
```suggestion
		self.written += 1;
```

src/codec.rs
```diff
@@ -290,6 +290,27 @@ pub trait Encode {
...
+
+	/// Calculates the encoded size, used when the encoded bytes is not interested.
+	fn encoded_size(&self) -> usize {
+		let mut size_tracker = SizeTracker { writen: 0 };
+		self.encode_to(&mut size_tracker);
+		size_tracker.writen
+	}
+}
+
+struct SizeTracker {
```
thiolliere, line 302:
```suggestion
// Implements `Output` and only keeps track of the number of written bytes
struct SizeTracker {
```

src/codec.rs
```diff
@@ -290,6 +290,27 @@ pub trait Encode {
 	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
 		f(&self.encode())
 	}
+
+	/// Calculates the encoded size, used when the encoded bytes is not interested.
+	fn encoded_size(&self) -> usize {
+		let mut size_tracker = SizeTracker { writen: 0 };
```
thiolliere, line 296:
```suggestion
		let mut size_tracker = SizeTracker { written: 0 };
```

thiolliere (approved):
some nits, but looks good otherwise

src/codec.rs
```diff
@@ -290,6 +290,28 @@ pub trait Encode {
 	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
 		f(&self.encode())
 	}
+
+	/// Calculates the encoded size, used when the encoded bytes are not interesting.
```
bkchr, line 294:
```suggestion
	/// Calculates the encoded size.
	///
	/// Should be used when the encoded data isn't required.
	///
	/// # Note
	///
	/// This works by using a special [`Output`] that only tracks the size. So, there are no allocations inside the 
	/// output. However, this can not prevent allocations that some types are doing inside their own encoding. 
```

bkchr (approved):

PR #250, opened 2021-02-15
Hackfix for funty breaking change

dvdplm:
Closes #249

The latest `funty` release seems to be backwards incompatible and as `bitvec` imports `functy = "1"` projects can start to break without notice. This PR doesn't address the issue but can work as a temporary solution.﻿

PR #253, opened 2021-02-24
Upgrade bitvec to 0.21 to unbreak the build

### LINKED ISSUE #249: funty 1.2 breaks constant resolution
`parity-scale-codec` depends on `bitvec` which depends on `funty` [at version `"1"`](https://github.com/bitvecto-rs/bitvec/blob/develop/Cargo.toml#L59). In projects that pull in the latest version, 1.2, there is a problem, e.g.:

```
error[E0034]: multiple applicable items in scope
   --> /Users/aggron/.cargo/registry/src/github.com-1ecc6299db9ec823/bitvec-0.20.1/src/order.rs:476:7
    |
476 |             R::BITS,
    |                ^^^^ multiple `BITS` found
    |
note: candidate #1 is defined in the trait `BitMemory`
   --> /Users/aggron/.cargo/registry/src/github.com-1ecc6299db9ec823/bitvec-0.20.1/src/mem.rs:44:2
    |
44  |     const BITS: u8 = mem::size_of::<Self>() as u8 * 8;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: candidate #2 is defined in the trait `IsNumber`
   --> /Users/aggron/.cargo/registry/src/github.com-1ecc6299db9ec823/funty-1.2.0/src/lib.rs:144:2
    |
144 |     const BITS: u32;
    |     ^^^^^^^^^^^^^^^^
help: disambiguate the associated constant for candidate #1
    |
476 |             BitMemory::BITS,
    |             ^^^^^^^^^^^^^^^
help: disambiguate the associated constant for candidate #2
    |
476 |             IsNumber::BITS,
    |             ^^^^^^^^^^^^^^
```

(See for instance [this build of `parity-common`](https://ci.appveyor.com/project/paritytech/parity-common/builds/37773927) for full error)

The issue can be reproduced easily by adding `funty = "1.2"` to `parity-scale-codec`'s Cargo.toml.

Upstream issue [here](https://github.com/myrrlyn/funty/issues/3).

gnunicorn:
alternative to #250 .
Closes #249 .

bitvec (the reason we have funty in the dependency tree in the first place) has meanwhile released an upgraded version that solves the fix without pinning.

### DIFF
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -15,8 +15,10 @@
 //! `BitVec` specific serialization.
 
 use bitvec::{
-	vec::BitVec, store::BitStore, order::BitOrder, slice::BitSlice, boxed::BitBox, mem::BitMemory
+	vec::BitVec, store::BitStore, order::BitOrder, slice::BitSlice, boxed::BitBox,
 };
+use bitvec::macros::internal::funty::IsNumber;
+
 use crate::{
 	EncodeLike, Encode, Decode, Input, Output, Error, Compact,
 	codec::{decode_vec_with_len, encode_slice_no_len},
@@ -35,7 +37,7 @@ impl<O: BitOrder, T: BitStore + Encode> Encode for BitSlice<O, T> {
 		// > The returned slice handle views all elements touched by self
 		//
 		// Thus we are sure the slice doesn't contain unused elements at the end.
-		let slice = self.as_slice();
+		let slice = self.as_raw_slice();
 
 		encode_slice_no_len(slice, dest)
 	}
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -17,7 +17,7 @@
 use bitvec::{
 	vec::BitVec, store::BitStore, order::BitOrder, slice::BitSlice, boxed::BitBox,
 };
-use bitvec::macros::internal::funty::IsNumber;
+use funty::IsNumber;
 
 use crate::{
 	EncodeLike, Encode, Decode, Input, Output, Error, Compact,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,14 +4,18 @@ All notable changes to this crate are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this crate adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+- [Fixing build with latest `bitvec`/`funty`](https://github.com/paritytech/parity-scale-codec/pull/253). If you are using the `bitvec` make sure to upgrade it to `0.21.0` to for the correct traits to be implemented.
+
 ## [2.0.0] - 2021-01-26
 
 ### Added
 - `Decode::skip` allows to skip some encoded types. Pr #243
 - `Decode::encoded_fixed_size` allows to get the fixed encoded size of a type. PR #243
 - `Error` now contains a chain of causes. This full error description can also be activated on
-  no std using the feature `chain-error`. PR #242 
-- `Encode::encoded_size` allows to get the encoded size of a type more efficiently. PR #245 
+  no std using the feature `chain-error`. PR #242
+- `Encode::encoded_size` allows to get the encoded size of a type more efficiently. PR #245
 
 ### Changed
 - `CompactAs::decode_from` now returns result. This allow for decoding to fail from their compact
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ arrayvec = { version = "0.5.1", default-features = false, features = ["array-siz
 serde = { version = "1.0.102", optional = true }
 parity-scale-codec-derive = { path = "derive", version = "2.0.0", default-features = false, optional = true }
 bitvec = { version = "0.21.0", default-features = false, features = ["alloc"], optional = true }
+funty  = { version = "1.2.0", default-features = false, optional = true}
 byte-slice-cast = { version = "1.0.0", default-features = false }
 generic-array = { version = "0.14.4", optional = true }
 arbitrary = { version = "0.4.1", features = ["derive"], optional = true }
@@ -34,7 +35,7 @@ bench = false
 default = ["std"]
 derive = ["parity-scale-codec-derive"]
 std = ["serde", "bitvec/std", "byte-slice-cast/std", "chain-error"]
-bit-vec = ["bitvec"]
+bit-vec = ["bitvec", "funty"]
 fuzz = ["std", "arbitrary"]
 
 # Make error fully descriptive with chaining error message.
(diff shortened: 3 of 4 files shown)


src/bit_vec.rs
```diff
@@ -15,8 +15,10 @@
 //! `BitVec` specific serialization.
 
 use bitvec::{
-	vec::BitVec, store::BitStore, order::BitOrder, slice::BitSlice, boxed::BitBox, mem::BitMemory
+	vec::BitVec, store::BitStore, order::BitOrder, slice::BitSlice, boxed::BitBox,
 };
+use bitvec::macros::internal::funty::IsNumber;
```
gnunicorn, line 20: This is what `cargo` suggested to resolve the Problem that `let element_bits = T::Mem::BITS as u32;` (line 100) doesn't work - and I couldn't find any good alternative in the docs. I am not convinced this is the best way, but not sure how to solve this otherwise...

bkchr: Nope. The solution is still to fix this upstream. And we will clearly not import something that has the name `internal`in it.

bkchr (changes requested):

dvdplm:
@gnunicorn in `parity-common` we did [this](https://github.com/paritytech/parity-common/blob/master/.travis.yml#L14) and on `scale-info` I suggested [this](https://github.com/paritytech/scale-info/blob/0426479f5db2fbf5f2bfb5a7404ca0b4eb5ae1a0/.github/workflows/rust.yml#L28-L53) to fix CI pending an actual upstream fix.

gnunicorn:
@dvdplm the problem is that this only fixes the local CI but not actually anyone adding `parity-scale-codec` into their dependencies. (the problem came up trying to test and release the contracts-pallet).

@bkchr I agree with the `internal` - the alternative is to depend on `funty 1.2.0` to get that trait in directly.
> . The solution is still to fix this upstream.

I am afraid we won't see any other "upstream fix" to this, considering that the problem arose from a trait defined in bitvec having the `BITS` (we depend on here), while then adding the same named `BITS` a trait that depended on causing the issue - and both crates being developed and maintained by _the same person_, who then released a [newer version of the dependent crate removing the field and instead referring to the dependency one](https://docs.rs/crate/bitvec/0.21.0/source/src/mem.rs).

I fear this _is_ the upstream fix. What other fix are you waiting for?

athei (approved):
I am not optimistic that there will be a better "upstream fix" soon.

bkchr (approved):
Yeah with the change to `0.21.0` it looks "okay".

However this is a breaking change, as we have this in our public interface.

I think in the future we need to get rid off this dependency.

gnunicorn:
by having in our "public interfaces" you mean we are implementing it for that specific version of bitvec? because I can't find us returning any bitvec in our API.

If so, arguably we could have two independent features that implement it for 0.20 and 0.21 respectively (where the earlier would have to pin to 1.1 of funty) - but I guess that's also not gonna help us here, as we activate it through the `std`-flag :( ...

bkchr:
> by having in our "public interfaces" you mean we are implementing it for that specific version of bitvec?

This means, that everyone that updates scale codec and also uses bitvec, will need to update their bitvec version. As we are the only ones using it, doing a `2.1.0` release should probably be fine.

gnunicorn:
> As we are the only ones using it, doing a 2.1.0 release should probably be fine.

Sounds reasonable. I am happy to add a changelog entry informing people that the dependency has upgraded and thus they also need to upgrade the bitvec dependency to get the support...

athei:
If we do a major version bump we could also remove the implicit addition of `bitvec` on `std`?

bkchr:
> If we do a major version bump we could also remove the implicit addition of `bitvec` when `std` is enabled?

If cargo would just support this...

thiolliere (approved):
PR is good to me

I'm ok to release 2.1, can you write this in the changelog.
I hope ppl will be fine with the breaking change.

gnunicorn:
@thiolliere already added a changelog entry in the last commit.

thiolliere:
Sorry to change my mind but actually I don't think we should publish 0.21 like this, some ppl can be affected by this breaking change and I think it can show bad quality image of our libraries.

I open a fix in bitvec which seems good to me https://github.com/bitvecto-rs/bitvec/pull/109

ConnorBP:
ahh i just ran into this frustrating build error, good to know it wasn't caused by me on my project.

athei:
> Sorry to change my mind but actually I don't think we should publish 0.21 like this, some ppl can be affected by this breaking change and I think it can show bad quality image of our libraries.
> 
> I open a fix in bitvec which seems good to me [bitvecto-rs/bitvec#109](https://github.com/bitvecto-rs/bitvec/pull/109)

This would then be released as a `bitvec` fix release? Otherwise we would been in the same situation, right?

thiolliere:
Yes I expect him to release 0.20.1 with it.
If he doesn't I think we can break semver and publish 2.1

thiolliere:
there is no answer from his side since 4 days,

note that if we release parity-scale-codec with bitvec 0.21, every user who did `funty = 1.1.0` will see bitvec not compiling (because bitvec-0.21.0 depends on `funty = "1"` while it must be `funty = 1.2` in order to compile).

I don't know how we push to have this 0.20.2 release.

I also think that if we use bitvec 0.21 now then nothing guarantee that he doesn't publish a new version of funty and never fix bitvec 0.21.

athei:
What about dropping the support for `bitvec`? We are not using it ourselves and this library seems to be trouble.

thiolliere:
> What about dropping the support for bitvec? We are not using it ourselves and this library seems to be trouble.

You mean dropping bitvec in parity-scale-codec 2.1, yes we can do that, and we can provide a few helper function to implement Encode/Decode for bitvec.

I think we use it in polkadot though for some parachain related code.

athei:
Yes I am talking about removing the dependency with `parity-scale-codev 2.1`. It seems to be a liability. Not sure how we handle the polkadot case then.

gnunicorn:
> I think we use it in polkadot though for some parachain related code.

[Polkadot uses it](https://github.com/paritytech/polkadot/search?q=bitvec), and needs these to be `codec::Encode`/`Decode`. But we could def. drop `bit-vec` from being activated through the `std`-feature and thus by default. 

https://github.com/paritytech/parity-scale-codec/blob/1beac25dff792f52a7173ee34db88cf89cf6b8ef/Cargo.toml#L34-L36

(would also reduce the dependency tree).

athei:
Isn't that the original problem? That cargo automatically adds a dependency when you activate a feature on it.

bkchr:
We can still remove `bitvec` from the `std` feature for now. Downstream than probably needs to enable `std` if required, but there isn't any better solution until weak features are stabilized.

bkchr:
As said above, I'm also for dropping bitvec in Polkadot. But that requires someone to write some small struct. Shouldn't be that hard.

thiolliere:
bitvec 0.20.2 is released fixing its dependency. (it now depends on funty = "~1.1") So the issue is solved

PR #261, opened 2021-03-31
Switch array impls to const generics

austinabell:
Necessary changes to make https://github.com/paritytech/ink/issues/752 happen

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -688,57 +688,32 @@ impl_for_non_zero! {
 	NonZeroU128,
 }
 
-macro_rules! impl_array {
-	( $( $n:expr, )* ) => {
-		$(
-			impl<T: Encode> Encode for [T; $n] {
-				fn size_hint(&self) -> usize {
-					mem::size_of::<T>() * $n
-				}
+impl<T: Encode, const N: usize> Encode for [T; N] {
+	fn size_hint(&self) -> usize {
+		mem::size_of::<T>() * N
+	}
 
-				fn encode_to<W: Output + ?Sized>(&self, dest: &mut W) {
-					encode_slice_no_len(&self[..], dest)
-				}
-			}
-
-			impl<T: Decode> Decode for [T; $n] {
-				fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-					let mut r = ArrayVec::new();
-					for _ in 0..$n {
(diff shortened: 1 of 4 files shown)


src/codec.rs
```diff
@@ -688,57 +688,32 @@ impl_for_non_zero! {
...
-					}
-				}
-			}
+impl<T: Decode, const N: usize> Decode for [T; N] {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let mut r = ArrayVec::new();
+		for _ in 0..N {
+			r.push(T::decode(input)?);
+		}
+		let i = r.into_inner();
```
bkchr, line 707:
```suggestion
```

src/codec.rs
```diff
@@ -688,57 +688,32 @@ impl_for_non_zero! {
...
-						Err(_) => Err("failed to get inner array from ArrayVec".into()),
-					}
-				}
-			}
+impl<T: Decode, const N: usize> Decode for [T; N] {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let mut r = ArrayVec::new();
+		for _ in 0..N {
+			r.push(T::decode(input)?);
+		}
```
bkchr, line 706:
```suggestion
		let mut array = ArrayVec::new();
		for _ in 0..N {
			array.push(T::decode(input)?);
		}
```

src/codec.rs
```diff
@@ -688,57 +688,32 @@ impl_for_non_zero! {
...
+		let i = r.into_inner();
 
-			impl<T: EncodeLike<U>, U: Encode> EncodeLike<[U; $n]> for [T; $n] {}
-		)*
+		match i {
+			Ok(a) => Ok(a),
+			Err(_) => Err("failed to get inner array from ArrayVec".into()),
+		}
```
bkchr, line 712:
```suggestion
		Ok(array.into_inner().expect("We decode `N` elements; qed"))
```

bkchr (approved):

src/codec.rs
austinabell: Can't apply this one because the error type (the `ArrayVec`) doesn't implement debug. Lmk if you want another form like `if let`

bkchr: Okay fine :)

PR #267, opened 2021-05-03
Implement `BoundedEncodedLength` trait and implementations on primitives

coriolinus:


src/bounded_encoded_length.rs
```diff
@@ -0,0 +1,73 @@
...
+use core::mem;
+use crate::Encode;
+
+/// Items implementing `BoundedEncodedLen` have a statically known maximum encoded size.
+///
+/// Some containers, such as `BoundedVec`, have enforced size limits and this trait
+/// can be implemented accurately. Other containers, such as `StorageMap`, do not have enforced size
+/// limits. For those containers, it is necessary to make a documented assumption about the maximum
+/// usage, and compute the max encoded length based on that assumption.
+pub trait BoundedEncodedLen: Encode {
```
bkchr, line 24: I still don't see why this should be implemented in scale. This is Substrate specific and not SCALE specific.

coriolinus: The short answer is that `Encode::size_hint` is toothless: it's a 'best-effort' thing with a useless default implementation. Still, we do have the need sometimes for a more realistic size hint, and that need is generalizable to all of SCALE, not just Substrate.

src/bounded_encoded_length.rs
```diff
@@ -78,3 +78,25 @@ impl<T: BoundedEncodedLen, const N: usize> BoundedEncodedLen for [T; N] {
 		T::max_encoded_len().saturating_mul(N)
 	}
 }
+
+impl<T: BoundedEncodedLen> BoundedEncodedLen for Option<T> {
+	fn max_encoded_len() -> usize {
+		1 + T::max_encoded_len()
```
shawntabrizi, line 84: everything should be saturating i think

coriolinus: Oops, right. https://github.com/paritytech/parity-scale-codec/pull/267/commits/5f886b45cebb31106c357ecc33a86b492c476a46

src/bounded_encoded_length.rs
bkchr: This is also "toothless" as what is the maximum size of a vec? 

As already said in element, I don't see this in SCALE.

coriolinus: Not implemented for `Vec` 😁. I'll move this work to the substrate repo instead though.

coriolinus:
Transferred to https://github.com/paritytech/substrate/pull/8720.

PR #269, opened 2021-05-24
Impl Decode/Encode for Range/RangeInclusive

c410-f3r:
Feel free to specify another deserialization/serialization format like `(T, '=', T)` if desirable.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -15,7 +15,14 @@
 //! Serialisation.
 
 use core::fmt;
-use core::{mem, ops::Deref, marker::PhantomData, iter::FromIterator, convert::TryFrom, time::Duration};
+use core::{
+	convert::TryFrom,
+	iter::FromIterator,
+	marker::PhantomData,
+	mem,
+	ops::{Deref, Range, RangeInclusive},
+	time::Duration,
+};
 use core::num::{
 	NonZeroI8,
 	NonZeroI16,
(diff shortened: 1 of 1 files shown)


src/codec.rs
```diff
@@ -1205,6 +1212,59 @@ impl Decode for Duration {
...
+impl<T> Encode for Range<T>
+where
+	T: Encode
+{
+	#[inline]
+	fn size_hint(&self) -> usize {
+		2 * mem::size_of::<T>()
+	}
+
+	#[inline]
```
bkchr, line 1224:
```suggestion
	fn size_hint(&self) -> usize {
		2 * mem::size_of::<T>()
	}

```

src/codec.rs
```diff
@@ -1205,6 +1212,59 @@ impl Decode for Duration {
...
+	fn encode(&self) -> Vec<u8> {
+		(self.start(), self.end()).encode()
+	}
```
bkchr, line 1253:
```suggestion
	fn size_hint(&self) -> usize {
		2 * mem::size_of::<T>()
	}

	fn encode(&self) -> Vec<u8> {
		(self.start(), self.end()).encode()
	}
```

bkchr (approved):

c410-f3r:
These functions are trivial one-line code candidates for `#[inline]` and sometimes the compiler generates layers of assembly jump indirections that could simply be inlined. In any case, all `#[inline]`s were removed.

bkchr:
I know what `inline` is doing. However, the compiler is most often smarter than any human and can already inline code if required.

PR #271, opened 2021-06-08
Improve macro hygiene of derives

Robbepop:
Just a small PR to improve macro hygiene in some places where ambiguities are rare but existing.
It might be that I overlooked some code segments. Please check.

Also one question on my side: What do we need the `extern crate` code generation for exactly?

### DIFF
--- a/derive/src/decode.rs
+++ b/derive/src/decode.rs
@@ -42,7 +42,7 @@ pub fn quote(
 			),
 			Fields::Unit => {
 				quote_spanned! { data.fields.span() =>
-					Ok(#type_name)
+					::core::result::Result::Ok(#type_name)
 				}
 			},
 		},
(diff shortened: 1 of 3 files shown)


bkchr (approved):

bkchr:
> Also one question on my side: What do we need the `extern crate` code generation for exactly?

Rust 2015. Aka not needed anymore :P

Robbepop:
> > Also one question on my side: What do we need the `extern crate` code generation for exactly?
> 
> Rust 2015. Aka not needed anymore :P

So do we still have this to support edition 2015 users? since macros are inlined into their crate this might be required in order to make a edition 2018 proc. macro work for edition 2015 users. Or don't we support edition 2015 so we could actually really remove this code?

bkchr:
IMHO we can remove it

PR #275, opened 2021-06-23
Mention new `#[codec(crate = ...)]` attribute for MaxEncodedLen derive

Xanewok:


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this crate adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 ### Added
 - `MaxEncodedLen` trait for items that have a statically known maximum encoded size. ([#268](https://github.com/paritytech/parity-scale-codec/pull/268))
+- `#[codec(crate = <path>)]` top-level attribute to be used with the new `MaxEncodedLen`
```
coriolinus, line 13:
```suggestion
- `#[codec(crate = "<path>")]` top-level attribute to be used with the new `MaxEncodedLen`
```

Right now it expects a string literal instead of a path item. I'm going to be changing that shortly, but for now let's keep the changelog accurate.

coriolinus (approved):

CHANGELOG.md
Xanewok: Makes sense, thanks

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this crate adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 ### Added
 - `MaxEncodedLen` trait for items that have a statically known maximum encoded size. ([#268](https://github.com/paritytech/parity-scale-codec/pull/268))
+- `#[codec(crate = "<path>")]` top-level attribute to be used with the new `MaxEncodedLen`
+trait, which allows to specify a different path to the crate that contains the `MaxEncodedLen` trait.
+Useful when using renamed dependencies, e.g. importing `parity_scale_codec` as `codec` for brevity.
```
bkchr, line 15:
```suggestion
Useful when using generating a type through a macro and this type should implement `MaxEncodedLen` and the final crate doesn't have `parity-scale-codec` as dependency.
```

bkchr: Renamed dependencies are already handled.

PR #277, opened 2021-06-24
Improve `#[codec(crate = path)]` attribute: path, not string literal

coriolinus:
Accepting a path instead of a string literal makes life easier for
the downstream macros which need to actually use that attribute.

### DIFF
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -134,31 +134,46 @@ fn crate_access() -> syn::Result<Ident> {
 	}
 }
 
-/// Match `#[codec(crate = ...)]` and return the `...`
-fn codec_crate_path_lit(attr: &Attribute) -> Option<Lit> {
-	// match `#[codec ...]`
-	if !attr.path.is_ident("codec") {
-		return None;
-	};
-	// match `#[codec(crate = ...)]` and return the `...`
-	match attr.parse_meta() {
-		Ok(Meta::NameValue(MetaNameValue { path, lit, .. })) if path.is_ident("crate") => {
-			Some(lit)
-		}
-		_ => None,
-	}
+/// This struct matches `crate = ...` where the ellipsis is a `Path`.
+struct CratePath {
+	_crate_token: Token![crate],
+	_eq_token: Token![=],
+	path: Path,
 }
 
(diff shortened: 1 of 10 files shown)


derive/src/utils.rs
```diff
@@ -134,31 +134,46 @@ fn crate_access() -> syn::Result<Ident> {
 	}
 }
 
-/// Match `#[codec(crate = ...)]` and return the `...`
-fn codec_crate_path_lit(attr: &Attribute) -> Option<Lit> {
+/// This struct matches `crate = ...` where the ellipsis is a `Path`.
```
Xanewok, line 137: If we're introducing a helper struct implementing `Parse`, would it be possible to abstract parsing the entire attribute `#[codec(crate = ...)]` rather than a part of it: `crate = ...`?

it seems like it's going to be used once for this specific attribute and if we're introducing more abstractions, it'd be worthwhile to abstract parsing the whole top-level attribute while we're at it.

Xanewok (approved):
Looks good! Left a note wrt new `CratePath` struct.

bkchr (approved):
Did you tried that this works?

I have somewhere in my brain something about problems with paths in attributes, but maybe I mix something up.

coriolinus:
> Did you tried that this works?

I made [a test](https://github.com/paritytech/parity-scale-codec/blob/2be8e7b3ab2b44cf6461438f1017ae28c608e8e7/tests/max_encoded_len_ui/pass/codec_crate.rs). If the test suite passes (with `derive` feature enabled), then this works.

I think that what you're thinking of is that you can't do paths in this style _if_ you use `parse_attr`, which is why I ended up introducing the `CratePath` struct instead.

derive/src/utils.rs
coriolinus: Probably possible, but this is actually the recommended style: match on the path, then put more complicated stuff into the parentheses. See the [`Attribute::parse_args` docs](https://docs.rs/syn/1.0.73/syn/struct.Attribute.html#method.parse_args).

Xanewok: Okay then, thanks for the link :slightly_smiling_face:

PR #278, opened 2021-06-24
Bump version to 2.2.0-rc.2

Xanewok:
So that we can release a version including https://github.com/paritytech/parity-scale-codec/pull/277.


coriolinus (approved):

coriolinus:
Oops, sorry for neglecting to put that into #277 directly.

coriolinus:
It may be necessary to make this into rc.3; I already tagged rc.2, and I'm not sure what will happen (if anything) if the `Cargo.toml` versions don't match the git tag versions.

Xanewok:
It didn't go out yet and Cargo does no validation of tags whatsoever, so I'd say we're free to force-push the tag instead.

PR #282, opened 2021-07-02
Bump version to 2.2.0 and hide `MaxEncodedLen` behind an additional crate feature

Xanewok:
Implemented the scheme outlined in https://github.com/paritytech/parity-common/pull/552#issuecomment-872252537.

This avoids the required otherwise cascade bumps across https://github.com/paritytech/parity-common which is something that I didn't consider at first due to the requirement of explicitly selecting pre-releases. Thanks to that, other dependencies can keep on depending on `^2` or `^2.1` versions of the codec, while the only crate outside of the Substrate workspace, which is `primitive-types -> impl-codec`, can just opt-in to `max-encoded-len` and directly implement the new trait. Since the trait definition is somewhat stable and FWIU only related new attributes require some more polish, this will probably not cause any breakages.

Tested locally https://github.com/paritytech/substrate/pull/9163 and it's looking good so far.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -281,6 +281,7 @@ mod depth_limit;
 mod encode_append;
 mod encode_like;
 mod error;
+#[cfg(feature = "max-encoded-len")]
 mod max_encoded_len;
 
 pub use self::error::Error;
@@ -297,6 +298,7 @@ pub use self::decode_all::DecodeAll;
 pub use self::depth_limit::DecodeLimit;
 pub use self::encode_append::EncodeAppend;
 pub use self::encode_like::{EncodeLike, Ref};
+#[cfg(feature = "max-encoded-len")]
 pub use max_encoded_len::MaxEncodedLen;
 /// Derive macro for [`MaxEncodedLen`][max_encoded_len::MaxEncodedLen].
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -343,5 +343,5 @@ pub use max_encoded_len::MaxEncodedLen;
 /// #[codec(crate = $crate::codec)]
 /// struct Example;
 /// ```
-#[cfg(feature = "derive")]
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -297,6 +298,7 @@ pub use self::decode_all::DecodeAll;
 pub use self::depth_limit::DecodeLimit;
 pub use self::encode_append::EncodeAppend;
 pub use self::encode_like::{EncodeLike, Ref};
+#[cfg(feature = "max-encoded-len")]
```
coriolinus, line 301: You also probably need to add this to `pub use parity_scale_codec_derive::MaxEncodedLen` below.

src/max_encoded_len.rs
```diff
@@ -15,6 +15,8 @@
 
 //! `trait MaxEncodedLen` bounds the maximum encoded length of items.
 
+#![cfg(feature = "max-encoded-len")]
```
coriolinus, line 18: This is redundant with the feature gate on `mod max_encoded_len;` in `lib.rs`; it should be fine to get rid of it.

coriolinus (approved):
Some nits, but a good idea to un-stick the process.

src/max_encoded_len.rs
Xanewok: Done in https://github.com/paritytech/parity-scale-codec/pull/282/commits/5c022a4250244afd0294b6963640184884ac3cbd

src/lib.rs
Xanewok: Done in https://github.com/paritytech/parity-scale-codec/pull/282/commits/5c022a4250244afd0294b6963640184884ac3cbd

Xanewok: Thanks for spotting it!

Xanewok:
Thanks for taking a look! I'll publish these crates now

KiChjang:
Any chance to also make changes to `CHANGELOG.md` at the repo root?

PR #286, opened 2021-08-09
Add associated method to DecodeLimit for advancing input index

KiChjang:


src/depth_limit.rs
```diff
@@ -82,6 +87,15 @@ impl<T: Decode> DecodeLimit for T {
...
+	fn decode_and_advance_with_depth_limit<I: Input>(limit: u32, input: &mut I) -> Result<Self, Error> {
+		let mut input = DepthTrackingInput {
+			input,
+			depth: 0,
+			max_depth: limit,
+		};
+		T::decode(&mut input)
+	}
+
 	fn decode_with_depth_limit(limit: u32, input: &[u8]) -> Result<Self, Error> {
```
shawntabrizi, line 99: I think we need to put a deprecated on this

bkchr: Not really. However, we should create an issue to merge both of these methods in the next major release. @KiChjang please also create some new milestone for the next major and attach the issue.

bkchr (approved):

arkpar (approved):

src/depth_limit.rs
KiChjang: Filed as #287 and added a SCALE codec 3.0 milestone.

KiChjang:
Unsure what to do about the status checks? The CI failures seem unrelated to this PR.

thiolliere:
if errors are legit then we should fix master

KiChjang:
This may need to be fast tracked to a new minor version to solve a security bug; it can't wait for 3.0.

src/bit_vec.rs
```diff
@@ -211,6 +211,10 @@ mod tests {
 	}
 
 	#[test]
+	// Flacky test due to:
+	// * https://github.com/bitvecto-rs/bitvec/issues/135
+	// * https://github.com/rust-lang/miri/issues/1866
+	#[cfg(not(miri))]
```
thiolliere, line 217: CI seems to fail due to this flacky test

thiolliere (approved):

thiolliere:
> This may need to be fast tracked to a new minor version to solve a security bug; it can't wait for 3.0.

yes we can bump to a new minor in this PR or just create a new PR to bump just after.

Also it needs an entry in the CHANGELOG

PR #289, opened 2021-09-08
Fix CI

thiolliere:
let's see if it works when we don't bump bitvec


thiolliere:
miri is also failing on bitvec 0.20.1, we probably need to look into it.
if it is serious we might want to change polkadot not to use bit-vec anymore and just get rid of this dependency.

PR #290, opened 2021-09-09
User core2 for no_std io support

olanod:
The [core2](https://crates.io/crates/core2) crate is usefull for no_std deployments as it implements the common `io` traits and uses `std::io` when `std` support is enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,6 +140,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "28b9d6de7f49e22cf97ad17fc4036ece69300032f45f78f30b4a4482cdc3f4a6"
 
 [[package]]
+name = "core2"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dbceae1ad0d3679d016526dd5cc9863156e9d394c1a9591cf7ab20bb25cab490"
+dependencies = [
+ "memchr",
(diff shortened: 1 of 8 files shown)


bkchr:
Why do you need io traits for `no_std` and why should we add this to SCALE?

olanod:
I'd say in big part for code cleanness, keeping things simple and convenient by keeping code generic down to `no_std` and `alloc` environments as well. In nice generic code It's common to receive an `io::{Read, Write}` when dealing with bytes but it in no_std and alloc means we additionally need a bunch of feature flags to provide the trimmed down specialized versions.  `Seek` and `BufRead` are also nice abstractions that come handy in no_std, in [`scales::Value`](https://github.com/valibre-org/scales/blob/main/src/value.rs#L12)(no_std) I ended up implementing a cursor like functionality that I rather have implemented as a generic `Cursor` to allow for more flexibility to users without extra feature flags.

bkchr:
You could for most of the stuff just use unit type wrappers to use core2 in your project.

I'm not very keen on using some random crate that copies code from Rust core. Where we don't know what happens in the future with it.

olanod:
Fair enough, wrapper types is part of the ugliness I wanted to avoid in the first place but I get your concern ... Feel free to close if it doesn't really add much value to people.

bkchr:
Ty nevertheless! Maybe in the future we can add it.

PR #291, opened 2021-09-12
Minor codegen hygiene improvements and modernizations

Robbepop:
This PR fixes some minor macro hygiene issues such as `::core::marker::Sized` and `::core::primitive::u8`.
Also it replaces the `extern crate $name as _parity_scale_codec` by using the imported name:

- For example `::scale` if `parity_scale_codec` has been renamed to `scale` by the dependent crate
- `crate` if the code is generated for the crate itself
- `::parity_scale_codec` for tests of the `parity_scale_codec` crate

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -38,17 +38,30 @@ mod trait_bounds;
 
 /// Include the `parity-scale-codec` crate under a known name (`_parity_scale_codec`).
 fn include_parity_scale_codec_crate() -> proc_macro2::TokenStream {
+	match parity_scale_codec_ident() {
+		Ok(ident) => quote! { extern crate #ident as _parity_scale_codec; },
+		Err(e) => e.into_compile_error(),
+	}
+}
+
+/// Returns the identifier of the `parity-scale-codec` crate as used.
+///
+/// The identifier might change if the depending crate imported it
+/// using a custom package name.
+fn parity_scale_codec_ident() -> Result<Ident, Error> {
+	static ORIGINAL_NAME: &str = "parity-scale-codec";
+	static ORIGINAL_IDENT: &str = "parity_scale_codec";
+	fn ident_from_str(name: &str) -> Ident {
+		Ident::new(name, Span::call_site())
+	}
 	// This "hack" is required for the tests.
-	if std::env::var("CARGO_PKG_NAME").unwrap() == "parity-scale-codec" {
-		quote!( extern crate parity_scale_codec as _parity_scale_codec; )
+	if std::env::var("CARGO_PKG_NAME").unwrap() == ORIGINAL_NAME {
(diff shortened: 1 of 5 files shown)


derive/src/lib.rs
```diff
@@ -36,34 +36,44 @@ mod max_encoded_len;
 mod utils;
 mod trait_bounds;
 
-/// Include the `parity-scale-codec` crate under a known name (`_parity_scale_codec`).
-fn include_parity_scale_codec_crate() -> proc_macro2::TokenStream {
+/// Returns a tokenstream to refer to the `parity-scale-codec` crate
+/// independent of it being renamed as import or it being the crate itself.
+pub(crate) fn parity_scale_codec_ident() -> TokenStream2 {
```
bkchr, line 41: This will generate really bad errors in the case that the crate isn't found.

Maybe we should do the check once and forward the ident to everywhere?

bkchr (approved):

derive/src/lib.rs
Robbepop: Good catch!

For a solution we need to work around the fact that `Ident` is not `Send` or `Sync`.
A solution using a custom `enum` capturing the important state won't be elegant but will work. Any other ideas?

thiolliere: I'm not sure I follow, is there an issue with forwarding the TokenStream ? we can clone it every time if we really want.

Robbepop: The problem is that we now have multiple calls to the new API and therefore we could have multiple expansions of this line:
https://github.com/paritytech/parity-scale-codec/pull/291/files#diff-60da9bde79deaa50c0ce6f42d1f47205ba31f448dc2713cc746339343581b699R44
Which each would produce another compile error. I think this is what Basti means and it indeed is not optimal.
Therefore we should ideally panic after we have seen the first error but panics aren't nice either.

Robbepop: We could also have just a single call and store the `TokenStream` somewhere as @thiolliere probably meant and reuse it. However, this would not fix the problem that we then still expand into the stored compile error token stream.

thiolliere: personally I would do:
* in encode_derive: check first that we can import parity_scale_codec. in case it fails then return a single error, in case it succeeds then pass it everywhere we need it.
* do same for decode_derive and other derive macro.

bkchr: I mean having it multiple times is fine. I just mean that code like this: 
https://github.com/paritytech/parity-scale-codec/pull/291/files#diff-ad3d5b95ee0d10f177d8b9e7a2046a8d1108bda4c0a04141090b18e67a573856R145

Could result in very weird compiler errors.

Robbepop: Implemented @thiolliere 's suggestion.

thiolliere (approved):

PR #288, opened 2021-08-23
Fix CI failures on master

KiChjang:
This happens because the latest stable Rust compiler has slightly changed its error message format for proc macros.

PR #293, opened 2021-09-28
Add test to assert macro hygiene for Encode and Decode derives

Robbepop:
Uses the `#![no_implicit_prelude]` attribute to add UI tests to guard `Encode` and `Decode` derives against macro hygiene mistakes.

This way the PR found a minor macro hygiene bug in `Decode` and fixes it.
Also fixed a minor bug with `MaxEncodedLen` test.

### DIFF
--- a/tests/scale_codec_ui.rs
+++ b/tests/scale_codec_ui.rs
@@ -1,0 +1,25 @@
+// Copyright (C) 2020-2021 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[test]
+#[cfg(feature = "derive")]
+fn scale_codec_ui_tests() {
+	// As trybuild is using `cargo check`, we don't need the real WASM binaries.
+	std::env::set_var("SKIP_WASM_BUILD", "1");
+
+	let t = trybuild::TestCases::new();
+	t.compile_fail("tests/scale_codec_ui/*.rs");
+	t.pass("tests/scale_codec_ui/pass/*.rs");
+}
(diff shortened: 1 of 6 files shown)


thiolliere:
miri is failing because of a false positive https://github.com/paritytech/parity-scale-codec/pull/288#issuecomment-916223198
I think you can remove the miri test

tests/scale_codec_ui.rs
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[test]
+#[cfg(feature = "derive")]
+fn scale_codec_ui_tests() {
+	// As trybuild is using `cargo check`, we don't need the real WASM binaries.
+	std::env::set_var("SKIP_WASM_BUILD", "1");
+
```
bkchr, line 21:
```suggestion
```

bkchr: We don't build any wasm files in here.

bkchr (approved):

Robbepop:
@bkchr applied your suggestion. I just copied the code and comment from the other UI test.

@thiolliere what do you mean by remove `miri` test? It is failing in another crate so I think it is okay to merge the PR, right?

Could we do a minor release after this? It would be nice to have the latest changes in dependencies such as ink! that also has tests for `no-implicit-prelude` but fails due to `parity-scale-codec`.

thiolliere:
> @thiolliere what do you mean by remove miri test? It is failing in another crate so I think it is okay to merge the PR, right?

the CI is failing, I think if we keep a ci test failing for a false positive we will never look at it again and we will miss in case another error is raised.

By removing the miri test I mean this:
```diff
diff --git a/src/bit_vec.rs b/src/bit_vec.rs
index 1beca32..35d3123 100644
--- a/src/bit_vec.rs
+++ b/src/bit_vec.rs
@@ -191,6 +191,10 @@ mod tests {
        }
 
        #[test]
+       // Flaky test due to:
+       // * https://github.com/bitvecto-rs/bitvec/issues/135
+       // * https://github.com/rust-lang/miri/issues/1866
+       #[cfg(not(miri))]
        fn bitvec_u32() {
                for v in &test_data!(u32) {
                        let encoded = v.encode();

Robbepop:
The `miri` test is removed now. Ready to merge. :)

thiolliere:
I don't mind another release, we can do another PR which update the changelog and version of parity_scale_codec.
Actually we can do a patch IMO 2.3.1

Robbepop:
> Actually we can do a patch IMO 2.3.1

:+1: 

> I don't mind another release, we can do another PR which update the changelog and version of parity_scale_codec.

Are you doing that or shall I?

thiolliere:
I'm not doing currently, you can do

PR #297, opened 2021-10-29
Bump bitvec dependency to 0.22.3

ed255:
There was a conflict between bitvec and a dependency of it (funty) that
caused a build failure. See:
- https://github.com/bitvecto-rs/bitvec/issues/105
- https://github.com/myrrlyn/funty/issues/3

Due to semver, when importing parity-scale-codec as a library and
resolving bitvec 0.20.1 dependencies, funty 0.12 is pulled, which causes
the aforementioned build failure.  I believe this is not happening when
testing parity-scale-codec itself because Cargo.lock pins funty to
1.1.0, but when importing parity-scale-codec from another crate, this
pinning is sometimes not possible.  Bumping bitvec to 0.22.3 solves this
issue, as it is compatible with funty 0.12.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,9 +51,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitvec"
-version = "0.20.4"
+version = "0.22.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7774144344a4faa177370406a7ff5f1da24303817368584c6206c8303eb07848"
+checksum = "5237f00a8c86130a0cc317830e558b966dd7850d48a953d998c813f01a41b527"
 dependencies = [
  "funty",
  "radium",
@@ -252,9 +252,9 @@ dependencies = [
 
 [[package]]
 name = "funty"
-version = "1.1.0"
+version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fed34cd105917e91daa4da6b3728c47b068749d6a62c59811f06ed2ac71d9da7"
+checksum = "1847abb9cb65d566acd5942e94aea9c8f547ad02c98e1649326fc0e8910b8b1e"
 
 [[package]]
 name = "generic-array"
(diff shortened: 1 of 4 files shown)


thiolliere:
if bitvec 0.20 which depends on funty version 1 cannot be compile with funty 1.2, then funty is breaking semver.
If the breaking is actually kind of minor, then they should release bitvec 0.20.5 to support funty 1.2, but instead they decided to make bitvec 0.20 depends on a pinned version of funty: 1.1

I already proposed the **correct** fix so that bitvec 0.20 is compatible with funty 1.2 https://github.com/bitvecto-rs/bitvec/pull/110
If you want this to happen then you should push for it in their repo.

I don't think it is a good idea to bump bitvec to 0.22 as this is a major breaking change and will force use to release parity-scale-codec version 3. And maybe some month later we will have 0.22 which only depends on funty 1.2 and incompatible with funty 1.3 and same story again.

If you want to be able to use funty 1.2 then try to release bitvec 0.20.5 with this fix: https://github.com/bitvecto-rs/bitvec/pull/110
Or try to yank funty 1.2 because it is a breaking change. But I don't think we can do anything on our side here.

ed255:
> if bitvec 0.20 which depends on funty version 1 cannot be compile with funty 1.2, then funty is breaking semver. If the breaking is actually kind of minor, then they should release bitvec 0.20.5 to support funty 1.2, but instead they decided to make bitvec 0.20 depends on a pinned version of funty: 1.1
> 
> I already proposed the **correct** fix so that bitvec 0.20 is compatible with funty 1.2 [bitvecto-rs/bitvec#110](https://github.com/bitvecto-rs/bitvec/pull/110) If you want this to happen then you should push for it in their repo.
> 
> I don't think it is a good idea to bump bitvec to 0.22 as this is a major breaking change and will force use to release parity-scale-codec version 3. And maybe some month later we will have 0.22 which only depends on funty 1.2 and incompatible with funty 1.3 and same story again.
> 
> If you want to be able to use funty 1.2 then try to release bitvec 0.20.5 with this fix: [bitvecto-rs/bitvec#110](https://github.com/bitvecto-rs/bitvec/pull/110) Or try to yank funty 1.2 because it is a breaking change. But I don't think we can do anything on our side here.

Thank you so much for the detailed reply!  And thanks for pointing me to the right solution for this issue.

PR #301, opened 2021-11-23
Add some doc for the derive macro

thiolliere:


derive/src/lib.rs
```diff
@@ -360,7 +367,19 @@ pub fn compact_as_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStr
...
+///
+/// By default the macro will try to bound the types needed to implement `MaxEncodedLen`, but the
+/// bounds can be specified manually with the top level attribute:
+/// ```
+/// # use parity_scale_codec_derive::Encode;
+/// # use parity_scale_codec::MaxEncodedLen;
+/// # #[derive(Encode, MaxEncodedLen)]
+/// #[codec(mel_bound(T: MaxEncodedLen))]
+/// # struct MyWrapper<T>(T);
+/// ```
```
thiolliere, line 382: yes it may be overkill to set up a compiling example just to check the syntax of the attribute, but it renders fine I think

kianenigma (approved):

derive/src/lib.rs
bkchr: Compiled examples are never an overkill 👍

bkchr (approved):

derive/src/lib.rs
```diff
@@ -75,6 +75,15 @@ fn wrap_with_dummy_const(input: DeriveInput, impl_block: proc_macro2::TokenStrea
...
 /// Derive `parity_scale_codec::Encode` and `parity_scale_codec::EncodeLike` for struct and enum.
 ///
+/// # Top level attributes
+///
+/// By default the macro will try to bound the types needed to implement their trait, but the
+/// bounds can be specified manually with the top level attributes:
+/// * `#[codec(encode_bound(T: Encode))]`: a custom where bound that will be used when deriving the
+///   `Encode` trait.
+/// * `#[codec(decode_bound(T: Decode))]`: a custom where bound that will be used when deriving the
+///   `Decode` trait.
```
dvdplm, line 85:
```suggestion
/// By default the macro will add [`Encode`] and [`Decode`] bounds to all types, but the
/// bounds can be specified manually with the top level attributes:
/// * `#[codec(encode_bound(T: Encode))]`: a custom bound added to the `where`-clause when deriving the
///   `Encode` trait, overriding the default.
/// * `#[codec(decode_bound(T: Decode))]`: a custom bound added to the `where`-clause when deriving the
///   `Decode` trait, overriding the default.
```

maybe?

thiolliere: yes thanks, in both case what we mean by `all types` is a bit vague but I think it is fine.

thiolliere:
bot merge

bkchr:
:robot: done

PR #304, opened 2021-11-26
Support `Usize`

NoCtrlZ:
Hi @bkchr 
I am implementing `pallet` and realized that the `usize` are not supported so I added `usize`.
I would appreciate you could confirm 😎
Thank you!

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -39,6 +39,7 @@ use core::num::{
 	NonZeroU32,
 	NonZeroU64,
 	NonZeroU128,
+	NonZeroUsize
 };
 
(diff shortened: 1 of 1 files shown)


bkchr:
`usize` is not platform independent, aka we have not implemented this on purpose and it will stay this way.

NoCtrlZ:
@bkchr 
Thank you for the review!
I have a question.
How can I implement `Decode` to `usize`?

bkchr:
You can not. Use an `u32` or `u64`.

NoCtrlZ:
Thank you for the answer!
Copy that.

PR #308, opened 2021-12-29
Fix test compilation for features

### LINKED ISSUE #307: derive `Encode`/`Decode` for `String` with `no_std`
Why is `String` encoding/decoding not supported with `no_std`? If there is a strong reason for this, can you give me some guidance how to best work around this without having to maintain a fork of this crate

mversic:
* fix tests compilation for features

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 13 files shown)


mversic:
I believe that compiler output has changed for the two failing compile tests. Should I update the expected output files so that the tests pass?

PR #314, opened 2022-02-02
BREAKING CHANGE: DecodeLimit and DecodeAll extensions now advance input

### LINKED ISSUE #287: Merge `decode_with_depth_limit` with `decode_and_advance_with_depth_limit`
#286 added a `decode_and_advance_with_depth_limit` associated method for the `DepthLimit` trait to advance the input pointer while decoding an item. We would like to merge this with `decode_with_depth_limit` once we release a new major version.

### LINKED ISSUE #313: `decode_all` should consume the input
`decode_all` should consume the `input` to be more consistent with the normal `Decode` interface. By consuming I mean `&mut &[u8]`, so that we advance the input as we decode it.

wigy-opensource-developer:
Fixes #287
Fixes #313

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,1 @@
+hard_tabs = true
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,1 +1,0 @@
-hard_tabs = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,8 @@ rust-version = "1.56.1"
 [dependencies]
 arrayvec = { version = "0.7", default-features = false }
 serde = { version = "1.0.102", optional = true }
-parity-scale-codec-derive = { path = "derive", version = "2.3.1", default-features = false, optional = true }
-bitvec = { version = "1", default-features = false, features = ["alloc"], optional = true }
+parity-scale-codec-derive = { path = "derive", version = "3.0.0", default-features = false, optional = true }
+bitvec = { version = "1", default-features = false, features = [ "alloc" ], optional = true }
 bytes = { version = "1", default-features = false, optional = true }
 byte-slice-cast = { version = "1.0.0", default-features = false }
 generic-array = { version = "0.14.4", optional = true }
@@ -59,7 +59,4 @@ chain-error = []
 full = []
 
 [workspace]
-members = [
-	"derive",
-	"fuzzer",
-]
+members = ["derive", "fuzzer"]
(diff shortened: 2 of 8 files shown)


KiChjang (dismissed):
Whitespace!!!! We use tabs instead of spaces here

wigy-opensource-developer:
> Whitespace!!!! We use tabs instead of spaces here

Oh. Checking my IDE and `cargo fmt` settings... Maybe it is different in Substrate and Polkadot.

src/decode_all.rs
```diff
@@ -23,15 +23,14 @@ pub trait DecodeAll: Sized {
 	/// Decode `Self` and consume all of the given input data.
 	///
 	/// If not all data is consumed, an error is returned.
-	fn decode_all(input: &[u8]) -> Result<Self, Error>;
+	fn decode_all<I: Input>(input: &mut I) -> Result<Self, Error>;
```
bkchr, line 26: Please just keep `[u8]`, because `Input` doesn't give you any guarantee that it can return a `remaining_len`

bkchr: Or we document this. But then we also need to return the error of `remaining_len`. If remaining_len would return `None` we would need to try to read on byte to ensure that there isn't anything anymore.

bkchr:
> > Whitespace!!!! We use tabs instead of spaces here
> 
> Oh. Checking my IDE and `cargo fmt` settings... Maybe it is different in Substrate and Polkadot.

You probably need to copy the rustfmt.toml file. You can add it here in the pr.

wigy-opensource-developer:
This has been a rollercoaster, but hopefully everything is better than how it started.

bkchr (approved):

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "parity-scale-codec"
 description = "SCALE - Simple Concatenating Aggregated Little Endians"
-version = "2.3.1"
+version = "3.0.0"
```
bkchr, line 4: We should also bump the proc-macro.

.rustfmt.toml
```diff
@@ -0,0 +1 @@
+hard_tabs = true
```
bkchr, line 1: Please just copy it from Substrate

wigy-opensource-developer:
bot merge

wigy-opensource-developer:
@KiChjang Could you please have a second look on this? I would like to merge to master and release 3.0.0

KiChjang (approved):

PR #317, opened 2022-02-09
Provides implementations for String, Rc, Arc

rise0chen:


bkchr:
As you already see these implementations exists and you can use them. We have put them behind features because of reasons and they will stay there.

rise0chen:
Why discourage the usage of `String`?

It is used in `sp_version::RuntimeVersion`.

rise0chen:
`String` needn't use  `std`.
Can we use an extra feature, like `alloc` ?

bkchr:
There is the `full` feature that enables all the stuff. You even removed the code for that here.

rise0chen:
yes, thinks.
I misunderstand it, I thought `full` will enable all features.

PR #326, opened 2022-03-01
[ci] Change image to ci-linux:production with clang13

alvicsam:
Pipeline with current image doesn't work

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -105,6 +105,7 @@ miri:
     RUST_BACKTRACE:                1
     MIRIFLAGS:                     "-Zmiri-disable-isolation"
   script:
+    - cargo +nightly install miri rust-src llvm-tools-preview grcov rust-covfix xargo cargo-contract
     - time cargo +nightly miri test --features bit-vec,bytes,generic-array,arbitrary --release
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -19,7 +19,8 @@ workflow:
     - if: $CI_COMMIT_BRANCH
 
 .docker-env:                       &docker-env
-  image:                           paritytech/ink-ci-linux:production
+  # Temporary image, remove when clang in "ink-ci-linux:production" is updated to 13
```
bkchr, line 22: Is that still true? Don't you actually just created this image for the future?

alvicsam: It depends on how quickly clang in ci image for `ink` would be updated. 
If it takes too much time I'll recreate this image with `production` tag and setup it's updates. So currently it is just a quick fix

radupopa2010 (approved):

.gitlab-ci.yml
bkchr: Do you keep track of this?

alvicsam: I'll create the issue

PR #339, opened 2022-05-20
Add dependabot

sergejparity:
This PR adds `dependabot` to repository. To automate dependency updates and security vulnerabilities fixes.  
Related issue https://github.com/paritytech/ci_cd/issues/114


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+
+updates:
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: daily
+  - package-ecosystem: cargo
+    directory: '/'
+    schedule:
+      interval: 'daily'
```
bkchr, line 11: For both you should set a maximum number of concurrent prs.

sergejparity: Default value is 5. How many would be ok from your perspective?

bkchr: Ahh okay, I didn't know there is a default value. 5 sounds good :)

radupopa2010 (approved):

PR #375, opened 2022-09-13
Don't include bitvec with std  feature unless asked for explicitly

jsdw:
Should not be a breaking code change since bitvec things are all controlled by the "bit-vec" feature, but should avoid pulling in the dep unless it's asked for.

(Part of trying to remove bitvec as a mandatory dependency for consumers (see https://github.com/paritytech/scale-info/pull/168))

Minimum Rust version from this is increased to 1.60.0.

What's our versioning policy w.r.t minimum supported rust version being increased?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ license = "Apache-2.0"
 repository = "https://github.com/paritytech/parity-scale-codec"
 categories = ["encoding"]
 edition = "2021"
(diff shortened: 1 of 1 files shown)


ascjones (approved):

bkchr:
> What's our versioning policy w.r.t minimum supported rust version being increased?

I'm not aware of any. We could probably do a minor version bump.

PR #389, opened 2022-11-23
[ci] add cargo-nono

alvicsam:
close https://github.com/paritytech/ci_cd/issues/681

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -116,14 +116,14 @@ miri:
   script:
     - time cargo +nightly miri test --features bit-vec,bytes,generic-array,arbitrary --release
 
-nono:
+build-no-std:
   stage:                           test
   <<:                              *docker-env
   variables:
     RUST_BACKTRACE:                1
   script:
-    - cargo install cargo-nono
-    - time cargo nono check --no-default-features
+    - rustup target add thumbv6m-none-eabi
+    - time cargo build --target thumbv6m-none-eabi --no-default-features
 
 #### stage:                        build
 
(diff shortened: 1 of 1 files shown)


alvicsam:
I've added `cargo-nono` but it fails. Should it have any additional options?

jsdw:
I believe you need to add something like `--no-default-features` so that "std" isn't enabled by default.

oleg-plakida (approved):

alvicsam:
Actually I didn't. @gilescope told me about the tool so I assumed that he already tested it

jsdw:
Sadly, I tried adding this to `src/lib.rs`:

```rust
use std::collections::HashMap;
pub fn has_map<K, V>(_map: HashMap<K, V>) {
	let _m: HashMap<u32, u32> = HashMap::new();
	todo!()
}
```

I'd have expected that to cause `cargo nono check --no-default-features` to fail. I wonder whether it only checks that the packages used are marked "no-std" rather than specifically checking for compat.

I got this advice from https://blog.dbrgn.ch/2019/12/24/testing-for-no-std-compatibility/.

One of the other things it suggested is running:

```
rustup target add thumbv6m-none-eabi
cargo build --target thumbv6m-none-eabi --no-default-features
```

If I do that, it compiles OK without my addition adbove and fails to compile if I try importing `std::collections::HashMap`.

I'm not sure if there is value in both that _and_ `cargo nono` but offhand that seems like the way to go to me!

jsdw (approved):
LGTM; let's see how it pans out :)

PR #403, opened 2023-01-23
Export `decode_vec_with_len`

mrcnski:
This is necessary to fix the custom `impl Decode` for `BoundedVec`. Currently when decoding, it will [allocate before doing a length check](https://github.com/paritytech/substrate/blob/2dff067e9f7f6f3cc4dbfdaaa97753eccc407689/primitives/core/src/bounded/bounded_vec.rs#L311-L315). See:

https://github.com/paritytech/polkadot/pull/6603#discussion_r1083461147

> You want to ensure that you don't even start try decoding/allocating when the length of the vector is more than the
> allowed maximum.

> You first decode length of the vector and then early reject if that is too long.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,19 @@ All notable changes to this crate are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this crate adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.3.0]
+
+This release exports `decode_vec_with_len` to support custom decoding of `Vec`s.
+
+### Added
+
+- Export `decode_vec_with_len`.
+
 ## [3.2.1] - 2022-09-14
 
 This release fixes compilation on no-std envs.
 
-## Changed
+### Changed
 
  - Use core RangeInclusive instead of std [#378](https://github.com/paritytech/parity-scale-codec/pull/378)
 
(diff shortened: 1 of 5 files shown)


bkchr (approved):

bkchr:
@mrcnski you also need to bump the patch version and then release this version!

mrcnski:
@bkchr I'll need some help with that. Should the version bump be a separate PR? How do I release it? (I don't think I have access rights, or do I?)

bkchr:
You can bump the patch version in this pr. Then merge it. Then do cargo publish. Then go on github and create some new release. I can create the github release for you if you don't know how to.

mrcnski:
Okay! Technically though, shouldn't this be a minor version bump as we are adding to the public API?

bkchr:
> Technically though, shouldn't this be a minor version bump as we are adding to the public API?

Why not? We don't break anything? I mean we can also bump the minor version :shrug:

ggwpez (approved):

mrcnski:
@bkchr The major version is > 0 so a minor version bump doesn't denote a breaking change, it just means that functionality was added.

bkchr:
> @bkchr The major version is > 0 so a minor version bump doesn't denote a breaking change, it just means that functionality was added.

I know.

I read again semver:
```
MINOR version when you add functionality in a backwards compatible manner
PATCH version when you make backwards compatible bug fixes
```

Let's bump the minor. I general before I meant that you can do whatever you want :D I would not have blocked you on bumping the minor :D

KiChjang (approved):

mrcnski:
Cool, I was able to `cargo publish` -- didn't know I could do that. Raised a new release as well!

PR #408, opened 2023-02-08
Remove `full` feature

athei:
Guarding things behind features only makes sense to reduce compile time or prevent pulling in situational dependencies.

It is not a good way to guard against accidental implementations on unwanted types. Due to feature unification it can be enabled through oopsies anyways. The feature just makes testing more complicated.

Hence we remove the `full` feature. We leave it in the manifest but it doesn't do anything (for backwards compat). But we also need to put `alloc::syn` behind conditional compilation as it is not available on all targets.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -42,15 +42,6 @@ default:
 
 #### stage:                        check
 
-check-rust-stable-no_derive_no_std_full:
-  stage:                           check
-  <<:                              *docker-env
-  variables:
-    RUSTFLAGS:                     "-Cdebug-assertions=y -Dwarnings"
-  script:
-    - time cargo +stable check --verbose --no-default-features --features bit-vec,bytes,generic-array,full
-
-
 check-rust-stable-no_derive_no_std:
   stage:                           check
   <<:                              *docker-env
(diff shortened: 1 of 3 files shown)


bkchr (approved):
I think in the future we should remove these features and just enable these types directly.

mrcnski:
Would be good to build this feature in CI, i.e. `string` versions of `check-rust-stable-no_derive_full` and `check-rust-stable-no_derive_no_std_full`. Otherwise we may not catch accidental build errors right away.

athei:
> I think in the future we should remove these features and just enable these types directly.

I think this is a good idea. Features are not a good way to protect anyone from pulling in anything dangerous as they can always activate by accident. I can do it in this PR if you want. I would of course leave the features in the manifest so we don't have a breaking change. They would just don't do anything until we remove them with the next major release.

This will also solve the "CI issue"  where we don't check any feature combination as brought up by @mrcnski.

bkchr:
Yeah, then let's do it in this pr.

athei:
> Yeah, then let's do it in this pr.

Done

bkchr (approved):

Issue #425, opened 2023-04-18
Stack overflow issue in scalecodec while decoding inputs


mmostafas:
There is an stack overflow issue in the [parity-scale-codec][2] crate that crashes while decoding inputs. We ran into this issue when decoding an XCM call but we believe it can be triggered from other code paths as well.

Here is a tests that triggers this bug:

```rust
#[test]
fn decode_instruction_segfault() {
	use parity_scale_codec::DecodeLimit;
	let mut decoded = VersionedXcm::V3(v3::Xcm::<()>::new());
	let encoded = hex_literal::hex!("03f800000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515151515");

	decoded  = DecodeLimit::decode_with_depth_limit(20, &mut encoded.as_slice()).unwrap();
}
```

Please note that the depth limit of `20` is used here which is relatively low.

There is also a similar issue submitted [here][2].

[2]: https://github.com/paritytech/parity-scale-codec/issues/419

koute:
@mmostafas Could you check whether this PR fixes your issue?

https://github.com/paritytech/parity-scale-codec/pull/426

mmostafas:
Thank you @koute,

Unfortunately #426 did not fix our issues. If you want to reproduce the bug, you can do so by adding the provided test above to the unit tests for the [XCM][1]. 

Please let me know if you need more info.

[1]: https://github.com/paritytech/polkadot/blob/master/xcm/src/tests.rs

koute:
@mmostafas Thanks for the explanation on how to reproduce it!

I confirm I can reproduce the issue. I'll fix it.

koute:
I've found why this happens; it's due to this upstream rustc bug:

https://github.com/rust-lang/rust/issues/34283

The `Instruction` enum is 1472 bytes big and has a ton of variants. When it is decoded there's a `match` over every variant, and for every variant it seems that extra stack space is consumed proportional to how big the enum is. Then nest it a few times and you get a stack overflow.

koute:
Okay, I think I fixed it this time!

[...]

Issue #419, opened 2023-03-26
Stack overflow when decoding large boxed arrays


nazar-pc:
Example:
```rust
#![feature(new_uninit)]
use parity_scale_codec::{Decode, Encode};

#[derive(Encode, Decode)]
struct S(Box<[u8; 100 * 1024 * 1024]>);

impl Default for S {
    fn default() -> Self {
        Self(unsafe { Box::new_zeroed().assume_init() })
    }
}

fn main() {
    let s = S::default();
    let encoded = s.encode();
    println!("Encoded successfully");
    S::decode(&mut encoded.as_slice()).unwrap();
}
```

Output:
```
Encoded successfully

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
fish: Job 1, 'cargo run --example overflow' terminated by signal SIGABRT (Abort)
```

Likely due to https://github.com/rust-lang/rust/issues/53827

nazar-pc:
This manual implementation works as a workaround, but it is slow:
```rust
#![feature(new_uninit)]

use parity_scale_codec::{Decode, Encode, Input};
use std::mem::ManuallyDrop;

#[derive(Encode)]
struct S(Box<[u8; 100 * 1024 * 1024]>);

impl Decode for S {
    fn decode<I: Input>(input: &mut I) -> Result<Self, parity_scale_codec::Error> {
        let piece = parity_scale_codec::decode_vec_with_len::<Vec<u8>, _>(input, 100 * 1024 * 1024)
            .map_err(|error| error.chain("Could not decode `S.0`"))?;
        let mut piece = ManuallyDrop::new(piece);
        // SAFETY: Original memory is not dropped and guaranteed to be allocated
        let piece = unsafe { Box::from_raw(piece.as_mut_ptr() as *mut [u8; 100 * 1024 * 1024]) };
        Ok(S(piece))
    }
}

impl Default for S {
    fn default() -> Self {
        Self(unsafe { Box::new_zeroed().assume_init() })
    }
}

fn main() {
    let s = S::default();
    let encoded = s.encode();
    println!("Encoded successfully");
    S::decode(&mut encoded.as_slice()).unwrap();
}
```

ggwpez:
Minimized example on stable rust (in case anyone doubts the `unsafe`):  
```rust
use parity_scale_codec::Decode;

fn main() {
    let data = &[];
    let _ = Box::<[u8; 100 * 1024 * 1024]>::decode(&mut data.as_slice());
[...]

PR #426, opened 2023-04-20
Rework decoding of `Box`es, `Rc`s, `Arc`s, arrays and enums (stack overflow fix)

koute:
This PR reworks how we decode `Box`es etc. and arrays; in particular it fixes three issues:

- Trying to deserialize a `Box` which contains a big array doesn't overflow the stack anymore. This is fixed by first preallocating an empty `Box` and then directly deserializing into it, instead of first deserializing the value on the stack and moving it into the `Box` after it's deserialized.
- Trying to deserialize big nested enums doesn't overflow the stack anymore. AFAIK this is due to an upstream bug in rustc (https://github.com/rust-lang/rust/issues/34283) and one workaround which seems to work is to just wrap the body of each `match` with a lambda and immediately call it and return the value. I've also added a test for this issue.
- Elements of partially read arrays are now properly dropped if the whole array can't be decoded.

I've also updated `Cargo.lock` and updated to `criterion` 0.4 while I was at it.

Fixes https://github.com/paritytech/parity-scale-codec/issues/419

Fixes https://github.com/paritytech/parity-scale-codec/issues/425


src/codec.rs
```diff
@@ -488,29 +550,125 @@ impl<T, X> Encode for X where
...
+		// Placement new is not yet stable, but we can just manually allocate a chunk of memory
+		// and convert it to a `Box` ourselves.
+		//
+		// The explicit types here are written out for clarity.
+		//
+		// TODO: Use `Box::new_uninit` once that's stable.
+		let layout = core::alloc::Layout::from_size_align(
+			core::mem::size_of::<MaybeUninit<T>>(),
+			core::mem::align_of::<MaybeUninit<T>>()
+		).expect("layout is always valid");
```
gilescope, line 580: Can we just do this?
```suggestion
        let layout = Layout::new::<MaybeUninit<T>>();
```
Source: https://doc.rust-lang.org/src/core/alloc/layout.rs.html#148-154

tests/mod.rs
```diff
@@ -644,3 +644,94 @@ fn no_warning_for_deprecated() {
 		VariantB,
 	}
 }
+
```
ggwpez, line 647:
```suggestion
#[test]
fn decoding_array_does_not_stack_overflow() {
	let data = &[];
	let _ = <[u8; 100 * 1024 * 1024]>::decode(&mut data.as_slice());
}
```
This one still fails…

ggwpez: Can we check the maximal allocateable stack size in a non-hacky way and bail early?  
Not sure what really to do either. Just wanted to point out that this could still panic…

koute: > Can we check the maximal allocateable stack size in a non-hacky way and bail early?

Hm... using Linux-specific APIs - possibly. We could grab the stack base, grab the stack pointer, and grab the maximum stack size, and see how much's left.

Not sure if we actually want to do that; seems a little bit excessive. (And it won't work inside e.g. WASM)

> Do you know how do other decoding libs handle this?

AFAIK, they don't. (:

ggwpez: I looked at `serde` and they only support arrays up to 32 entries.  

Anyway, can you maybe add a "known issues" or something so that we can link it if people report it as bug again in the future? And maybe a "should_panic" test to verify that arrays are not covered.

src/codec.rs
```diff
@@ -488,29 +550,122 @@ impl<T, X> Encode for X where
...
+			input.ascend_ref();
+			result
+		}
 
+		// No specialization exists, so just decode `T` and then `into()` it.
+		//
+		// This must be done through an `#[inline(never)]` function to prevent
+		// the compiler from eagerly preallocating the stack space for the return
+		// value even if `decode_wrapped` is implemented.
+		decode_inner(input)
```
michalkucharczyk, line 667: This could go into `WrapperTypeDecode::decode_wrapped` as the default/blanket implementation. The trick with `decode_inner` probably would not be needed then.

`decode_wrapped` could simply return `Result` instead of `Option`.

src/codec.rs
```diff
@@ -488,29 +550,122 @@ impl<T, X> Encode for X where
...
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-		input.descend_ref()?;
-		let result = Ok(T::decode(input)?.into());
-		input.ascend_ref();
-		result
-	}
+		// First try to see if a specialized implementation exists.
+		if let Some(result) = Self::decode_wrapped(input) {
+			return result;
+		}
```
michalkucharczyk, line 648: Maybe, we could simply call: `Self::decode_wrapped(input)`, and provide the `default` implementation with the code below?

src/codec.rs
```diff
@@ -488,29 +550,122 @@ impl<T, X> Encode for X where
 pub trait WrapperTypeDecode: Sized {
```
michalkucharczyk, line 550: Is it still marker trait?

michalkucharczyk (approved):

tests/mod.rs
koute: I can add a note somewhere, but as far as a test is concerned that can't be done with `should_panic` as a stack overflow doesn't actually produce a panic. (We could maybe test it by spawning a new process, but I'm not convinced it's actually worth it.)

src/codec.rs
koute: That's a good point; thanks!

src/codec.rs
koute: From the user's perspective, yes; from the implementation point of view, no. Practically it doesn't make much difference, I think.

koute:
Pushed a few more changes:
- The stack overflow issue from https://github.com/paritytech/parity-scale-codec/issues/425 is now fixed. AFAIK this is due to an upstream bug in rustc (https://github.com/rust-lang/rust/issues/34283) and one workaround which seems to work is to just wrap the body of each `match` with a lambda and immediately call it and return the value. I've also added a test for this issue.
- I noticed the README has bitrotted (the code snippets in it didn't compile anymore, etc.), so I deleted it and replaced it with the crate-level doc comment, and then I replaced the doc comment with the README using the `#![doc = include_str!("../README.md")]` trick. So now we'll have a guarantee that the README won't bitrot in the future.
- I've added the note regarding deserializing arrays as requested by @ggwpez 

Should be good to go pending reviews.

tests/mod.rs
ggwpez: Yes that sounds good, thanks!

README.md
```diff
@@ -7,70 +7,81 @@ SCALE is a light-weight format which allows encoding (and decoding) which makes
...
+* `fn decode<I: Input>(value: &mut I) -> Result<Self, Error>`: Tries to decode the value from
+  SCALE format to the type it is called on.
 Returns an `Err` if the decoding fails.
 
 ### CompactAs
 
-The `CompactAs` trait is used for wrapping custom types/structs as compact types, which makes them even more space/memory efficient.
-The compact encoding is described [here](https://docs.substrate.io/reference/scale-codec/#fnref-1).
+The `CompactAs` trait is used for wrapping custom types/structs as compact types, which makes
+them even more space/memory efficient. The compact encoding is described [here](https://docs.substrate.io/v3/advanced/scale-codec/#compactgeneral-integers).
```
kpp, line 55: Your link does not work for me for some reason but `https://docs.substrate.io/reference/scale-codec/#fn-1` does.

README.md
```diff
@@ -7,70 +7,81 @@ SCALE is a light-weight format which allows encoding (and decoding) which makes
...
 
 ### Decode
 
-The `Decode` trait is used for deserialization/decoding of encoded data into the respective types.
+The `Decode` trait is used for deserialization/decoding of encoded data into the respective
+types.
 
-* `fn decode<I: Input>(value: &mut I) -> Result<Self, Error>`: Tries to decode the value from SCALE format to the type it is called on.
+* `fn decode<I: Input>(value: &mut I) -> Result<Self, Error>`: Tries to decode the value from
+  SCALE format to the type it is called on.
```
kpp, line 49: Double spaces

README.md
```diff
@@ -7,70 +7,81 @@ SCALE is a light-weight format which allows encoding (and decoding) which makes
...
-It is important to note that the encoding context (knowledge of how the types and data structures look)
-needs to be known separately at both encoding and decoding ends.
+It is important to note that the encoding context (knowledge of how the types and data
+structures look) needs to be known separately at both encoding and decoding ends.
 The encoded data does not include this contextual information.
 
 To get a better understanding of how the encoding is done for different types,
 take a look at the
-[low-level data formats overview page at the Substrate docs site](https://docs.substrate.io/reference/scale-codec/).
+[SCALE Code page at the Substrate Knowledge Base](https://docs.substrate.io/v3/advanced/scale-codec/).
```
kpp, line 16: Link `https://docs.substrate.io/v3/advanced/scale-codec/` does not work for me

Cargo.toml
```diff
@@ -21,9 +21,9 @@ arbitrary = { version = "1.0.1", features = ["derive"], optional = true }
 impl-trait-for-tuples = "0.2.2"
 
 [dev-dependencies]
-criterion = "0.3.0"
+criterion = "0.4.0"
 serde_derive = { version = "1.0" }
-parity-scale-codec-derive = { path = "derive", default-features = false }
+parity-scale-codec-derive = { version = ">= 3.5.0", path = "derive", default-features = false }
```
kpp, line 26: Do you need to specify version req?

kpp (changes requested):

Cargo.toml
koute: Technically I don't, but I like to guarantee that the new derive crate is used if the main crate's bumped. Especially in cases like this where the bug that I'm fixing could in theory have security implications and the changes in the derive crate are necessary to fix it. We don't want someone to only bump the main crate and accidentally forget about bumping the derive crate and then have a surprise pikachu moment. (For [my own serialization crate I just always force a specific version of the derive crate](https://github.com/koute/speedy/blob/master/Cargo.toml#L17) so there's never no funny business going on.)

(Besides, we don't really test mismatched versions of those two crates, so in theory there could be other obscure bugs popping up if someone uses them like that.)

README.md
koute: Ah yeah, sorry. I just blindly copy-pasted this from the doc comment assuming it's correct. I'll clean it up.

README.md
koute: Hm, that's actually what we want to indicate that the text's part of the list's bullet point. :P I'll reformat it.

koute:
I've resolved conflicts, bumped version from 3.5.0 to 3.6.0 (since 3.5.0 was released in the meantime) and cleaned up the README.

I've also cleaned up our feature flags semantics:
* The `max-encoded-len` now doesn't automatically enable the derive macro reexports nor pull in the `parity-scale-codec-derive` dependency.
* Enabling the `parity-scale-codec-derive` without enabling the `derive` feature doesn't reexport the macros anymore; `derive` is now necessary.

These two issues are why @ggwpez made https://github.com/paritytech/parity-scale-codec/pull/429; now with these changes that PR should be not necessary anymore. (: (The issue was that the derive macros were accidentally imported *twice*, not that the derive macro and the trait were imported with the same name - that compiles just fine, so importing the derive macros and the traits under different names is not necessary.)

These are *technically* breaking changes, but `max-encoded-len` is explicitly documented an unstable, and I'd argue the derive macro reexports working without the `derive` feature enabled was a bug (everyone should have been enabling the `derive` feature), so I didn't bump the major version.

Should be good to go.

ggwpez (commented):
>These two issues are why @ggwpez made https://github.com/paritytech/parity-scale-codec/pull/429; now with these changes that PR should be not necessary anymore. (: (The issue was that the derive macros were accidentally imported twice, not that the derive macro and the trait were imported with the same name - that compiles just fine, so importing the derive macros and the traits under different names is not necessary.)

Could you point to the line of how this is fixed now?  
AFAIK it was importing `scale::Decode` (macro + trait) and `scale_derive::Decode` (macro).  
So yes, the macro was imported twice, but without the `derive` feature the macro was not available for import.

koute:
> Could you point to the line of how this is fixed now?

* The `use parity_scale_codec_derive::{Encode, Decode};` was feature gated on whether the `derive` feature is enabled.
* The reexport of `Encode` and `Decode` was was feature gated on whether the `parity-scale-codec-derive` feature is enabled.
* Enabling the `max-encoded-len` feature also implicitly enabled `parity-scale-codec-derive`.

So basically the `use parity_scale_codec_derive::{Encode, Decode};` was enabled because it thought that the derive reexports were not enabled (because the `derive` feature was not enabled), but the reexports weren't actually controlled by the `derive` feature but by the `parity-scale-codec-derive` feature, and that was transitively enabled by the `max-encoded-len` feature, so we ended up with the derive macro imported twice: once explicitly with `use` and once through the `parity-scale-codec` reexport (where it overlaps with the traits, so importing the trait also automatically imports the macro).

derive/src/decode.rs
```diff
@@ -72,7 +72,12 @@ pub fn quote(
 
 				quote_spanned! { v.span() =>
 					__codec_x_edqy if __codec_x_edqy == #index as ::core::primitive::u8 => {
-						#create
+						// NOTE: This lambda is necessary to work around an upstream bug
+						// where each extra branch results in excessive stack usage:
+						//   https://github.com/rust-lang/rust/issues/34283
+						return (move || {
```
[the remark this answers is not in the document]
koute: I haven't checked, but I highly doubt it will. And we're not serializing/deserializing that much anyway (if we would I would have worked on optimizing `parity-scale-codec` more, as it is relatively slow, e.g. [compare it to my `speedy` here](https://github.com/djkoloski/rust_serialization_benchmark#log))

If you want maybe we could run the benchmarks once I'll make the PR to pull it into substrate?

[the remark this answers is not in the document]
ggwpez: I run the criterion before and after and there is no regression, so should be good.

davxy (approved):

src/codec.rs
```diff
@@ -286,6 +284,52 @@ pub trait DecodeLength {
...
 }
 
+/// A decoding context.
+///
+/// Used in [`Decode::decode_into`] to allow the implementation to explicitly
+/// assert that the `MaybeUninit` passed into that function was properly initialized.
+pub struct DecodeContext(PhantomData<*const ()>);
+
+/// A marker type signifying that the decoding finished.
+pub struct DecodeFinished(PhantomData<*const ()>);
```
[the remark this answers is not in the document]
skunert: Also as discussed this could be moved somewhere else to avoid being able to just instantiate `DecodeFinished` in this file.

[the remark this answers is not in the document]
koute: 1. The `PhantomData<*const ()>` makes the type `!Send` and `!Sync` because of the `*const ()` inside, which makes it harder to accidentally misuse. (Not strictly necessary, but nice to have.)
2. A type like `struct Foo` is publicly constructible anywhere, so e.g. if someone would implement the `Decode::decode_into` *outside* of this crate they'd be able to construct `DecodeFinished` without calling `assert_decoding_finished`.

src/codec.rs
```diff
@@ -488,29 +550,99 @@ impl<T, X> Encode for X where
...
+		Ok(boxed)
+	}
 }
+
 impl<T> WrapperTypeDecode for Rc<T> {
 	type Wrapped = T;
+
+	fn decode_wrapped<I: Input>(input: &mut I) -> Result<Self, Error> where Self::Wrapped: Decode {
+		// TODO: This is inefficient; use `Rc::new_uninit` once that's stable.
+		Box::<T>::decode(input).map(|output| output.into())
```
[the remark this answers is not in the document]
koute: > This and the one for `Arc` are using the `Box::decode` just to not rewrite the whole pointer logic right?

Mostly yes. 

I just didn't do it for simplicity's sake. We don't decode into these types very often anyway. I was planning to fix this TODO once all of the necessary relevant functions are stabilized so that we can further simplify the whole thing.

skunert (approved):

michalkucharczyk (approved):

src/codec.rs
```diff
@@ -885,9 +933,118 @@ impl<T: Encode, const N: usize> Encode for [T; N] {
...
+				}
+			}
+		}
+
+		let mut state = State {
+			count: 0,
+			slice
+		};
+
+		while state.count < state.slice.len() {
```
kpp, line 1036: How about `for` loop to enable possible vectorization?

kpp (approved):

src/codec.rs
```diff
@@ -286,6 +284,52 @@ pub trait DecodeLength {
 	fn len(self_encoded: &[u8]) -> Result<usize, Error>;
 }
 
+/// A decoding context.
+///
+/// Used in [`Decode::decode_into`] to allow the implementation to explicitly
+/// assert that the `MaybeUninit` passed into that function was properly initialized.
+pub struct DecodeContext(PhantomData<*const ()>);
```
bkchr, line 291: Is there more "value" in this than described in the docs? If no, can we not just add a `unsafe fn new` to `DecodeFinished`?

src/codec.rs
```diff
@@ -885,9 +933,118 @@ impl<T: Encode, const N: usize> Encode for [T; N] {
...
+			TypeInfo::Unknown => false
+		};
+
+		if is_primitive {
+			// Let's read the array in bulk as that's going to be a lot
+			// faster than just reading each element one-by-one.
+
+			let ptr: *mut [T; N] = dst.as_mut_ptr();
+			let ptr: *mut u8 = ptr.cast();
+			let bytesize = core::mem::size_of::<T>().checked_mul(N).expect("array lengths are sane");
```
bkchr, line 972:
```
struct AssertNotOverflow<T, const N: usize>(std::marker::PhantomData<T>);

impl<T, const N: usize> AssertNotOverflow<T, N> {
    const OK: () = assert!(std::mem::size_of::<T>().checked_mul(N).is_some(), "Array size would overflow memory size");
}

fn test<T, const N: usize>() {
    let () = AssertNotOverflow::<T, N>::OK;
}
```

If we add this here, we should get this error at compile time :D

src/codec.rs
```diff
@@ -296,6 +340,24 @@ pub trait Decode: Sized {
 	/// Attempt to deserialise the value from input.
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error>;
 
+	/// Attempt to deserialise the value from input into a preallocated piece of memory.
```
bkchr, line 343:
```suggestion
	/// Attempt to deserialize the value from input into a pre-allocated piece of memory.
```

bkchr (approved):
Nice implementation, I like :D 

Left some minor comments on things that should be improved. Especially the story around `DecodeContext` and `decode_into` is not really well explained. Currrently this isn't callable by someone outside of the crate, which is fine, but should be left some reason why.

It would also be nice to create some issue for the `assume_init` stuff and the best would be to directly link to the Rust issues.

src/codec.rs
kpp: state.slice[state.count] Inserts a bound check which can be eliminated with for and the cycle can be optimized for primitives like u32. Just a guess :D

koute: AFAIK the compiler should be able to elide the bounds check here as-is. Whether the loop is written with a `for` or with a `while` doesn't matter.

Nevertheless, it's unlikely that there will an opportunity for vectorization here though. Everything inside the loop would have to be inlined, the decoded `T` would have to be `Copy`, and reading from the `input` would also need to have its bounds elided.

One possible extra optimization that could be done here would be to calculate the minimum amount of bytes that has to be read to decode one `T`, multiply that by the number of elements, and then add a new method to `Input` which would use that size to insert a precheck (if possible for a given `Input`) that would allow the compiler to elide the bounds when reading from `Input`. But that's out of scope for this PR.

Cargo.toml
```diff
@@ -21,9 +21,9 @@ arbitrary = { version = "1.0.1", features = ["derive"], optional = true }
 impl-trait-for-tuples = "0.2.2"
 
 [dev-dependencies]
-criterion = "0.3.0"
+criterion = "0.4.0"
 serde_derive = { version = "1.0" }
-parity-scale-codec-derive = { path = "derive", default-features = false }
+parity-scale-codec-derive = { version = ">= 3.6.0", path = "derive", default-features = false }
```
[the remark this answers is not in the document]
koute: Ah, good catch. This was a brainfart on my part!

This should have been on the non-dev dependency (to make sure a compatible `derive` crate is always used, if someone updates `parity-scale-codec` without updating `parity-scale-codec-derive`)

src/codec.rs
koute: Hmm.... the idea was that it's only possible to materialize a `DecodeFinished` inside of the `decode_into` (because that's the only place where `DecodeContext` is available) but perhaps you're right. Since the function to create a `DecodeFinished` is `unsafe` anyway the extra `DecodeContext` doesn't really bring too much benefit.

I'll simplify it.

src/codec.rs
```diff
@@ -885,9 +933,118 @@ impl<T: Encode, const N: usize> Encode for [T; N] {
...
+
+		// SAFETY: `decode_into` succeeded, so the array is initialized.
+		unsafe {
+			Ok(array.assume_init())
+		}
+	}
+
+    fn decode_into<I: Input>(ctx: DecodeContext, input: &mut I, dst: &mut MaybeUninit<Self>) -> Result<DecodeFinished, Error> {
+		let is_primitive = match <T as Decode>::TYPE_INFO {
+			| TypeInfo::U8
```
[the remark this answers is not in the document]
koute: This trick cannot be used for booleans because they have a niche. That is, not every bit pattern for booleans is valid: only `0b00000001` and `0b00000000` are, so we can't just read them in bulk.

(Well, we *can*, but we'd need to add an extra loop that would verify that every `byte & !1 == 0`, and with SIMD such a loop would most likely be pretty cheap. Probably not worth the effort, and out of scope of this PR. :P)

src/codec.rs
```diff
@@ -885,9 +933,118 @@ impl<T: Encode, const N: usize> Encode for [T; N] {
...
+
+		if is_primitive {
+			// Let's read the array in bulk as that's going to be a lot
+			// faster than just reading each element one-by-one.
+
+			let ptr: *mut [T; N] = dst.as_mut_ptr();
+			let ptr: *mut u8 = ptr.cast();
+			let bytesize = core::mem::size_of::<T>().checked_mul(N).expect("array lengths are sane");
+
+			// TODO: This is potentially slow; it'd be better if `Input` supported
```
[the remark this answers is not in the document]
koute: Hm, sorry, I'm not sure I understand what you're suggesting here. What exactly would this change? (: They'd have to be zeroed anyway as long as `Input` accepts a `&[u8]` instead of `&mut [MaybeUninit<u8>]`.

src/codec.rs
```diff
@@ -885,9 +933,118 @@ impl<T: Encode, const N: usize> Encode for [T; N] {
...
+					// make a difference in debug mode.
+					return;
+				}
+
+				for item in &mut self.slice[..self.count] {
+					// SAFETY: Each time we've read a new element we incremented `count`,
+					//         and we only drop at most `count` elements here,
+					//         so all of the elements we drop here are valid.
+					unsafe {
+						item.assume_init_drop();
```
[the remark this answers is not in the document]
koute: That's not stable, but I'll add a TODO comment.

src/codec.rs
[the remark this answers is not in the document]
kpp: > They'd have to be zeroed anyway

Yes, but they will be zeroed by compiler and there will be no need for an unsafe code to memzero them.

The previous implementation used that trick for i8/u8, see `let mut array: [u8; N] = [0; N];`. Still this comment is not a blocker, you can ignore it/implement in another PR.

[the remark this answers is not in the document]
koute: Okay, I see what you mean now!

Hmm... it which case it'd probably be better to just make it possible to read into the array without initializing it. (:

src/codec.rs
```diff
@@ -296,6 +295,24 @@ pub trait Decode: Sized {
...
+	/// Attempt to deserialize the value from input into a pre-allocated piece of memory.
+	///
+	/// The default implementation will just call [`Decode::decode`].
+	///
+	/// # Safety
+	///
+	/// If this function returns `Ok` then `dst` **must** be properly initialized.
+	///
+	/// This is enforced by requiring the implementation to return a [`DecodeFinished`]
+	/// which can only be created by calling [`DecodeFinished::assert_decoding_finished`] which is `unsafe`.
```
bkchr, line 307: Docs still need to be updated

src/codec.rs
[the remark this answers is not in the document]
koute: > Could we not just put the code for the `Box` implementation into some macro and reuse it here? see_no_evil

That'd be tricky because without `::new_uninit` we have no good way to preallocate the uninitialized memory for the `Rc`/`Arc`.

In case of a `Box<T>` the only thing that the memory holds is a `T`, so we can get its layout and manually allocate the memory.

In case of `Rc`/`Arc` there are also the refcounts allocated in the same chunk of memory, and there's AFAIK there's no good way to get a `Layout` that could be used to manually preallocate that memory and pass it to `Rc::from_raw`. One way we could do it would be to just hardcode it (this is basically equivalent `Box<(usize, usize, T)>`), and then verify in a build script that it's correct (in case it suddenly changes, since it's an internal implementation detail), and if it is then do it the fast way, and if it isn't then emit a compile-time warning and do it the slow way. So for all practical purposes that'd work and be safe, but obviously it is very hacky and I don't really like it. And I'm not convinced it's worth it. (As opposed to just waiting waiting until the `new_uninit` methods are stabilized.)

src/codec.rs
```diff
@@ -296,6 +340,24 @@ pub trait Decode: Sized {
...
+	///
+	/// The default implementation will just call [`Decode::decode`].
+	///
+	/// # Safety
+	///
+	/// If this function returns `Ok` then `dst` **must** be properly initialized.
+	///
+	/// This is enforced by requiring the implementation to return a [`DecodeFinished`]
+	/// which can only be created by calling [`DecodeContext::assert_decoding_finished`] which is `unsafe`.
+	fn decode_into<I: Input>(ctx: DecodeContext, input: &mut I, dst: &mut MaybeUninit<Self>) -> Result<DecodeFinished, Error> {
```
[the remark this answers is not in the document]
koute: > Ahh, you may want people that they can implement it, but not call it themselves?

It wasn't entirely a deliberate "let's block it", but with no immediate use case in mind I just didn't make it publicly constructible.

Anyhow, it's a moot point now though as I nuked the `DecodeContext` type anyway. :P

PR #429, opened 2023-05-03
Make tests build

ggwpez:
Tests do not build with only MEL feature as pointed out in https://github.com/paritytech/parity-scale-codec/pull/428#pullrequestreview-1407066751

This standard pattern that is used multiple times in the code does not actually work:  
```rust
#[cfg(not(feature = "derive"))]
use parity_scale_codec_derive::Decode;
use parity_scale_codec::Decode;
```
It errors with a redefinition of `Decode` when the feature is disabled. I aliased them in the tests now to `DeriveDecode`.  
It is a bit unfortunate that `parity_scale_codec::Decode` is both a trait and a derive macro, depending on the features.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,7 +91,8 @@
 //!
 //! ### Simple types
 //!
-//! ```
+#![cfg_attr(feature = "derive", doc = "```rust")]
+#![cfg_attr(not(feature = "derive"), doc = "*(Only compiled with feature `derive`)*\n```ignore")]
 //! # use parity_scale_codec_derive::{Encode as DeriveEncode, Decode as DeriveDecode};
 //!
 //! use parity_scale_codec::{Encode, Decode};
@@ -140,7 +141,8 @@
 //!
 //! ### Compact type with HasCompact
 //!
-//! ```
+#![cfg_attr(feature = "derive", doc = "```rust")]
+#![cfg_attr(not(feature = "derive"), doc = "*(Only compiled with feature `derive`)*\n```ignore")]
 //! # use parity_scale_codec_derive::{Encode as DeriveEncode, Decode as DeriveDecode};
 //!
 //! use parity_scale_codec::{Encode, Decode, Compact, HasCompact};
@@ -167,7 +169,8 @@
 //! ```
 //! ### Type with CompactAs
 //!
-//! ```rust
+#![cfg_attr(feature = "derive", doc = "```rust")]
+#![cfg_attr(not(feature = "derive"), doc = "*(Only compiled with feature `derive`)*\n```ignore")]
 //! # use parity_scale_codec_derive::{Encode as DeriveEncode, Decode as DeriveDecode};
 //! use serde_derive::{Serialize, Deserialize};
 //! use parity_scale_codec::{Encode, Decode, Compact, HasCompact, CompactAs, Error};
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -143,19 +141,17 @@
...
 //! ```
-//! # // Import macros if derive feature is not used.
-//! # #[cfg(not(feature="derive"))]
-//! # use parity_scale_codec_derive::{Encode, Decode};
+//! # use parity_scale_codec_derive::{Encode as DeriveEncode, Decode as DeriveDecode};
 //!
 //! use parity_scale_codec::{Encode, Decode, Compact, HasCompact};
 //!
-//! #[derive(Debug, PartialEq, Encode, Decode)]
+//! #[derive(Debug, PartialEq, DeriveEncode, DeriveDecode)]
```
bkchr, line 148: I'm not happy with this. People will be confused on where `DeriveEncode` is coming from. 

https://users.rust-lang.org/t/how-to-document-optional-features-in-api-docs/64577/3 above the code example with `feature = "derive"` would be really nice!

tests/mod.rs
```diff
@@ -15,24 +15,23 @@
 use parity_scale_codec::{
 	Compact, CompactAs, Decode, Encode, EncodeAsRef, Error, HasCompact, Output,
 };
-#[cfg(not(feature = "derive"))]
-use parity_scale_codec_derive::{Decode, Encode};
+use parity_scale_codec_derive::{Decode as DeriveDecode, Encode as DeriveEncode};
 use serde_derive::{Deserialize, Serialize};
 
-#[derive(Debug, PartialEq, Encode, Decode)]
+#[derive(Debug, PartialEq, DeriveEncode, DeriveDecode)]
```
bkchr, line 21: For tests it is fine!

src/lib.rs
ggwpez: This auto-generated feature warning seems to only work on whole modules or types. Not on single doc lines, so i added:  
```rust
#![cfg_attr(feature = "derive", doc = "```rust")]
#![cfg_attr(not(feature = "derive"), doc = "*(Only compiled with feature `derive`)*\n```ignore")]
```
To each example to not compile them without the feature.

src/lib.rs
```diff
@@ -91,11 +91,8 @@
 //!
 //! ### Simple types
 //!
-//! ```
-//! # // Import macros if derive feature is not used.
-//! # #[cfg(not(feature="derive"))]
-//! # use parity_scale_codec_derive::{Encode, Decode};
-//!
+#![cfg_attr(feature = "derive", doc = "```rust")]
+#![cfg_attr(not(feature = "derive"), doc = "*(Only compiled with feature `derive`)*\n```ignore")]
```
mrcnski, line 95: Is the part in `*(...)*` a comment? Haven't seen this syntax before.

mrcnski (approved):
> It errors with a redefinition of Decode when the feature is disabled.

Instead of the macro import could you have renamed the trait import, or even brought it in without a name, e.g. `use Decode as _`? But this is fine too.

src/lib.rs
ggwpez: Just italic with `()`, renders as such:

image

mrcnski: Oh right, had a brainfart. 🙈 Thanks for the responses, looks good!

bkchr (approved):

Issue #454, opened 2023-06-16
parity-scale-codec-derive 3.6.0 generates a lot of clippy warnings


nazar-pc:
Upgraded from 3.1.4 to 3.6.0 and getting a lot of warnings like these in generated code:
```
warning: unneeded `return` statement
   --> crates/subspace-archiving/src/archiver.rs:170:5
    |
170 | /     /// Segment header of the parent
171 | |     #[codec(index = 4)]
172 | |     ParentSegmentHeader(SegmentHeader),
    | |______________________________________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
    = help: remove `return`
```
```
warning: try not to call a closure in the expression where it is declared
  --> crates/subspace-networking/src/request_handlers/peer_info.rs:11:5
   |
11 | /     /// Special status for starting peer. Receiving it in the running mode means an error.
12 | |     Unknown,
   | |___________^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure_call
   = note: `#[warn(clippy::redundant_closure_call)]` on by default
help: try doing something like
   |
11 ~     /// Special status for starting peer. Receiving it in the running mode means an error.
12 ~     Unknown,
   |
```

I think they need to be fixed or suppressed for generated code.

ggwpez:
Yea… same is happening in Polkadot https://github.com/paritytech/polkadot/pull/7382

ggwpez:
@koute the closure wraning be this one maybe:  
https://github.com/paritytech/parity-scale-codec/blob/0688c04994bdcf7ad817eddd26077625458102c4/derive/src/decode.rs#L78

koute:
Ah, yep. I'll fix it.

SkymanOne:
We had the same problem in ink!

koute:
PR with a fix here: https://github.com/paritytech/parity-scale-codec/pull/455

PR #455, opened 2023-06-16
Fix all clippy warnings and add clippy to the CI

koute:
Clippy doesn't like the theoretically redundant closure calls that I've added in my last PR, and since clippy doesn't run on the CI here I didn't notice.

While I was at it I also fixed all other clippy warnings, plus one error as it was complaining about an `unsafe` that I've now removed. I've also added clippy to the CI (hopefully it works).

Fixes https://github.com/paritytech/parity-scale-codec/issues/454

Fixes https://github.com/paritytech/parity-scale-codec/issues/456


ggwpez (commented):
Maybe we should add a Clippy config file?  
Could be the one from Substrate in `.cargo/config.toml`.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this crate are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this crate adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.6.1]
+
+### Fixed
+
+- Deriving `Decode` will not trigger `redundant_closure_call` clippy warnings anymore
```
ggwpez, line 12: And the unneeded return warning?

src/codec.rs
```diff
@@ -1093,7 +1094,7 @@ pub(crate) fn compact_encode_len_to<W: Output + ?Sized>(dest: &mut W, len: usize
 
 impl<T: Encode> Encode for [T] {
 	fn size_hint(&self) -> usize {
-		mem::size_of::<u32>() + mem::size_of::<T>() * self.len()
+		mem::size_of::<u32>() + core::mem::size_of_val(self)
```
ggwpez, line 1097: Not sure if saturating add would make sense here.

src/compact.rs
```diff
@@ -493,7 +493,7 @@ impl Decode for Compact<u32> {
 					return Err(U32_OUT_OF_RANGE.into());
 				}
 			},
-			3|_ => {	// |_. yeah, i know.
+			3 => {	// |_. yeah, i know.
```
ggwpez, line 496:
```suggestion
			3 => {
```
:sparkles:

src/encode_append.rs
```diff
@@ -80,25 +80,15 @@ impl<T: Encode> EncodeAppend for crate::alloc::collections::VecDeque<T> {
...
-	Ok((new_len, encoded_len, encoded_new_len))
-}
-
-// Item must have same encoding as encoded value in the encoded vec.
-fn append_or_new_vec_with_any_item<Item, I>(
-	mut self_encoded: Vec<u8>,
+/// Extends a SCALE-encoded vector with elements from the given `iter`.
+///
+/// `vec` must either be empty, or contain a valid SCALE-encoded `Vec<Item>` payload.
+fn extend_encoded_vec<Item, I>(
```
ggwpez, line 90: Nit: I think the work `append` is also used in the `sp-io` storage API to refer to this operation.

src/encode_append.rs
```diff
@@ -107,46 +97,41 @@ where
...
+			// The size of the length as encoded by SCALE didn't change, so we can just
+			// keep the old buffer as-is. We just need to update the length prefix.
+			Compact(new_item_count).using_encoded(|length_encoded|
+				vec[..old_item_count_encoded_bytesize].copy_from_slice(length_encoded)
+			);
+		} else {
+			// We can't update the length as the new length prefix will take up more
+			// space when encoded, so we need to move our data to make space for it.
+			let mut new_vec = Vec::with_capacity(
+				new_item_count_encoded_bytesize + new_item_count as usize * mem::size_of::<Item>()
```
ggwpez, line 123: I dont quite understand the code. Why is it using `mem::size_of` instead of some `Encode::size_hint` or something?  
Or is there some assumed relation between in-memory and encoded size?

ggwpez (commented):
Looks good, just some Qs.

koute:
> Maybe we should add a Clippy config file? Could be the one from Substrate in `.cargo/config.toml`.

Hmmm.... I think we should be fine with the default one here, as this is orders of magnitude smaller so we don't need to disable a bunch of lints like in Substrate.

ggwpez:
> Hmmm.... I think we should be fine with the default one here, as this is orders of magnitude smaller so we don't need to disable a bunch of lints like in Substrate.

Its not just disabling, it is also denying some. But only `correctness` and `complexity` as it seems.  
Up to you :shrug:

src/encode_append.rs
koute: In `std` it's called `extend`, so that's why I called it like that. (:

But I don't particularly care. Do you want me to rename it?

ggwpez (approved):

src/encode_append.rs
koute: Ah, you're right! Good catch! It's quite late here so I didn't notice that and just copied what the code was originally doing here, but this *is* incorrect.

In fact, now this is harmless since it's only used for `with_capacity` (as I removed the `unsafe`, but previously with the `unsafe` it looks like we had a potential out-of-bounds write...  yikes.

src/encode_append.rs
koute: Okay, I renamed it to be consistent with the method name in the trait where it's used.

CHANGELOG.md
koute: Yes, no warnings should appear now. (Or at least they don't appear now when I tested it running `cargo clippy` in `polkadot` :P)

src/encode_append.rs
koute: I've simplified this. Not sure if there's a better way of doing this. (We can't really get the exact encoded bytesize of the new items since we haven't iterated over them yet.)

src/codec.rs
koute: Hmm... perhaps, but then it's just a size hint so shouldn't really matter if it overflows.

ggwpez: I dont know if the runtime is normally compiled with overflow checks enabled (that would be my only concern).

bkchr: Overflow checks are not enabled for release builds as they would be too slow.

src/encode_append.rs
```diff
@@ -107,46 +97,39 @@ where
...
+		if old_item_count_encoded_bytesize == new_item_count_encoded_bytesize {
+			// The size of the length as encoded by SCALE didn't change, so we can just
+			// keep the old buffer as-is. We just need to update the length prefix.
+			Compact(new_item_count).using_encoded(|length_encoded|
+				vec[..old_item_count_encoded_bytesize].copy_from_slice(length_encoded)
+			);
+		} else {
+			// We can't update the length as the new length prefix will take up more
+			// space when encoded, so we need to move our data to make space for it.
+			let mut new_vec = Vec::with_capacity(vec.len() * 2);
```
bkchr, line 122: Here we should use `checked_mul`

bkchr (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this crate are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this crate adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.6.1]
```
davxy, line 8: Should we add the date?

davxy (approved):

CHANGELOG.md
koute: Yes, I'll do that before I'll merge it (so that the date will actually correspond to the day it'll be released :P)

koute:
Added a fix + test for https://github.com/paritytech/parity-scale-codec/issues/456 and also fixed clippy lints in the derive crate (this was as far as I can see necessary to test the fix for #456). Should be good to go once the CI passes.

.gitlab-ci.yml
```diff
@@ -48,6 +48,8 @@ check-clippy:
   script:
     - rustup component add clippy --toolchain stable-x86_64-unknown-linux-gnu
     - time cargo +stable clippy --locked -- -Dwarnings
+    - time cargo +stable clippy --locked -p parity-scale-codec-derive -- -Dwarnings
```
ggwpez, line 51: Can you not use something like `--workspace --tests`?

koute: That would run clippy on all of the integration tests, which at this point are not clippy clean, and at this point I didn't feel like also cleaning up all of the tests.

tests/clippy.rs
```diff
@@ -0,0 +1,26 @@
...
+
+//! This file is checked by clippy to make sure that the code generated by the derive macro
+//! doesn't spew out warnings/errors in users' code.
+
+use parity_scale_codec_derive::{Decode, Encode};
+
+#[repr(u8)]
+#[derive(Decode, Encode)]
+pub enum CLike {
+	Foo = 0,
```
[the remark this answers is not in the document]
koute: Yes, good idea. I'll create an issue.

PR #462, opened 2023-06-28
Fix stack overflow when decoding big array newtypes wrapped in a `Box`

koute:
Followup of https://github.com/paritytech/parity-scale-codec/pull/426

This PR fixes decoding of big array newtypes wrapped in a `Box`, e.g. previously only this was correctly deserialized:

```rust
#[derive(Decode)]
struct Foobar(Box<[1024 * 1024 * 1024; u8]>);
```

And now this will also be correctly handled:

```rust
#[repr(transparent)]
struct Newtype([1024 * 1024 * 1024; u8]);

#[derive(Decode)]
struct Foobar(Box<Newtype>);
```

(The newtype *must* be marked with `#[repr(transparent)]`; without that attribute it will still crash.)

Fixes https://github.com/paritytech/parity-scale-codec/issues/419

### DIFF
--- a/derive/src/decode.rs
+++ b/derive/src/decode.rs
@@ -160,7 +160,7 @@ pub fn quote_decode_into(
 	let mut decode_fields = Vec::new();
 	let mut sizes = Vec::new();
 	for field in fields {
-	let field_type = &field.ty;
+		let field_type = &field.ty;
 		decode_fields.push(quote! {{
 			let dst_: &mut ::core::mem::MaybeUninit<Self> = dst_; // To make sure the type is what we expect.
 
--- a/derive/src/decode.rs
+++ b/derive/src/decode.rs
@@ -113,6 +113,84 @@ pub fn quote(
 	}
 }
 
+pub fn quote_decode_into(
+	data: &Data,
+	crate_path: &syn::Path,
+	input: &TokenStream,
+	attrs: &[syn::Attribute]
+) -> Option<TokenStream> {
+	// Make sure the type is `#[repr(transparent)]`, as this guarantees that
+	// there can be only one field that is not zero-sized.
+	if !crate::utils::is_transparent(attrs) {
+		return None;
+	}
+
+	let fields = match data {
+		Data::Struct(
+			syn::DataStruct {
+				fields: Fields::Named(syn::FieldsNamed { named: fields, .. }) |
+				        Fields::Unnamed(syn::FieldsUnnamed { unnamed: fields, .. }),
+				..
+			}
+		) => {
+			fields
+		},
+		_ => return None
+	};
+
+	if fields.is_empty() {
+		return None;
+	}
+
+	// Bail if there are any extra attributes which could influence how the type is decoded.
+	if fields.iter().any(|field|
+		utils::get_encoded_as_type(field).is_some() ||
+		utils::is_compact(field) ||
+		utils::should_skip(&field.attrs)
+	) {
+		return None;
+	}
+
+	// Go through each field and call `decode_into` on it.
+	//
+	// Normally if there's more than one field in the struct this would be incorrect,
+	// however since the struct's marked as `#[repr(transparent)]` we're guaranteed that
+	// there's at most one non zero-sized field, so only one of these `decode_into` calls
+	// should actually do something, and the rest should just be dummy calls that do nothing.
+	let mut decode_fields = Vec::new();
+	let mut sizes = Vec::new();
+	for field in fields {
+	let field_type = &field.ty;
+		decode_fields.push(quote! {{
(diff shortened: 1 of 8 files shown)


derive/src/decode.rs
```diff
@@ -113,6 +113,84 @@ pub fn quote(
...
+	// Go through each field and call `decode_into` on it.
+	//
+	// Normally if there's more than one field in the struct this would be incorrect,
+	// however since the struct's marked as `#[repr(transparent)]` we're guaranteed that
+	// there's at most one non zero-sized field, so only one of these `decode_into` calls
+	// should actually do something, and the rest should just be dummy calls that do nothing.
+	let mut decode_fields = Vec::new();
+	let mut sizes = Vec::new();
+	for field in fields {
+	let field_type = &field.ty;
```
bkchr, line 163: FMT? :D

bkchr (approved):

derive/src/decode.rs
koute: Ah, oops. (:

Yeah, rustfmt is not a thing in this repo; we should apply it eventually.

derive/src/decode.rs
```diff
@@ -113,6 +113,84 @@ pub fn quote(
...
+
+		if !sizes.is_empty() {
+			sizes.push(quote! { + });
+		}
+		sizes.push(quote! { ::core::mem::size_of::<#field_type>() });
+	}
+
+	Some(quote!{
+		// Just a sanity check. This should always be true and be optimized-out.
+		assert_eq!(::core::mem::size_of::<Self>(), #(#sizes)*);
```
michalkucharczyk, line 185: I am not sure if this is really needed here. Would you please explain why you've put it here?

tests/mod.rs
```diff
@@ -662,6 +662,42 @@ fn decoding_a_huge_array_inside_of_arc_does_not_overflow_the_stack() {
...
+		fn decode<I: parity_scale_codec::Input>(input: &mut I) -> Result<Self, Error> {
+			let mut buffer = [0; 1];
+			input.read(&mut buffer).unwrap();
+			Ok(Self)
+		}
+	}
+
+	let data = &[1, 2, 3];
+	assert_eq!(NewtypeWithZst::decode(&mut data.as_slice()).unwrap().1, [2]);
+}
```
michalkucharczyk, line 699: Maybe a more obvious test with Zst not consuming input bytes would also good here.

michalkucharczyk (approved):

derive/src/decode.rs
```diff
@@ -113,6 +113,84 @@ pub fn quote(
...
+	// Go through each field and call `decode_into` on it.
+	//
+	// Normally if there's more than one field in the struct this would be incorrect,
+	// however since the struct's marked as `#[repr(transparent)]` we're guaranteed that
+	// there's at most one non zero-sized field, so only one of these `decode_into` calls
+	// should actually do something, and the rest should just be dummy calls that do nothing.
+	let mut decode_fields = Vec::new();
+	let mut sizes = Vec::new();
+	for field in fields {
+		let field_type = &field.ty;
```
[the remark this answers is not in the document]
koute: It's a valid use case to have e.g. a `PhantomData` in your newtype to hold a generic `T`, so I didn't really want to limit this.

davxy (approved):

derive/src/decode.rs
koute: Yes, this is not needed per-se; it's just a defense-in-depth sanity check to make sure the assumptions hold. And because I really like sprinkling assertions around for things that should always hold true. (:

(Normally I'd do this kind of an assert as a static assert, but this being a derive macro makes it very tricky due to generics.)

koute: Actually, this reminds me - we also need one more assert here to be comprehensive (an assert which checks that there's only one non ZST field); let me add that in real quick.

And to further explain why it's nice to have these: the `#[repr(transparent)]` should pretty much guarantee that this is correct, but if for some reason we'd have a bug where the check for `#[repr(transparent)]` doesn't work then the asserts would trigger and the whole thing would fail without being unsound (which we must avoid at all costs).

This is not just a theoretical scenario; this has *actually* happened in my own serialization crate! When I upgraded to syn 2.x the new version of syn had changed the way how it exposes attributes in a backwards-incompatible way, and my check for `#[repr(transparent)]` started to silently fail (well, not really because I had tests, but you do get the point).

koute:
Okay, added an extra assert to also make sure the number of non-ZST fields is correct, and added more tests. (Now the tests are guaranteed to fail if the check for `#[repr(transparent)]` commented out.)

ggwpez (approved):

PR #509, opened 2023-09-03
Use unique encoded generic parameter names in derive

bkchr:
This ensures that we don't clash with the generic parameter names of the user.


koute (approved):
Would be nice to have the `rustfmt` changes in a separate commit as it's kinda hard to review otherwise. :P (Yeah, we should finally run `rustfmt` on the whole thing and add a CI job to enforce it.)

bkchr:
Yeah sorry, I thought we already had this configured :see_no_evil:

PR #508, opened 2023-09-01
Fix max_encoded_len for Compact fields

pgherveou:
Noticed that when using compact fields the max_encoded_len macro expand to the same as if we were not using the compact attribute. 

Screenshot 2023-09-01 at 21 44 55

```rust
#[derive(Encode, Decode, MaxEncodedLen)]
pub struct Weight {
    #[codec(compact)]
    ref_time: u64,
    #[codec(compact)]
    proof_size: u64,
}
```
Will expand to 
```rust
    impl ::codec::MaxEncodedLen for Weight {
        fn max_encoded_len() -> ::core::primitive::usize {
            0_usize
                .saturating_add(<u64>::max_encoded_len())
                .saturating_add(<u64>::max_encoded_len())
        }
    }
```

This PR fix it to 
```rust
    impl ::codec::MaxEncodedLen for Weight {
        fn max_encoded_len() -> ::core::primitive::usize {
            0_usize
                .saturating_add(::parity_scale_codec::Compact<u64>::max_encoded_len())
                .saturating_add(::parity_scale_codec::Compact<u64>::max_encoded_len())
        }
    }
```

PR #512, opened 2023-09-06
Update MaxEncodedLen derive macro

pgherveou:
#508  fixed the generated code for MaxEncoded derive trait

```rust
#[codec(compact)]
foo: u64

// compile to
// old:  u64::max_encoded_len()
// new: <::codec::Compact<u64> as ::codec::MaxEncodedLen>::max_encoded_len()
```

This is all fine but cause some compilation errors when this field is generic, since Rust can not tell that `Compact<T>`  implements `MaxEncodedLen` when `T` implements it, without extra constraints

```rust
#[codec(compact)]
deposit: BalanceOf<T>,

// 53 |         pub deposit: BalanceOf<T, OldCurrency>,
//    |                      ^^^^^^^^^ the trait `parity_scale_codec::MaxEncodedLen` is not implemented for `parity_scale_codec::Compact<<OldCurrency as Currency<<T as frame_system::Config>::AccountId>>::Balance>`
```

### DIFF
--- a/derive/src/trait_bounds.rs
+++ b/derive/src/trait_bounds.rs
@@ -161,7 +161,6 @@ pub fn add<N>(
 
 		codec_types
 			.into_iter()
-			// .take(1)
 			.for_each(|ty| where_clause.predicates.push(parse_quote!(#ty : #codec_bound)));
 
 		let has_compact_bound: syn::Path = parse_quote!(#crate_path::HasCompact);
--- a/derive/src/trait_bounds.rs
+++ b/derive/src/trait_bounds.rs
@@ -161,7 +161,7 @@ pub fn add<N>(
 
 		codec_types
 			.into_iter()
-			.take(1)
+			// .take(1)
 			.for_each(|ty| where_clause.predicates.push(parse_quote!(#ty : #codec_bound)));
 
 		let has_compact_bound: syn::Path = parse_quote!(#crate_path::HasCompact);
--- a/derive/src/trait_bounds.rs
+++ b/derive/src/trait_bounds.rs
@@ -161,6 +161,7 @@ pub fn add<N>(
 
 		codec_types
 			.into_iter()
+			.take(1)
 			.for_each(|ty| where_clause.predicates.push(parse_quote!(#ty : #codec_bound)));
 
 		let has_compact_bound: syn::Path = parse_quote!(#crate_path::HasCompact);
--- a/src/compact.rs
+++ b/src/compact.rs
@@ -19,7 +19,7 @@ use arrayvec::ArrayVec;
 use crate::alloc::vec::Vec;
 use crate::codec::{Encode, Decode, Input, Output, EncodeAsRef};
 use crate::encode_like::EncodeLike;
-use crate::Error;
+use crate::{Error, MaxEncodedLen};
 #[cfg(feature = "fuzz")]
 use arbitrary::Arbitrary;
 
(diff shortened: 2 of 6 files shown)


pgherveou:
With that fix, we should be able to use 3.6.6 in substrate by simply adding this trait to 

```rust
pub trait Balance:
	AtLeast32BitUnsigned + FullCodec + Copy + Default + Debug + scale_info::TypeInfo + MaxEncodedLen
{
}
```

src/max_encoded_len.rs
```diff
@@ -34,6 +34,10 @@ pub trait MaxEncodedLen: Encode {
 	fn max_encoded_len() -> usize;
 }
 
+/// Similar to [`MaxEncodedLen`], but will return the maximum encoded length of the compact encoding
+pub trait MaxCompactEncodedLen {
+	fn max_compact_encoded_len() -> usize;
```
pgherveou, line 39: or maybe we just add it to MaxEncodedLen with a default implementation, so we don't have to introduce an extra trait 🤔

pgherveou: actually, it might be cleaner to use a `HasCompactMaxEncodedLen` like we do with `HasCompact` for encoding

src/max_encoded_len.rs
```diff
@@ -32,6 +32,10 @@ use crate::alloc::sync::Arc;
 pub trait MaxEncodedLen: Encode {
 	/// Upper bound, in bytes, of the maximum encoded size of this item.
 	fn max_encoded_len() -> usize;
+	/// Upper bound, in bytes, of the maximum encoded size of the compact encoding of this item.
+	fn max_compact_encoded_len() -> usize {
+		Self::max_encoded_len()
+	}
```
koute, line 38: Hmm... personally I'd probably prefer to keep this as a `#[no_doc]` and make it an internal implementation detail, instead of making it public.

koute (commented):
Could you add a test which would not compile without this change?

koute:
Also, a side note only tangentially related to this PR - when reviewing this code I've noticed that a compact `u16`'s maximum encoded length is apparently 4 bytes!? That is strange, because for all of the other types it only takes at most one extra byte over the size of the type itself, and IIRC the bare minimum to represent a u16 in a space-efficient varint encoding should be 3 bytes (which can actually encode a number up to 21 bits).

But apparently we kinda screwed this up, and that's how we encode `u16`s:

`assert_eq!(Compact(u16::MAX).encode().len(), 4); // Doesn't panic!`

Kinda unfortunate that we waste space here.

src/max_encoded_len.rs
pgherveou: by internal you mean simply adding the `#[doc(hidden)]` attribute or is there other things you can do to truly keep it internal?

koute: > by internal you mean simply adding the `#[doc(hidden)]` attribute or is there other things you can do to truly keep it internal?

Just hide it. I guess you could also prefix its name, e.g. prepend `scale_internal_` to the name (which I did when I was adding `scale_internal_decode_bytes` on `Input`)

bkchr (changes requested):

src/max_encoded_len.rs
```diff
@@ -46,16 +50,25 @@ macro_rules! impl_primitives {
 	};
 }
 
-impl_primitives!(u8, u16, u32, u64, u128, i8, i16, i32, i64, i128, bool);
-
 impl_primitives!(
+	i8, i16, i32, i64, i128, bool,
```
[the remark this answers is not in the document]
koute: For the unsigned types the `impl_compact` macro is used now, so they're not skipped. (:

tests/max_encoded_len.rs
```diff
@@ -65,18 +65,17 @@ fn generic_max_length() {
...
 #[test]
 fn compact_field_max_length() {
-	assert_eq!(
-		CompactField::max_encoded_len(),
-		Compact::<u64>::max_encoded_len() + u64::max_encoded_len()
-	);
+	let max_len = CompactField::<u64>::max_encoded_len();
+	assert_eq!(max_len, Compact::<u64>::max_encoded_len() + u64::max_encoded_len());
+	assert_eq!(max_len, 17);
 }
```
koute, line 79: I think I'd prefer to have a separate test for this, instead of modifying an existing one.

bkchr: Yes

derive/src/trait_bounds.rs
```diff
@@ -161,6 +161,7 @@ pub fn add<N>(
 
 		codec_types
 			.into_iter()
+			// .take(1)
```
pgherveou, line 164:
```suggestion
```

ggwpez (commented):
Wait, why is this adding a new function `max_compact_encoded_len` instead of fixing the impl of the `max_encoded_len` for the `Compact` wrapper?

tests/max_encoded_len.rs
pgherveou: done, this test would fail with

```
  Compiling parity-scale-codec v3.6.6 (/Users/pg/github/parity-scale-codec)                                                                                                                                                                                            
error[E0277]: the trait bound `Compact<T>: MaxEncodedLen` is not satisfied                                                                                                                                                                                              
  --> tests/max_encoded_len.rs:87:5                                                                                                                                                                                                                                     
   |                                                                                                                                                                                                                                                                    
87 |     t: T,                                                                                                                                                                                                                                                          
   |        ^ the trait `MaxEncodedLen` is not implemented for `Compact<T>`                                                                                                                                                                                             
   |                                                                                                                                                                                                                                                                    
help: consider extending the `where` clause, but there might be an alternative better way to express this requirement                                                                                                                                                   
   |                                                                                                                                                                                                                                                                    
84 | #[derive(Encode, MaxEncodedLen, Compact<T>: MaxEncodedLen)]                                                                    
   |                               +++++++++++++++++++++++++++
```

bkchr:
Pushed the proper solution. We add a new required trait for `HasCompact::Type`. But I doubt that someone implements this trait downstream.

We need to seal this trait.

bkchr (approved):

src/compact.rs
```diff
@@ -206,18 +208,39 @@ impl<'de, T> serde::Deserialize<'de> for Compact<T> where T: serde::Deserialize<
...
+#[cfg(feature = "max-encoded-len")]
+impl<T> MaxEncodedLen for Compact<T> where T: CompactAs, Compact<T::As>: MaxEncodedLen, Compact<T>: Encode {
+	fn max_encoded_len() -> usize {
+		Compact::<T::As>::max_encoded_len()
+	}
+}
+
 impl<T: 'static> HasCompact for T where
-	Compact<T>: for<'a> EncodeAsRef<'a, T> + Decode + From<Self> + Into<Self>
+	Compact<T>: for<'a> EncodeAsRef<'a, T> + Decode + From<Self> + Into<Self> + MaybeMaxEncodedLen
```
pgherveou, line 243: we can't use conditional compilation on just the constraint here right? That's why we need the MaybeMaxEncodedLen trait

bkchr: I didn't checked directly, but I don't think so.

pgherveou:
lgtm too, checking if everything compile properly on `polkadot-sdk` and will merge this next and create a new release

PR #515, opened 2023-09-08
Bump to 3.6.7

pgherveou:


ggwpez (approved):

ggwpez:
Did you try building Polkadot-SDK with this new version?

pgherveou:
> Did you try building Polkadot-SDK with this new version?

Yes I did, should I also run tests before releasing this?

ggwpez:
> Yes I did `cargo build` ran just fine, should I also run tests before releasing this?

I you want to be really sure then you can push a dienerd version for the CI, but i think its fine if it builds. Thanks!

PR #532, opened 2023-10-18
Add dedup_len trait

0xmovses:
This is a fix for https://github.com/paritytech/polkadot-sdk/issues/126

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -12,21 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{time::Duration, any::type_name, convert::{TryFrom, TryInto}};
+use std::{
+	any::type_name,
+	convert::{TryFrom, TryInto},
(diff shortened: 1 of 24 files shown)


bkchr:
Please revert the FMT commit.

ggwpez:
Wait... this adds a `DecodeDedupLength` trait for `BTreeSet`? I dont think thats possible. The advantage of the `DecodeLen` is that is does *not* have to decode everything, but only a `Compact<u32>`.  
What i meant in the other issue to just remove the `impl DecodeLen for BTreeSet` and replace it with a `DecodeNonDedupLen` or so.  
cc @bkchr

0xmovses:
Yeah that's clearer now, let me update this.

0xmovses:
I don't think we need to add this here anymore, as we are just adding a more descriptive trait in `polkado-sdk`, closing this.

PR #565, opened 2024-01-12
Add implementation of Decode for Box<str> and Box<[T]>

### LINKED ISSUE #564: Implement `Decode` for `Box<T>` where T: ?Sized
Why isn't `Decode` implemented for these types? The 2 obvious examples are `Box<str>` and `Box<[T]>`

mversic:
Closes #564

I can't think of a way to implement `Decode` for all unsized types

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -595,6 +595,77 @@ impl<T> WrapperTypeDecode for Box<T> {
 
 impl<T: DecodeWithMemTracking> DecodeWithMemTracking for Box<T> {}
 
+impl<T: Decode> Decode for Box<[T]> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let len = <Compact<u32>>::decode(input).map(|Compact(len)| len as usize)?;
+
+		input.descend_ref()?;
+
+		// Placement new is not yet stable, but we can just manually allocate a chunk of memory
+		// and convert it to a `Box` ourselves.
+		//
+		// The explicit types here are written out for clarity.
+		//
+		// TODO: Use `Box::new_uninit_slice` once that's stable.
+		let layout = core::alloc::Layout::array::<MaybeUninit<T>>(len)
+			.map_err(|_| Error::from("Item is too big and cannot be allocated"))?;
+
+		input.on_before_alloc_mem(layout.size())?;
+		let ptr: *mut MaybeUninit<T> = if layout.size() == 0 {
+			core::ptr::NonNull::dangling().as_ptr()
+		} else {
+			// SAFETY: Layout has a non-zero size so calling this is safe.
+			let ptr = unsafe { crate::alloc::alloc::alloc(layout) };
+
+			if ptr.is_null() {
+				crate::alloc::alloc::handle_alloc_error(layout);
+			}
+
+			ptr.cast()
+		};
+
+		// SAFETY: Constructing a `Box` from a piece of memory allocated with `std::alloc::alloc`
+		//         is explicitly allowed as long as it was allocated with the global allocator
+		//         and the memory layout matches.
+		//
+		//         Constructing a `Box` from `NonNull::dangling` is also always safe as long
+		//         as the underlying type is zero-sized.
+		let mut boxed_slice: Box<[MaybeUninit<T>]> = unsafe {
+			Box::from_raw(core::slice::from_raw_parts_mut(ptr, len))
+		};
+
+		for elem in &mut *boxed_slice {
+			T::decode_into(input, elem)?;
+		}
+
+		// Decoding succeeded, so let's get rid of `MaybeUninit`.
+		// TODO: Use `Box::assume_init` once that's stable.
+		let boxed_slice = Vec::from(boxed_slice)
+			.into_iter()
+			.map(|elem| unsafe { MaybeUninit::assume_init(elem) })
+			.collect();
+
+		input.ascend_ref();
+		Ok(boxed_slice)
+	}
+}
+
+impl<T: DecodeWithMemTracking> DecodeWithMemTracking for Box<[T]> {}
+
+impl Decode for Box<str> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		// Guaranteed to create a Vec with capacity == len
+		let vec = Vec::from(Box::<[u8]>::decode(input)?);
+		// Guaranteed not to reallocate the vec, only transmute to String
+		let str = String::from_utf8(vec).map_err(|_| "Invalid utf8 sequence")?;
+
+		assert_eq!(str.capacity(), str.len());
+		Ok(str.into_boxed_str())
+	}
+}
+
+impl DecodeWithMemTracking for Box<str> {}
+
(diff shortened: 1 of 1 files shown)


src/codec.rs
```diff
@@ -1058,6 +1058,18 @@ impl Encode for str {
...
 
+impl Decode for Box<str> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		Ok(String::decode(input)?.into_boxed_str())
+	}
+}
+
+impl<T: Decode> Decode for Box<[T]> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		Ok(Vec::decode(input)?.into_boxed_slice())
```
ggwpez, line 1069: So this is correct since `&[T]` implements `EncodeLike<Vec<T>>`?

bkchr: Every array is encoded the same way in Scale. First the length as compact and then the data.

src/codec.rs
```diff
@@ -1058,6 +1058,18 @@ impl Encode for str {
 	}
 }
 
+impl Decode for Box<str> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		Ok(String::decode(input)?.into_boxed_str())
```
bkchr, line 1063: Can you check the already existing `Box` `Decode` implementation. It should be in the same way as this implementation.

src/codec.rs
mversic: Will `Vec<T>::decode` create vector where `capacity` == `len`? 
If not, I will make an impl comparable to `Box<T>`

Otherwise, there is no downside in decoding to `Vec<T>` and then converting to `Box<[T]>`

src/codec.rs
mversic: Do you think that decoded `String` will have some excess capacity or will it have `capacity` == `len`?

mversic: I've taken a bit different route here

mversic: should I maybe use `WrapperTypeDecode` like this:

```rs
impl WrapperTypeDecode for Box<str> {
	type Wrapped = String;

	fn decode_wrapped<I: Input>(input: &mut I) -> Result<Self, Error> {
		// Guaranteed to create a Vec with capacity == len
		let vec = Vec::from(Box::<[u8]>::decode(input)?);
		// Guaranteed not to reallocate the vec, only transmute to String
		let str = String::from_utf8(vec).map_err(|_| "Invalid utf8 sequence")?;

		// At this point we have String with capacity == len,
		// therefore String::into_boxed_str will not reallocate
		Ok(str.into_boxed_str())
	}
}
```

src/codec.rs
```diff
@@ -554,25 +552,42 @@ impl<T> WrapperTypeDecode for Box<T> {
...
 		//
 		// TODO: Use `Box::assume_init` once that's stable.
-		let ptr: *mut MaybeUninit<T> = Box::into_raw(boxed);
-		let ptr: *mut T = ptr.cast();
+		let ptr: *mut T = Box::into_raw(boxed).cast();
 
 		// SAFETY: `MaybeUninit` doesn't affect the memory layout, so casting the pointer back
 		//         into a `Box` is safe.
-		let boxed: Box<T> = unsafe { Box::from_raw(ptr) };
+		let boxed = unsafe { Box::from_raw(ptr) };
```
koute, line 565: Please don't delete the explicit types nor "simplify" this code; they're here for readability in what is a highly tricky piece of code, *and* to make sure the compiler statically checks things for us.

We do *not* want the compiler to automatically infer things here. It might still work after being simplified, but then the next person will come, make a change, and then subtly break this code. The seemingly redundant types are a safeguard against that happening.

src/codec.rs
```diff
@@ -554,25 +552,42 @@ impl<T> WrapperTypeDecode for Box<T> {
...
+impl Decode for Box<str> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		// Guaranteed to create a Vec with capacity == len
+		let vec = Vec::from(Box::<[u8]>::decode(input)?);
+		// Guaranteed not to reallocate the vec, only transmute to String
+		let str = String::from_utf8(vec).map_err(|_| "Invalid utf8 sequence")?;
+
+		// At this point we have String with capacity == len,
+		// therefore String::into_boxed_str will not reallocate
+		Ok(str.into_boxed_str())
```
koute, line 587: Add `debug_assert_eq!`s for the `capacity == len`; then you don't need to spell it out in the comments.

src/codec.rs
```diff
@@ -1612,6 +1627,28 @@ mod tests {
...
+	fn boxed_slice_works() {
+		let v = vec![1u32, 2, 3, 4, 5, 6];
+		let b = v.clone().into_boxed_slice();
+
+		let encoded = b.encode();
+		assert_eq!(v.encode(), encoded);
+
+		assert_eq!(*b, *Box::<[u32]>::decode(&mut &b.encode()[..]).unwrap());
+	}
+
```
koute, line 1651: You've added a blanket impl for `Box<[T]>`, so please also add a test for a more complicated variable length type `T`, e.g. maybe `Box<[String]>` or something like that.

src/codec.rs
koute: > Will `Vec<T>::decode` create vector where `capacity` == `len`?

Not always; only if it takes the optimized code path. (So e.g. for `u8` it should, but not for any arbitrary `T`.)

src/codec.rs
mversic: restored

src/codec.rs
mversic: done

src/codec.rs
mversic: done

src/codec.rs
mversic: fixed, please check again

mversic:
hi, this PR should be ready for merge. Can someone approve it now?

nazar-pc:
Would be great to get an impl for `Arc<[T]>` as well

mversic:
> Would be great to get an impl for `Arc<[T]>` as well (see #633)

I don't want to complicate this PR further as it's not being reviewed as it is. 
I'd be happy to implement `Rc<[T]>`/`Arc<[T]>` in a separate PR.

koute (commented):
Sorry for the delay.

src/codec.rs
```diff
@@ -595,6 +595,77 @@ impl<T> WrapperTypeDecode for Box<T> {
...
+		let len = <Compact<u32>>::decode(input).map(|Compact(len)| len as usize)?;
+
+		input.descend_ref()?;
+
+		// Placement new is not yet stable, but we can just manually allocate a chunk of memory
+		// and convert it to a `Box` ourselves.
+		//
+		// The explicit types here are written out for clarity.
+		//
+		// TODO: Use `Box::new_uninit_slice` once that's stable.
```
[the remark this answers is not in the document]
mversic: see [this](https://github.com/paritytech/parity-scale-codec/pull/565/files#r2434970720)

nazar-pc:
@mversic do you plan to get this PR over the finish line?

PR #582, opened 2024-04-29
Fix PrefixInput read

kalaninja:
The current implementation of `PrefixInput` seems to be inconsistent. It checks for the buffer being empty, but only after the prefix is [taken](https://github.com/paritytech/parity-scale-codec/blob/5f4fb36df6b146dfa5417018ae831f92bab17400/src/compact.rs#L66), so trying to read with an zero-length buffer discards the prefix and it cannot be read any longer.

Though the behavior can be fixed for the empty buffer case, I think it is better just to remove the check completely and add a "note" comment, since this is an internal mechanism used to parse compacts only and the buffer is always allocated properly.

### DIFF
--- a/src/compact.rs
+++ b/src/compact.rs
@@ -62,8 +62,11 @@ impl<'a, T: 'a + Input> Input for PrefixInput<'a, T> {
 	}
 
 	fn read(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
+		if buffer.is_empty() {
+			return Ok(());
+		}
 		match self.prefix.take() {
-			Some(v) if !buffer.is_empty() => {
+			Some(v) => {
 				buffer[0] = v;
 				self.input.read(&mut buffer[1..])
 			}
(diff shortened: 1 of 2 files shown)


src/compact.rs
```diff
@@ -63,7 +64,7 @@ impl<'a, T: 'a + Input> Input for PrefixInput<'a, T> {
 
 	fn read(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
 		match self.prefix.take() {
```
bkchr, line 66: The correct solution would be that the check `if buffer.is_empty()` is moved before the match.

bkchr (changes requested):

src/compact.rs
```diff
@@ -62,8 +62,11 @@ impl<'a, T: 'a + Input> Input for PrefixInput<'a, T> {
 	}
 
 	fn read(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
+		if buffer.is_empty() {
+			return Ok(());
+		}
 		match self.prefix.take() {
-			Some(v) if !buffer.is_empty() => {
+			Some(v) => {
```
ggwpez, line 69: If the subsequent read fails, then it cannot be retried, since the `prefix` was already moved out.

I guess you could put it back in the error case, not sure what is expected behaviour for this?

bkchr (approved):

src/compact.rs
kalaninja: The Input trait doesn't say anything about providing guarantees  of correct subsequent reads after a failure. As I can see `IoReader` has a similar behavior as well, since it relies on `read_exact` which is "If this function returns an error, it is unspecified how many bytes it has read..."

PR #602, opened 2024-05-28
Decode with mem limit

ggwpez:
Adds a `try_alloc` function to the `Input` trait to track allocations. This makes it necessary to implement it in the  `decode` any type, but is not type checked. Currently it is only implemented for vectors.

Some tests are in place to showcase the usage.

<s>This MR adds functionality to restrict the memory usage while decoding a type.  
Some tests show how to use it with nested types or tuples, but mostly a PoC to evaluate the approach.  
We could also do it like the `DecodeLimit` and track the memory consumption in the `Input` stream, but it cannot be enforced by the type system.

PS: I will change it to track the limit in the `Input`, as it allows to also track the other limits (like recursion depth) at the expense of not being type checked.

**Changes**
Adds a generic `DecodeWithContext` that allows a `decode_with_context` call to pass down some context when decoding.

This new trait is then used together with `MemLimit` such that `Vec` implements `DecodeWithContext<MemLimit>`.  
There is a blanket in place that converts `DecodeWithContext<MemLimit>` to `DecodeWithMemLimit`, and therebyte, `Vec` is now `DecodeWithMemLimit`.</s>

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -85,6 +85,11 @@ pub trait Input {
 	/// This is called when decoding reference-based type is finished.
 	fn ascend_ref(&mut self) {}
 
+	/// Try to allocate a contiguous chunk of memory of `size` bytes.
+	fn try_alloc(&mut self, size: usize) -> Result<(), Error> {
+		Ok(())
+	}
+
 	/// !INTERNAL USE ONLY!
 	///
 	/// Decodes a `bytes::Bytes`.
@@ -842,6 +847,8 @@ pub fn decode_vec_with_len<T: Decode, I: Input>(
 	input: &mut I,
 	len: usize,
 ) -> Result<Vec<T>, Error> {
+	input.try_alloc(mem::size_of::<T>().saturating_mul(len))?;
+
 	fn decode_unoptimized<I: Input, T: Decode>(
 		input: &mut I,
 		items_len: usize,
@@ -1187,6 +1194,7 @@ impl<T: EncodeLike<U>, U: Encode> EncodeLike<Vec<U>> for &[T] {}
 
 impl<T: Decode> Decode for Vec<T> {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		input.try_alloc(mem::size_of::<Self>())?;
 		<Compact<u32>>::decode(input)
 			.and_then(move |Compact(len)| decode_vec_with_len(input, len as usize))
 	}
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -280,6 +280,10 @@ pub trait DecodeLength {
 	fn len(self_encoded: &[u8]) -> Result<usize, Error>;
 }
 
+/// Decoding ran out of memory.
+// TODO: Custom error variant
+pub const DECODE_OOM_ERROR: &str = "OOM while decoding";
+
 /// Trait that allows zero-copy read of value-references from slices in LE format.
 pub trait Decode: Sized {
 	// !INTERNAL USE ONLY!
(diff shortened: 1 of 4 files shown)


src/codec.rs
```diff
@@ -331,6 +335,135 @@ pub trait Decode: Sized {
...
+		input: &mut I,
+		ctx: &mut C,
+	) -> Result<Self, Error> {
+		Ok((for_tuples!( #( Tuple::decode_with_context(input, ctx)? ),* )))
+	}
+}
+
+/// Decode some data while being constrained on the allocated memory.
+pub trait DecodeMemLimited: Sized {
+	const TYPE_INFO: TypeInfo = TypeInfo::Unknown;
```
ggwpez, line 362: It looks a bit odd to me that we need this here, but without it, the `decode_vec_with...` function calls goes into the `decode_unoptimized` branch instead of doing the unsafe cast.

src/memory_limit.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::{Decode, Error, Input};
+
+/// The error message returned when the memory limit is reached.
+const DECODE_OOM_ERROR: &str = "Out of memory when decoding";
+
+/// Extension trait to [`Decode`] for decoding with a maximum memory consumption.
+///
+/// Should be used as a marker for types that do the memory tracking in their `decode` implementation with [`Input::try_alloc`].
+pub trait DecodeMemLimit: Decode + Sized {
+	fn decode_with_mem_limit<I: Input>(limit: MemLimit, input: &mut I) -> Result<Self, Error> {
```
serban300, line 24: This approach seems a bit limiting since we can only do `decode_with_mem_limit` or `decode_with_depth_limit`. Looks like we can't combine them. It would be nice to be able to do something more generic. For example something like:

`decode_with_check(input: &mut I, check: FnMut(DecodeContext) -> Result<(), Error>)` 

where `DecodeContext` would be something like

```
DecodeContext {
    depth: u32,
    used_memory: usize,
    ...
}
```

This way we could check both for depth and memory limit at the same time. Maybe for other parameters as well. Also would be nice to be able to keep track of the number of objects of a certain type that we encounter recursively while decoding since this is also something that we needed in Polkadot.

serban300 (commented):
Thank you for the suggestion and for opening the PR ! The high level idea looks good ! I just left a couple of comments.

Also could you please let me know how you would prefer to move forward from here ? Would you have time to continue this PR ? If not I would be happy to take it since we need it for the Bridges + XCM.

src/memory_limit.rs
ggwpez: > Looks like we can't combine them. 

We can. Which is why i chose this approach. These functions are just wrappers, but there is nobody stopping you from doing:

```rust
let mut input = MemTrackingInput { inner: DepthTrackingInput { inner: .. }}
```

>This way we could check both for depth and memory limit at the same time. Maybe for other parameters as well. Also would be nice to be able to keep track of the number of objects of a certain type that we encounter recursively while decoding since this is also something that we needed in Polkadot.

Yea, i had some `decode_with_context<C>(..)` implemented before, but it requires adapting all implementations of the type and forced them to use a specific `Context` type, unless we make the context a trait, but that clusterfucks everything.

src/memory_limit.rs
```diff
@@ -0,0 +1,182 @@
...
+	}
+}
+
+/// Alignment of some amount of memory.
+///
+/// Normally the word `alignment` is used in the context of a pointer - not an amount of memory, but this is still
+/// the most fitting name.
+pub enum MemAlignment {
+	/// Round up to the next power of two.
+	NextPowerOfTwo,
```
ggwpez, line 84: Note: Currently when using this with a Vector, we align each element individually.  
But a Vec allocates all elements in a chunk through the WASM bump allocator. So what we have to do is `round(sum(e...))` over all elements instead of `sum(round(e)..)`.  

Otherwise it would round each element size up to the next power of two, but the vector may still allocate more since it aligns with the sum of the sizes of all elements.

serban300:
I've been thinking about this some more and I'm not sure if this solution is feasible. My main concern is related to enums. For example, for the following enum:
```
enum TestEnum {
    A {field: String},
    B {field: u32},
    C {field: [u8; 10000]},
}
```

- for `TestEnum::A` we need to account for `mem::size_of<TestEnum>` + the size of the decoded String. 
- for `TestEnum::B` and `TestEnum::C` we need to account just for `mem::size_of<TestEnum>` 

I don't know if we can check the type of each field in an enum variant and call `try_alloc()` just for some certain types. And even if we could, this process would be manual and seems very error-prone.

Another concern is that for structures like `BTreeMap`, `BTreeSet`, etc, we'll only record the memory of the elements, but we won't account for the memory of the internal nodes of the map/set. Altough this is not significant, it's still not ideal.

Another problem would be if we have skipped fields that consume significant memory. For example a `[u8; 10000]`.

Another issue is that here we do `impl<T: Decode> DecodeMemLimit for Vec<T> {}`. This will track memory only for the structures for which we implement `DecodeMemLimit`. For custom structures memory tracking will work just to the extent to which they actually call `decode()` for the primitive types inside them. I think the right approach would be `impl<T: DecodeMemLimit> DecodeMemLimit for Vec<T> {}`, but then the problem is that there will be a lot of custom structures for which `DecodeMemLimit` is not implemented and it would be a burden to implement it for all.

Also, since I was looking on these, I think there are some issues with the `decode_with_depth_limit` as well. 
1. it also does `impl<T: Decode> DecodeLimit for T {` . This might miss structures.
2. for example should't we do `input.descend_ref()` and `input.ascend_ref()` when decoding an enum ?

I'm not sure. Maybe I'm missing something. @ggwpez  , @bkchr  what do you think ?

serban300:
We'll limit

> I've been thinking about this some more and I'm not sure if this solution is feasible. My main concern is related to enums. For example, for the following enum:
> 
> ```
> enum TestEnum {
>     A {field: String},
>     B {field: u32},
>     C {field: [u8; 10000]},
> }
> ```
> 
>     * for `TestEnum::A` we need to account for `mem::size_of<TestEnum>` + the size of the decoded String.
> 
>     * for `TestEnum::B` and `TestEnum::C` we need to account just for `mem::size_of<TestEnum>`
> 
> 
> I don't know if we can check the type of each field in an enum variant and call `try_alloc()` just for some certain types. And even if we could, this process would be manual and seems very error-prone.
> 
> Another concern is that for structures like `BTreeMap`, `BTreeSet`, etc, we'll only record the memory of the elements, but we won't account for the memory of the internal nodes of the map/set. Altough this is not significant, it's still not ideal.
> 
> Another problem would be if we have skipped fields that consume significant memory. For example a `[u8; 10000]`.
> 
> Another issue is that here we do `impl<T: Decode> DecodeMemLimit for Vec<T> {}`. This will track memory only for the structures for which we implement `DecodeMemLimit`. For custom structures memory tracking will work just to the extent to which they actually call `decode()` for the primitive types inside them. I think the right approach would be `impl<T: DecodeMemLimit> DecodeMemLimit for Vec<T> {}`, but then the problem is that there will be a lot of custom structures for which `DecodeMemLimit` is not implemented and it would be a burden to implement it for all.
> 
> Also, since I was looking on these, I think there are some issues with the `decode_with_depth_limit` as well.
> 
>     1. it also does `impl<T: Decode> DecodeLimit for T {` . This might miss structures.
> 
>     2. for example should't we do `input.descend_ref()` and `input.ascend_ref()` when decoding an enum ?
> 
> 
> I'm not sure. Maybe I'm missing something. @ggwpez , @bkchr what do you think ?

We'll limit this to tracking only heap size, but we want to make it type safe.

Tracked as part of: https://github.com/paritytech/parity-scale-codec/issues/609

Will close this draft for the moment and will open a separate PR later

PR #605, opened 2024-06-19
Use `MAX_PREALLOCATION` consistently

serban300:
Related to #609 

Use `MAX_PREALLOCATION` both when reading a vec from bytes and when decoding each element.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -1253,32 +1253,9 @@ impl<T: Encode> Encode for VecDeque<T> {
 	fn encode_to<W: Output + ?Sized>(&self, dest: &mut W) {
 		compact_encode_len_to(dest, self.len()).expect("Compact encodes length");
 
-		macro_rules! encode_to {
-			( $ty:ty, $self:ident, $dest:ident ) => {{
-				if cfg!(target_endian = "little") || mem::size_of::<T>() == 1 {
-					let slices = $self.as_slices();
-					let typed =
-						unsafe { core::mem::transmute::<(&[T], &[T]), (&[$ty], &[$ty])>(slices) };
-
-					$dest.write(<[$ty] as AsByteSlice<$ty>>::as_byte_slice(typed.0));
-					$dest.write(<[$ty] as AsByteSlice<$ty>>::as_byte_slice(typed.1));
-				} else {
-					for item in $self {
-						item.encode_to($dest);
-					}
-				}
-			}};
-		}
-
-		with_type_info! {
-			<T as Encode>::TYPE_INFO,
-			encode_to(self, dest),
-			{
-				for item in self {
-					item.encode_to(dest);
-				}
-			},
-		}
+		let slices = self.as_slices();
+		encode_slice_no_len(slices.0, dest);
+		encode_slice_no_len(slices.1, dest);
 	}
 }
 
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -47,7 +47,7 @@ use crate::{
 	DecodeFinished, Error,
 };
 
-pub(crate) const MAX_PREALLOCATION: usize = 4 * 1024;
+pub(crate) const MAX_PREALLOCATION: usize = 16 * 1024;
 const A_BILLION: u32 = 1_000_000_000;
 
 /// Trait that allows reading of data into a slice.
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -834,52 +834,6 @@ pub(crate) fn encode_slice_no_len<T: Encode, W: Output + ?Sized>(slice: &[T], de
 	}
 }
 
-/// Decode the vec (without a prepended len).
-///
-/// This is equivalent to decode all elements one by one, but it is optimized in some
-/// situation.
-pub fn decode_vec_with_len<T: Decode, I: Input>(
-	input: &mut I,
-	len: usize,
-) -> Result<Vec<T>, Error> {
-	fn decode_unoptimized<I: Input, T: Decode>(
-		input: &mut I,
-		items_len: usize,
-	) -> Result<Vec<T>, Error> {
-		let input_capacity = input
-			.remaining_len()?
-			.unwrap_or(MAX_PREALLOCATION)
-			.checked_div(mem::size_of::<T>())
-			.unwrap_or(0);
-		let mut r = Vec::with_capacity(input_capacity.min(items_len));
-		input.descend_ref()?;
-		for _ in 0..items_len {
-			r.push(T::decode(input)?);
-		}
-		input.ascend_ref();
-		Ok(r)
-	}
-
-	macro_rules! decode {
-		( $ty:ty, $input:ident, $len:ident ) => {{
-			if cfg!(target_endian = "little") || mem::size_of::<T>() == 1 {
-				let vec = read_vec_from_u8s::<_, $ty>($input, $len)?;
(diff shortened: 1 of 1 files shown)


serban300:
@bkchr @koute could you PTAL on this PR since it's somewhat related to #609 ? Even though it's not a prerequisite, I think it would be nice to have.

bkchr (approved):

src/codec.rs
```diff
@@ -1113,71 +1067,110 @@ impl<T: Encode> Encode for [T] {
...
+		let chunk_len = chunk_len.min(num_undecoded_items);
+		decoded_vec.reserve_exact(chunk_len);
 
-	let input_len = input.remaining_len()?;
+		decode_chunk(&mut decoded_vec, chunk_len)?;
 
-	// If there is input len and it cannot be pre-allocated then return directly.
-	if input_len.map(|l| l < byte_len).unwrap_or(false) {
-		return Err("Not enough data to decode vector".into());
+		num_undecoded_items = num_undecoded_items.saturating_sub(chunk_len);
```
koute, line 1085: This `saturating_sub`'s completely unnecessary here since impossible to have `chunk_len > num_undecoded_items` due to the `min`.

src/codec.rs
```diff
@@ -1113,71 +1067,110 @@ impl<T: Encode> Encode for [T] {
...
+where
+	T: Decode,
+	I: Input,
+{
+	// Check if there is enough data in the input buffer.
+	if let Some(input_len) = input.remaining_len()? {
+		if input_len < len {
+			return Err("Not enough data to decode vector".into());
+		}
+	}
```
koute, line 1133: This isn't correct as deserializing `T` might take any number of bytes (including even zero bytes, e.g. `()`).

What we should do here is to have a `serialized_size_hint()` method (or, more specifically, probably an associated const so that it can be checked statically to fit within `MAX_PREALLOCATION`) or something like that on `T` which would return a value that could allow this check. (We already have `encoded_fixed_size` there, but that returns an *exact* number of bytes; it could be used here, but technically that's too strict and we can do better here by using the minimum.)

src/codec.rs
```diff
@@ -1113,71 +1067,110 @@ impl<T: Encode> Encode for [T] {
...
-/// The encoding of `T` must be equal to its binary representation, and size of `T` must be less or
-/// equal to [`MAX_PREALLOCATION`].
-pub(crate) fn read_vec_from_u8s<I, T>(input: &mut I, items_len: usize) -> Result<Vec<T>, Error>
+fn decode_vec_chunked<T, F>(len: usize, mut decode_chunk: F) -> Result<Vec<T>, Error>
 where
-	I: Input,
-	T: ToMutByteSlice + Default + Clone,
+	F: FnMut(&mut Vec<T>, usize) -> Result<(), Error>,
 {
 	debug_assert!(MAX_PREALLOCATION >= mem::size_of::<T>(), "Invalid precondition");
```
koute, line 1074: We should make this into a static assert and check it at compile time.

src/codec.rs
koute: Yeah, alternatively we can drop it. Although having it here can have one benefit - if we end up not having enough data then this will return an early error instead of wasting time trying to deserialize it. Nice to have, but not strictly necessary.

serban300:
Thanks for the review ! Will address the comments in a new PR today.

src/codec.rs
bkchr: To implement this, we would need to write quite a lot of code. For example for an `enum` we would need to know the variant that requires the least amount of bytes. However, it could then still fail at decoding because we try to decode always the enum variant that uses much more bytes etc.

koute: Hm, well, would it be *that* much code? I implemented this in my serialization crate and it's mostly fine; with `enum`s you essentially just autogenerate a `(min(variant1, variant2, ..), max(variant1, variant2, ..))` in your impl. Of course this is just an optimization (in some cases it would make incomplete deserializations fail early, and in some cases it would allow the compiler to remove per-element size checks), and as you've said it can still fail at decoding depending on what you're decoding.

Anyway, I'm fine with going with your suggestion to just delete the check.

src/codec.rs
serban300: I couldn't manage to do this so far. I tried something like

```
const _: () = {
    assert!(MAX_PREALLOCATION >= mem::size_of::<T>())
}
```

inside `decode_vec_chunked()`

But I'm getting an error: `can't use generic parameters from outer item`.

Any suggestion would be helpful

src/codec.rs
serban300: Addressed in #615

src/codec.rs
serban300: Dropped it: #615

src/codec.rs
koute: You don't need to define a constant; since Rust 1.79 you can use a `const {}` block to force const evaluation of an expression.

serban300: Yes, this works, thanks ! PTAL on #615 

But the CI fails, because the CI image uses rust 1.73.0 .  We can either try to release a `paritytech/ci-unified:bullseye-1.79.0` image. Will check how this can be done.

src/codec.rs
```diff
@@ -1113,71 +1067,110 @@ impl<T: Encode> Encode for [T] {
...
+		let decoded_vec_len = decoded_vec.len();
+		let decoded_vec_size = decoded_vec_len * mem::size_of::<T>();
+		unsafe {
+			decoded_vec.set_len(decoded_vec_len + chunk_len);
+		}
 
-		while items_remains > 0 {
-			let items_len_read = max_preallocated_items.min(items_remains);
+		let bytes_slice = decoded_vec.as_mut_byte_slice();
+		input.read(&mut bytes_slice[decoded_vec_size..])
```
nazar-pc, line 1119: This is not the right way of doing it and actually **unsound**.

`Input::read()` is a safe method of a safe trait, it doesn't guarantee an invariant of not reading its argument. So it is possible to have a perfectly safe implementation of `Input` that reads some bytes before writing to them, but since they are uninitialized (you just blindly called `Vec::set_len()`), it is an instant undefined behavior!

The right thing to do is to have a separate unsafe method that takes a pointer or add a method that takes something like `&mut [MaybeUninit<T>]` instead.

koute: Yes, you're right. A few points though:

* A lot of this code traces back to a time where the rules about UB weren't entirely clear, and it was still an open question whether reading uninitialized memory of a type which doesn't have a niche is UB or not.
* The "proper" blessed way of doing this (that is, `Vec::spare_capacity_mut`) was introduced long after some of the code in this crate was written.
* Changing this means we need to break backwards compatibility (which isn't technically a problem per se, it's just painful because of all of the extra churn as now everyone needs to update).
* Unless the `read` actually reads the memory it's not actually a problem, both in practice and in theory (but it's technically a landmine because there's no `unsafe` and reading the memory is technically unsound).
* Even *if* `read` actually reads the uninitialized memory currently nothing bad actually happens (it's only UB in theory AFAIK), and it has been like this for years.

I'm not trying to excuse this, just trying to be pragmatic, as this is essentially an issue of "this code was designed when the rules about UB weren't clear, and now we have a function which should be marked `unsafe` but it isn't because it can trigger UB when misused".

Anyway, feel free to create an issue about this. If/when we'll be bumping the major version we'd definitely want to fix this.

PR #615, opened 2024-07-22
Follow-up on #605

serban300:
Addressing the code review comments for https://github.com/paritytech/parity-scale-codec/pull/605

Also calling `descend_ref()` and `ascend_ref()` for `read_vec_from_u8s()`. Not sure if it was desired, or it was just a miss. To me it seems like they should be called. I hope I understood correctly. 

**Please note that this PR also updates the rust toolchain used by the CI from 1.73.0 to version 1.79.0.** We need this in order to use `inline const`.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -1110,8 +1110,7 @@ where
 		}
 	}
 
-	input.descend_ref()?;
-	let vec = decode_vec_chunked(len, |decoded_vec, chunk_len| {
+	decode_vec_chunked(len, |decoded_vec, chunk_len| {
 		let decoded_vec_len = decoded_vec.len();
 		let decoded_vec_size = decoded_vec_len * mem::size_of::<T>();
 		unsafe {
@@ -1120,10 +1119,7 @@ where
 
 		let bytes_slice = decoded_vec.as_mut_byte_slice();
 		input.read(&mut bytes_slice[decoded_vec_size..])
-	})?;
-	input.ascend_ref();
-
-	Ok(vec)
+	})
 }
 
 fn decode_vec_from_items<T, I>(input: &mut I, len: usize) -> Result<Vec<T>, Error>
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -1110,7 +1110,8 @@ where
 		}
 	}
 
-	decode_vec_chunked(len, |decoded_vec, chunk_len| {
+	input.descend_ref()?;
+	let vec = decode_vec_chunked(len, |decoded_vec, chunk_len| {
 		let decoded_vec_len = decoded_vec.len();
 		let decoded_vec_size = decoded_vec_len * mem::size_of::<T>();
 		unsafe {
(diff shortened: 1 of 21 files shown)


src/codec.rs
```diff
@@ -1082,7 +1084,7 @@ where
 
 		decode_chunk(&mut decoded_vec, chunk_len)?;
 
-		num_undecoded_items = num_undecoded_items.saturating_sub(chunk_len);
+		num_undecoded_items -= chunk_len;
```
ggwpez, line 1087: I think for such critical code its fine to stay a bit defensive with `saturating` or `checked` even if it *should* not happen.

serban300: Personally I agree. I would stay on the safe side. @koute  WDYT ?

bkchr: We already do the `saturated_sub` by doing the `chunk_len.min(num_undecoded_items)` above.

src/codec.rs
```diff
@@ -1108,7 +1110,8 @@ where
 		}
 	}
 
-	decode_vec_chunked(len, |decoded_vec, chunk_len| {
+	input.descend_ref()?;
+	let vec = decode_vec_chunked(len, |decoded_vec, chunk_len| {
```
bkchr, line 1114:
```suggestion
	decode_vec_chunked(len, |decoded_vec, chunk_len| {
```

We are not descending here any further. We just read the straight bytes from the vector.

TBH, the entire "when to descend" is a little bit blurry, but I don't see this case here.

src/codec.rs
```diff
@@ -1117,21 +1120,17 @@ where
 
 		let bytes_slice = decoded_vec.as_mut_byte_slice();
 		input.read(&mut bytes_slice[decoded_vec_size..])
-	})
+	})?;
+	input.ascend_ref();
+
+	Ok(vec)
```
bkchr, line 1126:
```suggestion
	})
```

src/codec.rs
bkchr: Because the idea is just that you want to prevent too much nesting.

bkchr (approved):
This time please wait for @koute to review :D

src/codec.rs
serban300: I see it more like a safety measure., In case someone changes the logic sometime in the future, and the min disappears somehow, but forgets to then also use `saturated_sub`. But I agree that it's a very unlikely scenario. I'm ok with keeping it as it is.

src/codec.rs
serban300: Ok, I don't know. It seemed like we are still descending in the vec, even though we are reading the bytes. Also seemed strange for example that `Box<u32>` descends, but `Vec<u32>` doesn't. But anyway, I don't know and I don't have strong preferences. Reverted this change.

src/codec.rs
serban300: Done

src/codec.rs
koute: In general I don't necessarily like including useless code for extra "safety" as it muddles the intent of the code, and then whoever is reading is left scratching their head why that code is there.

If we need to ensure that something is true then I prefer, in no particular order: a) adding extra asserts, b) adding tests, c) producing and enforcing a proof that the condition is impossible (quite easy to do in a lot of cases with something like [Kani](https://github.com/model-checking/kani), which is essentially black magic).

koute (approved):

Issue #609, opened 2024-07-05
Limit `decode()` heap memory usage


serban300:
We would like to be able to specify a rough limit for the heap memory usage while decoding a structure.

We have this draft PR as a starting point: https://github.com/paritytech/parity-scale-codec/pull/602 . The problem with this approach is that it is intentionally not type safe, for backwards compatibility. We would like to make it type-safe and release it as part of v4.

serban300:
The PR was merged. Resolving.

PR #616, opened 2024-07-24
Decode with mem limit

serban300:
Related to https://github.com/paritytech/parity-scale-codec/issues/609


gui1117:
The current implementation is enough for the usage or should we also implement it automatically in derive macro?

src/codec.rs
```diff
@@ -466,17 +486,29 @@ where
...
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-		input.scale_internal_decode_bytes()
+		let len = <Compact<u32>>::decode(input)?.0 as usize;
+		if input.__private_bytes_cursor().is_some() {
+			input.on_before_alloc_mem(len)?;
+		}
+
+		if let Some(bytes_cursor) = input.__private_bytes_cursor() {
+			bytes_cursor.decode_bytes_with_len(len)
+		} else {
```
gui1117, line 503:
```suggestion
		if let Some(bytes_cursor) = input.__private_bytes_cursor() {
			input.on_before_alloc_mem(len)?;
			bytes_cursor.decode_bytes_with_len(len)
		} else {
```

is it more readable?

src/codec.rs
```diff
@@ -1207,13 +1274,20 @@ macro_rules! impl_codec_through_iterator {
 			fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 				<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 					input.descend_ref()?;
-					let result = Result::from_iter((0..len).map(|_| Decode::decode(input)));
+					let result = Result::from_iter((0..len).map(|_| {
+						input.on_before_alloc_mem(0 $( + mem::size_of::<$generics>() )*)?;
```
gui1117, line 1278: This is an underestimate of the allocation, is it fine?

All the collections below will use more space than what is reported.

maybe it should be documented in `Input` that it is a rough estimation then.

src/codec.rs
```diff
@@ -1260,6 +1334,8 @@ impl<T: Decode> Decode for VecDeque<T> {
 	}
 }
 
+impl<T: DecodeWithMemTracking> DecodeWithMemTracking for VecDeque<T> {}
+
 impl EncodeLike for () {}
 
 impl Encode for () {
```
gui1117, line 1341: It seems `DecodeWithMemTracking` is missed for tuples and for `()`.

gui1117 (commented):
overall looks good,

src/codec.rs
serban300: Yes, this is as desired. It is acceptable to be a bit off. The main thing is to track the memory used by the decoded data inside the type.

You're right, it's good to document this. Added a comment in `Input`.

src/codec.rs
gui1117: ah indeed, my bad

src/codec.rs
gui1117: great :+1:

src/codec.rs
gui1117: ah I missed it, no it is good as it is.

gui1117:
> Yes, it would be nice to also implement it automatically. Normally it should be an auto trait, but this feature is still unstable so I couldn't use that.
> 
> Implementing it automatically in the derive macros would be the next best thing, but I didn't manage to do it. For example for:
> 
> ```
> struct Test 
> {
>     a: u32
> }
> ```
> 
> I tried something that would expand to:
> 
> ```
> impl DecodeWithMemTracking for Test where u32: DecodeWithMemTracking {
> }
> ```
> 
> Which generates compile errors.
> 
> So probably we'll need to implement it manually for each custom struct.

I see we can't derive it automatically with `Decode` derive-macro, but we could provide a `DecodeWithMem` derive-macro.

This macro will do the bound same as Encode/Decode, i.e. bound only the generic types. and it will fail if some concrete types don't implement `DecodeWithMemTracking`.

It would still be more handy than manually write the code I guess.

src/mem_tracking.rs
```diff
@@ -0,0 +1,86 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::{Decode, Error, Input};
+use impl_trait_for_tuples::impl_for_tuples;
+
+/// Marker trait used for identifying types that call the mem tracking hooks exposed by `Input`
+/// while decoding.
+pub trait DecodeWithMemTracking: Decode {}
```
gui1117, line 21: I guess it would be great to implement it also on:
* `Compact<T>` where `Compact<T::As>: DecodeWithMemTracking`
* all the explicit compact: `Compact<()>`, `Compact<u8>`...`Compact<u128>`
* stuff in bit_vec module
* stuff in generic_array module

tests/mem_tracking.rs
```diff
@@ -0,0 +1,132 @@
...
+			test_enum: TestEnum::Array([0; 1000]),
+			boxed_test_enum: Box::new(TestEnum::Empty),
+			box_field: Box::new(1),
+			vec: vec![1; 10],
+		},
+		usize::MAX,
+		1015
+	)
+	.is_ok())
+}
```
gui1117, line 132: maybe one more test where the decoding fails because mem limit is reached.

gui1117 (commented):
I suggested some more types to implement the trait on.

But part from this it looks good to me.

src/mem_tracking.rs
serban300: Done for all except `generic_array` with no additional changes needed.

With `generic_array` I'm not familiar and I don't know if anyone uses it. In polkadot-sdk it doesn't seem to be used. We could implement it later if someone requested it.

tests/mem_tracking.rs
serban300: Good point ! Done.

serban300:
> I see we can't derive it automatically with `Decode` derive-macro, but we could provide a `DecodeWithMem` derive-macro.
> 
> This macro will do the bound same as Encode/Decode, i.e. bound only the generic types. and it will fail if some concrete types don't implement `DecodeWithMemTracking`.
> 
> It would still be more handy than manually write the code I guess.

I'm not sure if it's possible. Unless we add a mock method to the `DecodeWithMemTracking` trait and try to call it in the derived implementation. Which would be a bit of a hack. But even so, implementing it manually isn't so hard. It's just an empty marker trait. I'm not sure if deriving it would make things much easier.

src/mem_tracking.rs
gui1117: Yes we can implement generic_array on demand.

src/codec.rs
```diff
@@ -408,12 +420,32 @@ mod feature_wrapper_bytes {
 	impl EncodeLike<Bytes> for Vec<u8> {}
 }
 
+/// `Input` implementation optimized for decoding `bytes::Bytes`.
 #[cfg(feature = "bytes")]
-struct BytesCursor {
+pub struct BytesCursor {
```
ggwpez, line 425: Why are you making it pub? Then we have to include it our public SemVer API.

gui1117:
> > I see we can't derive it automatically with `Decode` derive-macro, but we could provide a `DecodeWithMem` derive-macro.
> > This macro will do the bound same as Encode/Decode, i.e. bound only the generic types. and it will fail if some concrete types don't implement `DecodeWithMemTracking`.
> > It would still be more handy than manually write the code I guess.
> 
> I'm not sure if it's possible. Unless we add a mock method to the `DecodeWithMemTracking` trait and try to call it in the derived implementation. Which would be a bit of a hack. But even so, implementing it manually isn't so hard. It's just an empty marker trait. I'm not sure if deriving it would make things much easier.

I see it now, indeed.

Then ok

src/codec.rs
```diff
@@ -1207,13 +1280,20 @@ macro_rules! impl_codec_through_iterator {
 			fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 				<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 					input.descend_ref()?;
-					let result = Result::from_iter((0..len).map(|_| Decode::decode(input)));
+					let result = Result::from_iter((0..len).map(|_| {
+						input.on_before_alloc_mem(0 $( + mem::size_of::<$generics>() )*)?;
```
ggwpez, line 1284: Since this is low level runtime code, we need to be extremely paranoid about panics. Please use saturating math.

src/mem_tracking.rs
ggwpez, line 21: Can you mention what hook it calls please?

tests/mem_tracking.rs
```diff
@@ -0,0 +1,158 @@
...
+struct ComplexStruct {
+	test_enum: TestEnum,
+	boxed_test_enum: Box<TestEnum>,
+	box_field: Box<u32>,
+	vec: Vec<u8>,
+}
+
+impl DecodeWithMemTracking for ComplexStruct {}
+
+fn decode_object<T>(obj: T, mem_limit: usize, expected_used_mem: usize) -> Result<T, Error>
```
ggwpez, line 47: This function could also call the decode again with one byte less memory limit to check that it fails.  
Then you get all the error cases also covered.

src/codec.rs
bkchr: `mem::size_of` is a constant function, so it should already fail at compile time.

bkchr (commented):
Did you try this will work on Substrate? I mean you will need to ensure that `Call` implements `DecodeWithMemTracking` and all the types we are using there.

Also you should expose a function `decode_with_mem_limit` instead of exposing the `MemLimitInput`

serban300:
> Did you try this will work on Substrate? I mean you will need to ensure that `Call` implements `DecodeWithMemTracking` and all the types we are using there.

Oh, you're right. I can patch polkadot-sdk locally and try this. Will do it and get back.

> Also you should expose a function `decode_with_mem_limit` instead of exposing the `MemLimitInput`

Yes, will do this too.

src/codec.rs
```diff
@@ -85,16 +86,27 @@ pub trait Input {
...
+	}
+
 	/// !INTERNAL USE ONLY!
 	///
-	/// Decodes a `bytes::Bytes`.
+	/// Used when decoding a `bytes::Bytes` from a `BytesCursor` input.
 	#[cfg(feature = "bytes")]
 	#[doc(hidden)]
-	fn scale_internal_decode_bytes(&mut self) -> Result<bytes::Bytes, Error>
+	fn __private_bytes_cursor(&mut self) -> Option<&mut BytesCursor>
```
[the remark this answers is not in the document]
bkchr: What you are saying there makes no real sense. Decoding `Bytes` from bytes currently doesn't work with any other input. This is by design. So, the changes you have done here are not really required.



> The same applies for combining `BytesCursor` with `DepthLimit` or anything else.

You can simply not do this right now. `BytesCursor` is right now not exposed to the outside and thus, you can also not use it as `input` for `DepthLimit`.

In the end what you would need to do is that you forward the `scale_internal_decode_bytes` call in `DepthLimit` to the inner input for example. Still, this would require that `BytesCursor` is public, which we should not do right now.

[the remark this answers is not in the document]
serban300: Hmmm yes, makes sense. Because `BytesCursor` is private. I didn't think of that. Ok, reverted this change then.

src/codec.rs
```diff
@@ -1207,13 +1258,20 @@ macro_rules! impl_codec_through_iterator {
 			fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
 				<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
 					input.descend_ref()?;
-					let result = Result::from_iter((0..len).map(|_| Decode::decode(input)));
+					let result = Result::from_iter((0..len).map(|_| {
```
bkchr, line 1261: Most of the time containers are implemented to always increase the capacity by two when hitting the limit. This means here that we may only decode 190 elements while the capacity is already at 256. What I want to say is that you will not track the size correctly.

src/codec.rs
serban300: Reverted the change. Not applicable anymore.

src/codec.rs
```diff
@@ -408,12 +420,32 @@ mod feature_wrapper_bytes {
...
 
+#[cfg(feature = "bytes")]
+impl BytesCursor {
+	/// Create a new instance of `BytesCursor`.
+	pub fn new(bytes: bytes::Bytes) -> Self {
+		Self { bytes, position: 0 }
+	}
+
+	fn decode_bytes_with_len(&mut self, length: usize) -> Result<bytes::Bytes, Error> {
+		bytes::Buf::advance(&mut self.bytes, self.position);
```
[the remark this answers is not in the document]
serban300: Reverted the change. Not applicable anymore.

serban300:
> I still think requiring to implement it manually can be painful. But if the types requiring this are not many then it good.

Yes, sorry, you're right. We need to be able to derive it since for types like `RuntimeCall` we can't implement it manually. Done.

src/codec.rs
serban300: True. Initially I wanted to avoid the complexity of keeping track of this. I saw that other libraries did the same. For example the `heapsize` lib from `servo`: https://github.com/servo/heapsize/blob/master/src/lib.rs#L295 . But anyway, done. PTAL !

serban300:
> > Did you try this will work on Substrate? I mean you will need to ensure that `Call` implements `DecodeWithMemTracking` and all the types we are using there.
> 
> Oh, you're right. I can patch `polkadot-sdk` locally and try this. Will do it and get back.

Done. This compiles for `rococo-runtime`: https://github.com/serban300/polkadot-sdk/tree/mem-limit-decode . It's very verbose but seems to be implemented for all the types that we need.

> > Also you should expose a function `decode_with_mem_limit` instead of exposing the `MemLimitInput`
> 
> Yes, will do this too.

Done

@bkchr Addressed all the comments. PTAL !

src/codec.rs
serban300: Changed this. Not applicable anymore.

src/codec.rs
```diff
@@ -1223,17 +1261,99 @@ macro_rules! impl_codec_through_iterator {
...
 
-impl_codec_through_iterator! {
+// Constants from rust's source:
+// https://doc.rust-lang.org/src/alloc/collections/btree/node.rs.html#43-45
+const BTREE_B: usize = 6;
+const BTREE_CAPACITY: usize = 2 * BTREE_B - 1;
+const BTREE_MIN_LEN_AFTER_SPLIT: usize = BTREE_B - 1;
+
+/// Estimate the mem size of a btree.
+fn mem_size_of_btree<T>(len: u32) -> usize {
```
bkchr, line 1271: Would be nice to have some sort of test for this, but also not that easy to implement. We would require some custom allocator to get the statistics.

bkchr: But not a requirement for this pr

derive/src/decode.rs
```diff
@@ -300,3 +285,44 @@ fn create_instance(
...
+		}
+
+		let field_type = if let Some(compact) = utils::get_compact_type(field, crate_path) {
+			compact
+		} else if let Some(encoded_as) = utils::get_encoded_as_type(field) {
+			encoded_as
+		} else {
+			field.ty.to_token_stream()
+		};
+		Some(quote! {<#field_type as #crate_path::DecodeWithMemTracking>::__is_implemented();})
```
bkchr, line 322: This `__is_implemented` function should be removed and the checks should be put into the unnamed `const` and using something like this:
https://docs.rs/static_assertions/1.1.0/src/static_assertions/assert_impl.rs.html#57-83

(We should copy this and not import the crate) Generally we could probably also have the following code:
```
fn require_decode_with_mem::<T: DecodeWithMemTracking>() {}

require_decode_with_mem::<field_ty>()
```

This should probably also work.

bkchr (commented):
Some last nitpicks. Otherwise it already looks good :)

derive/src/utils.rs
```diff
@@ -449,3 +467,16 @@ pub fn is_transparent(attrs: &[syn::Attribute]) -> bool {
...
+	let data_variants = || data.variants.iter().filter(|variant| !should_skip(&variant.attrs));
+
+	if data_variants().count() > 256 {
+		return Err(syn::Error::new(
+			data.variants.span(),
+			"Currently only enums with at most 256 variants are encodable/decodable.",
+		))
+	}
+
+	Ok(data_variants().collect())
```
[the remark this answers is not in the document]
serban300: Yes, looks better. Done. Thanks !

src/codec.rs
```diff
@@ -1223,17 +1261,99 @@ macro_rules! impl_codec_through_iterator {
...
+}
+
+impl<T: Decode> Decode for LinkedList<T> {
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
+			input.descend_ref()?;
+			let result = Result::from_iter((0..len).map(|_| {
+				// We account for the size of the `prev` and `next` pointers of each list node,
+				// plus the decoded element.
+				input.on_before_alloc_mem(mem::size_of::<(usize, usize, T)>())?;
```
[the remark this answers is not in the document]
serban300: Good point. Done.

derive/src/decode.rs
serban300: Nice ! Done.

src/codec.rs
serban300: Good point ! Added a simple test and also adjusted `mem_size_of_btree()` a bit.

src/btree_utils.rs
```diff
@@ -0,0 +1,109 @@
...
+	let num_nodes = (len as usize).saturating_div((CAPACITY + MIN_LEN_AFTER_SPLIT) * 2 / 3);
+
+	// If the tree has only one node, it's a leaf node.
+	if num_nodes == 0 {
+		return leaf_node_size;
+	}
+	num_nodes.saturating_mul(internal_node_size)
+}
+
+#[cfg(all(test, feature = "unstable-tests"))]
```
bkchr, line 45:
```suggestion
#[cfg(test)]
#[rustversion::nightly]
```

https://crates.io/crates/rustversion

This way we don't need to expose an feature for this.

serban300: Nice, but also had to add a build script to make it work, as stated here: https://github.com/dtolnay/rustversion/issues/8

PTAL !

derive/src/decode.rs
```diff
@@ -300,3 +285,48 @@ fn create_instance(
...
+		Data::Struct(data) => Box::new(data.fields.iter()),
+		Data::Enum(ref data) => {
+			let variants = match utils::try_get_variants(data) {
+				Ok(variants) => variants,
+				Err(e) => return e.to_compile_error(),
+			};
+
+			let mut fields: Box<dyn Iterator<Item = &Field>> = Box::new(iter::empty());
+			for variant in variants {
+				fields = Box::new(fields.chain(variant.fields.iter()));
```
ggwpez, line 300: Why do you use an iter here instead of a vector?

serban300: It should be more efficient. If we want to build a vector we have to iterate through the fields and `collect()`. This way we just get the iterator and use it below.

ggwpez: Huh, derive macros dont have to be optimized, but okay.

src/compact.rs
```diff
@@ -249,7 +256,12 @@ impl<T> MaybeMaxEncodedLen for T {}
 /// Trait that tells you if a given type can be encoded/decoded in a compact way.
 pub trait HasCompact: Sized {
 	/// The compact type; this can be
-	type Type: for<'a> EncodeAsRef<'a, Self> + Decode + From<Self> + Into<Self> + MaybeMaxEncodedLen;
+	type Type: for<'a> EncodeAsRef<'a, Self>
+		+ Decode
+		+ DecodeWithMemTracking
```
ggwpez, line 261: So only mem tracked types can be compact now? I guess that is fine, since `Compact` should be sealed anyway.

tests/mem_tracking.rs
```diff
@@ -0,0 +1,153 @@
...
+	assert_eq!(&decoded_obj, &obj);
+
+	let raw_input = &mut &encoded_bytes[..];
+	let mut input = MemTrackingInput::new(raw_input, mem_limit);
+	let decoded_obj = T::decode(&mut input)?;
+	assert_eq!(&decoded_obj, &obj);
+	assert_eq!(input.used_mem(), expected_used_mem);
+	if expected_used_mem > 0 {
+		let raw_input = &mut &encoded_bytes[..];
+		let mut input = MemTrackingInput::new(raw_input, expected_used_mem);
```
ggwpez, line 62: Why does it fail with this? I thought that it should fail only with `expected_used_mem - 1`.

ggwpez (approved):

src/compact.rs
serban300: Yes, this was also my reasoning.

serban300:
@bkchr just wanted to check if you would like to take another look on the PR or if I could merge it as it is since it already has 2 approving reviews

src/compact.rs
```diff
@@ -273,7 +285,12 @@ where
 
 impl<T: 'static> HasCompact for T
 where
-	Compact<T>: for<'a> EncodeAsRef<'a, T> + Decode + From<Self> + Into<Self> + MaybeMaxEncodedLen,
+	Compact<T>: for<'a> EncodeAsRef<'a, T>
+		+ Decode
+		+ DecodeWithMemTracking
```
bkchr, line 290: This and the change above are semver breaking and require a major bump.

derive/src/decode.rs
```diff
@@ -300,3 +285,48 @@ fn create_instance(
...
+			compact
+		} else if let Some(encoded_as) = utils::get_encoded_as_type(field) {
+			encoded_as
+		} else {
+			field.ty.to_token_stream()
+		};
+		Some(quote! {#field_type})
+	});
+
+	quote! {
```
bkchr, line 325: `quote_spanned` with the span of each type to get a better error. Maybe also a ui test would be nice

serban300: Done. I hope I understood correctly how `quote_spanned` works.

src/compact.rs
serban300: Hmmm yes, you're right, good catch ! But, I wouldn't be very keen on bumping the major version, since then we would also have to implement https://github.com/paritytech/parity-scale-codec/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22SCALE+codec+4.0%22 in order to do the release, and it would delay this feature. So WDYT about https://github.com/paritytech/parity-scale-codec/pull/616/commits/336c813bec1c27f4b413a536824ad46dc1be0865 as an alternative ? At least temporary until the other work for releasing 4.0 is done ?

serban300: @bkchr sorry for the ping. Can you take a look on the answer above please ?

derive/src/utils.rs
```diff
@@ -85,17 +86,22 @@ pub fn get_encoded_as_type(field: &Field) -> Option<TokenStream> {
 }
 
 /// Look for a `#[codec(compact)]` outer attribute on the given `Field`.
```
bkchr, line 88: Docs should be updated.

Cargo.toml
```diff
@@ -1,24 +1,26 @@
...
 serde = { version = "1.0.204", default-features = false, optional = true }
 parity-scale-codec-derive = { path = "derive", version = ">= 3.6.8", default-features = false, optional = true }
-bitvec = { version = "1", default-features = false, features = [ "alloc" ], optional = true }
+bitvec = { version = "1", default-features = false, features = ["alloc"], optional = true }
 bytes = { version = "1", default-features = false, optional = true }
 byte-slice-cast = { version = "1.2.2", default-features = false }
 generic-array = { version = "0.14.7", optional = true }
 arbitrary = { version = "1.3.2", features = ["derive"], optional = true }
 impl-trait-for-tuples = "0.2.2"
+rustversion = "1.0.17"
```
bkchr, line 23: We don't need it for the normal dependencies?

bkchr (approved):
Nice work :+1:  Thank you for doing this!

derive/src/utils.rs
serban300: Done

Cargo.toml
serban300: You're right. Missed it. Done.

PR #622, opened 2024-08-16
Fix `MaxEncodedLen` derive macro for enum with skipped variant

gui1117:
right now `MaxEncodedLen` do not differentiate betwen skipped variants and normal variants on enum.

Now skipped variant are correctly skipped.

(also thinking again, it is not so much consistent to have skipped variant being not decodable and skipped fields being decoded to default value. Maybe it should be 2 different names)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 # Parity SCALE Codec
 
 Rust implementation of the SCALE (Simple Concatenated Aggregate Little-Endian) data format
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+
 # Parity SCALE Codec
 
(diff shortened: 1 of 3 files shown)


davxy (approved):

shawntabrizi (approved):

gui1117:
CI fails because I am external maybe? or CI is down?

ggwpez (approved):

ggwpez:
@gui1117 i started the CI now. Looks like the format is not happy and the clippy because of unconstructed variants 🙃

gui1117:
> @gui1117 i started the CI now. Looks like the format is not happy and the clippy because of unconstructed variants 🙃

Thanks, it should be fixed hopefully

Issue #507, opened 2023-08-29
Disallow duplicated enum variant indices


nazar-pc:
Currently it is possible to do this:
```
#[derive(Encode, Decode)]
enum E {
    #[codec(index = 0)]
    A,
    #[codec(index = 0)]
    B,
}
```

Logically it seems like an error and should not compile at all. Admittedly this is a minor issue.

gui1117:
related https://github.com/paritytech/parity-scale-codec/issues/173

Issue #173, opened 2019-12-03
Derive encode for variants can result in having multiple variants having same index.


thiolliere:
for instance this example does compile without warning:
```rust
#[test]
fn discriminant_variant_counted_in_default_index() {
	#[derive(Encode)]
	enum T {
		A = 1,
		B,
	}

	assert_eq!(T::A.encode(), vec![1]);
	assert_eq!(T::B.encode(), vec![1]);
}
```

We should make this a compile_error or try to give some warning in some way

thiolliere:
as warning is difficult in proc_macro, an idea would just be compile-error when it happens and adding a new attribute "#[codec(AllowVariantIndexConflict)]"

thiolliere:
implementation is actually difficult because disciminant is an expression: user can write:
```rust
enum A { A = 1+1 };
```

Thus the only solution I see is to implement a tests which test that no index are same

thiolliere:
unfortunately it is not possible to derive some test if the macro is called inside a function like this:
```rust
fn enum_struct_test() {
	#[derive(Encode, Decode, Debug)]
	enum A {
		A = 1,
		B,
		C,
	}
}
```
It result in:
```
warning: cannot test inner items
 --> tests/skip.rs:6:11
  |
6 |     #[derive(Encode, Decode, Debug)]
  |              ^^^^^^
  |
  = note: `#[warn(unnameable_test_items)]` on by default
  = note: this warning originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

gslancaster:
The Rust Reference indicates that the correct discriminant value for `T::B` in the example given in the OP is 2.

https://doc.rust-lang.org/reference/items/enumerations.html#implicit-discriminants

It is unfortunate that the SCALE serialized discriminant is 1.

SCALE is free to decide its own discriminator values for serialized data, I suppose. However, creating ambiguity between two discriminants, where the standard Rust discriminants are unambiguous, seems like an unintentional bug. It can create the situation where you encode one enum value (e.g., `T::B`) but decode another (e.g., `T::A`).

thiolliere:
yes, but it seems difficult to implement like this in scale codec because discriminant is an expression, and we don't have access to the result AFAIK.
Also ideally it would be great to have a way forward without breaking any change.

thiolliere:
We could still implement the correct discriminant by reusing the expression and adding `+1` maybe.
[...]

PR #628, opened 2024-09-11
Disallow duplicated enum variant indices

pkhry:
## Description
- Disallows usage of duplicated variant indices during derivation.
e.g 
```rust
enum Example {
    #[codec(index = 0)]
    A, 
    B = 0,  // Error here
    C,
}
```
- Assigns unique indexes for each enum variant based on heuristic described below: 
  - if codec::index is present -> use it and add the index to the set of duplicates. If there are duplicate indexes -> return an error. 
  - if there is explicit discriminant(e.g. `A = 3`) -> use it and add it to the set of duplicates. If there are any duplicates in the set of used indexes or explicit indexes -> return an error.
  - no discriminant/explicit index -> use an incrementing number starting from 0, if number is in the used set it will get incremented until its no longer in set. 
```rust
enum Example {
    #[codec(index = 0)]
    A, 
    B = 1, 
    C, // Will have index `2`
}

enum Example {
    #[codec(index = 1)]
    A, 
    B = 2, 
    C, // Will have index `0`
}
```

### See also: 
#507


derive/src/decode.rs
```diff
@@ -57,9 +57,17 @@ pub fn quote(
 				.to_compile_error();
 			}
 
-			let recurse = data_variants().enumerate().map(|(i, v)| {
+			let mut used_indexes = match UsedIndexes::from_iter(data_variants()) {
+				Ok(index) => index,
+				Err(e) => return e.into_compile_error(),
+			};
```
bkchr, line 63:
```suggestion
			let mut used_indexes = match UsedIndexes::from_iter(data_variants()).map_err(|e| e.into_compile_error())?;
```

derive/src/utils.rs
```diff
@@ -37,32 +38,90 @@ where
...
-	// first look for an attribute
-	let index = find_meta_item(v.attrs.iter(), |meta| {
-		if let NestedMeta::Meta(Meta::NameValue(ref nv)) = meta {
-			if nv.path.is_ident("index") {
-				if let Lit::Int(ref v) = nv.lit {
-					let byte = v
-						.base10_parse::<u8>()
-						.expect("Internal error, index attribute must have been checked");
-					return Some(byte);
+pub struct UsedIndexes {
```
bkchr, line 41: Missing docs.

Also this struct should only store the `used_set`. The function `from_iter` is not required.

You only need `variant_index`, which should be a verbatim copy of the old `variant_index` function plus that you add the resulting index to `used_set` and check if it already exists.

derive/src/utils.rs
```diff
@@ -37,32 +38,90 @@ where
...
+	used_set: HashSet<u8>,
```
bkchr, line 42: Could be a `HashMap<u8, Span>` to show a proper error message.

pkhry: done, not it's a hashMap. Now when we encounter a duplicate variant index we would also see span of the previous use site

derive/src/utils.rs
pkhry: added docs. 
we need current to assign incremental indexes to variants in ascending order also we need to traverse the variants first and collect all the assigned indexes before we can process them to check for duplicates and have a set of indexes already used, so that when we encounter a variant without explicit discriminant/`codec(index = ?)` we could assign an index using `UsedIndexes.current` by incrementing it until the `u8` is not contained inside the used set.

derive/src/decode.rs
pkhry: done

tests/variant_number.rs
```diff
@@ -36,5 +36,16 @@ fn index_attr_variant_counted_and_reused_in_default_index() {
 	}
 
 	assert_eq!(T::A.encode(), vec![1]);
-	assert_eq!(T::B.encode(), vec![1]);
+	assert_eq!(T::B.encode(), vec![0]);
+}
+#[test]
+fn index_attr_vairant_duplicates_indices() {
```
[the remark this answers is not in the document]
pkhry: added comments, but basically that indexes are correct in the presence of explicit disciminants/ `codec(index =?)`

gui1117:
This looks like a difficult breaking change for users.

They would have to manually check all their usage to see if any index is changing.

Maybe it would be better to have a safe transition.

Also if we are to change the indices, the most idiomatic order could be to follow the rust index implementation maybe.
```
enum Foo
    A = 3,
    B, // is 4
```

derive/src/utils.rs
```diff
@@ -37,32 +38,99 @@ where
...
+				None
+			}) {
+				if let Some(span) = map.insert(index, nv.span()) {
+					let mut error = syn::Error::new(nv.span(), "Duplicate variant index. qed");
+					error.combine(syn::Error::new(span, "Variant index already defined here."));
+					return Err(error)
+				}
+			} else if let Some((
+				_,
+				expr @ syn::Expr::Lit(ExprLit { lit: syn::Lit::Int(lit_int), .. }),
```
gui1117, line 76: I think this doesn't capture any discriminant, user can write this:
```rust
#[derive(Encode, Decode)]
enum A {
	A = 3 + 4,
}
```

I think it can be ok to constraint our implementation, but it should be done in a safe way, maybe requiring a new attribute which specifies legacy implementation or new implementation.

pkhry: updated with to disallow this pattern

pkhry:
> This looks like a difficult breaking change for users.
> 
> They would have to manually check all their usage to see if any index is changing.
> 
> Maybe it would be better to have a safe transition.
> 
> Also if we are to change the indices, the most idiomatic order could be to follow the rust index implementation maybe.
> 
> ```rust
> enum Foo
>     A = 3,
>     B, // is 4
> ```
> 
> This discussion can be relevant: #173

I feel like it's very hard to achieve safe transition, initial impl that i pushed tried to index each variant in a manner that was somewhat compatible with some of the things already existing. 

I also pushed an update that indexes variants in a way that's closer to the rust spec

gui1117:
> > This looks like a difficult breaking change for users.
> > They would have to manually check all their usage to see if any index is changing.
> > Maybe it would be better to have a safe transition.
> > Also if we are to change the indices, the most idiomatic order could be to follow the rust index implementation maybe.
> > ```rust
> > enum Foo
> >     A = 3,
> >     B, // is 4
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This discussion can be relevant: #173
> 
> I feel like it's very hard to achieve safe transition, initial impl that i pushed tried to index each variant in a manner that was somewhat compatible with some of the things already existing.
> 
> I also pushed an update that indexes variants in a way that's closer to the rust spec

I see the commit for the compile error on complex expression, but not the other commit?

Thinking again the index variant closer to rust spec is not so important. I think the most important is to not break user code.

IMO we don't need to change the index, I know the index is not ideal, but any index will feel not so ideal (some people will prefer your initial implementation which is great, some other will prefer to follow rust discriminant order).

So maybe we should just not change the order and instead have a compile error when there is conflict? what do you think?

pkhry:
> > > This looks like a difficult breaking change for users.
> > > They would have to manually check all their usage to see if any index is changing.
> > > Maybe it would be better to have a safe transition.
> > > Also if we are to change the indices, the most idiomatic order could be to follow the rust index implementation maybe.
> > > ```rust
> > > enum Foo
> > >     A = 3,
> > >     B, // is 4
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This discussion can be relevant: #173
> > 
> > 
> > I feel like it's very hard to achieve safe transition, initial impl that i pushed tried to index each variant in a manner that was somewhat compatible with some of the things already existing.
> > I also pushed an update that indexes variants in a way that's closer to the rust spec
> 
> I see the commit for the compile error on complex expression, but not the other commit?
> 
> Thinking again the index variant closer to rust spec is not so important. I think the most important is to not break user code.
> 
> IMO we don't need to change the index, I know the index is not ideal, but any index will feel not so ideal (some people will prefer your initial implementation which is great, some other will prefer to follow rust discriminant order).
> 
> So maybe we should just not change the order and instead have a compile error when there is conflict? what do you think?

- reverted to the indexing strategy used initially, we now just give out a compile error after encountering duplicates.
- added ui tests for failing cases

derive/src/utils.rs
```diff
@@ -38,11 +38,60 @@ where
...
+				None =>
+					if let Some(span) = map.insert(i.try_into().unwrap(), v.span()) {
+						let mut error =
+							syn::Error::new(span, "Custom variant index is duplicated later. qed");
+						error.combine(syn::Error::new(v.span(), "Variant index derived here."));
+						return Err(error)
+					},
+			}
+		}
+	}
```
gui1117, line 86: I think this code is doing same conversion as `variant_index`, we can merge both into:
```rust
/// check usage of variant indexes with #[scale(index = $int)] attribute or
/// explicit  discriminant on the variant
pub fn check_indexes<'a, I: Iterator<Item = &'a &'a Variant>>(values: I) -> syn::Result<()> {
	let mut map: HashMap<u8, Span> = HashMap::new();
	for (i, v) in values.enumerate() {
		let index = variant_index(v, i)?;
		if let Some(span) = map.insert(index, v.span()) {
			let mut error = syn::Error::new(
				v.span(),
				"scale codec error: Invalid variant index, the variant index is duplicated.",
			);
			error.combine(syn::Error::new(span, "Variant index used here."));
			return Err(error)
		}
	}
	Ok(())
}

/// Look for a `#[scale(index = $int)]` attribute on a variant. If no attribute
/// is found, fall back to the discriminant or just the variant index.
pub fn variant_index(v: &Variant, index: usize) -> syn::Result<u8> {
	// first look for an attribute
	let codec_index = find_meta_item(v.attrs.iter(), |meta| {
		if let Meta::NameValue(ref nv) = meta {
			if nv.path.is_ident("index") {
				if let Expr::Lit(ExprLit { lit: Lit::Int(ref v), .. }) = nv.value {
					let byte = v
						.base10_parse::<u8>()
						.expect("Internal error, index attribute must have been checked");
					return Some(byte);
				}
			}
		}

		None
	});
	if let Some(index) = codec_index {
		Ok(index)
	} else {
		match v.discriminant.as_ref() {
			Some((_, syn::Expr::Lit(ExprLit { lit: syn::Lit::Int(v), .. }))) => {
				let byte = v
					.base10_parse::<u8>()
					.expect("scale codec error: Invalid variant index, discriminant doesn't fit u8.");
				Ok(byte)
			},
			Some((_, expr)) => Err(syn::Error::new(
					expr.span(),
					"scale codec error: Invalid discriminant, only int literal are accepted, e.g. \
					`= 32`.",
			)),
			None => index.try_into().map_err(|_| syn::Error::new(
					v.span(),
					"scale codec error: Variant index is too large, only 256 variants are \
					supported.",
			)),
		}
	}
}
```

or you can apply this patch:
```diff
diff --git a/derive/src/utils.rs b/derive/src/utils.rs
index 2812fc6..c803ff7 100644
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -43,45 +43,14 @@ where
 pub fn check_indexes<'a, I: Iterator<Item = &'a &'a Variant>>(values: I) -> syn::Result<()> {
 	let mut map: HashMap<u8, Span> = HashMap::new();
 	for (i, v) in values.enumerate() {
-		if let Some(index) = find_meta_item(v.attrs.iter(), |meta| {
-			if let Meta::NameValue(ref nv) = meta {
-				if nv.path.is_ident("index") {
-					if let Expr::Lit(ExprLit { lit: Lit::Int(ref v), .. }) = nv.value {
-						let byte = v
-							.base10_parse::<u8>()
-							.expect("Internal error, index attribute must have been checked");
-						return Some(byte);
-					}
-				}
-			}
-			None
-		}) {
-			if let Some(span) = map.insert(index, v.span()) {
-				let mut error = syn::Error::new(v.span(), "Duplicate variant index. qed");
-				error.combine(syn::Error::new(span, "Variant index already defined here."));
-				return Err(error)
-			}
-		} else {
-			match v.discriminant.as_ref() {
-				Some((_, syn::Expr::Lit(ExprLit { lit: syn::Lit::Int(lit_int), .. }))) => {
-					let index = lit_int
-						.base10_parse::<u8>()
-						.expect("Internal error, index attribute must have been checked");
-					if let Some(span) = map.insert(index, v.span()) {
-						let mut error = syn::Error::new(v.span(), "Duplicate variant index. qed");
-						error.combine(syn::Error::new(span, "Variant index already defined here."));
-						return Err(error)
-					}
-				},
-				Some((_, _)) => return Err(syn::Error::new(v.span(), "Invalid discriminant. qed")),
-				None =>
-					if let Some(span) = map.insert(i.try_into().unwrap(), v.span()) {
-						let mut error =
-							syn::Error::new(span, "Custom variant index is duplicated later. qed");
-						error.combine(syn::Error::new(v.span(), "Variant index derived here."));
-						return Err(error)
-					},
-			}
+		let index = variant_index(v, i)?;
+		if let Some(span) = map.insert(index, v.span()) {
+			let mut error = syn::Error::new(
+				v.span(),
+				"scale codec error: Invalid variant index, the variant index is duplicated.",
+			);
+			error.combine(syn::Error::new(span, "Variant index used here."));
+			return Err(error)
 		}
 	}
 	Ok(())
@@ -89,7 +58,7 @@ pub fn check_indexes<'a, I: Iterator<Item = &'a &'a Variant>>(values: I) -> syn:
 
 /// Look for a `#[scale(index = $int)]` attribute on a variant. If no attribute
 /// is found, fall back to the discriminant or just the variant index.
-pub fn variant_index(v: &Variant, index: usize) -> syn::Result<TokenStream> {
+pub fn variant_index(v: &Variant, index: usize) -> syn::Result<u8> {
 	// first look for an attribute
 	let codec_index = find_meta_item(v.attrs.iter(), |meta| {
 		if let Meta::NameValue(ref nv) = meta {
@@ -106,13 +75,25 @@ pub fn variant_index(v: &Variant, index: usize) -> syn::Result<TokenStream> {
 		None
 	});
 	if let Some(index) = codec_index {
-		Ok(quote! { #index })
+		Ok(index)
 	} else {
 		match v.discriminant.as_ref() {
-			Some((_, expr @ syn::Expr::Lit(ExprLit { lit: syn::Lit::Int(_), .. }))) =>
-				Ok(quote! { #expr }),
-			Some((_, expr)) => Err(syn::Error::new(expr.span(), "Invalid discriminant. qed")),
-			None => Ok(quote! { #index }),
+			Some((_, syn::Expr::Lit(ExprLit { lit: syn::Lit::Int(v), .. }))) => {
+				let byte = v
+					.base10_parse::<u8>()
+					.expect("scale codec error: Invalid variant index, discriminant doesn't fit u8.");
+				Ok(byte)
+			},
+			Some((_, expr)) => Err(syn::Error::new(
+					expr.span(),
+					"scale codec error: Invalid discriminant, only int literal are accepted, e.g. \
+					`= 32`.",
+			)),
+			None => index.try_into().map_err(|_| syn::Error::new(
+					v.span(),
+					"scale codec error: Variant index is too large, only 256 variants are \
+					supported.",
+			)),
 		}
 	}
 }
```

gui1117 (commented):
PR is good, some refactor and then I will approve.

gui1117:
Also I just found out we can assert that variant indices are unique using constant evaluation.
No need to enforce that discriminant is a literal (so no breaking change at all)

A draft implementation look like this:
```diff
diff --git a/derive/src/encode.rs b/derive/src/encode.rs
index a29eec7..798237f 100644
--- a/derive/src/encode.rs
+++ b/derive/src/encode.rs
@@ -345,7 +345,7 @@ fn impl_encode(data: &Data, type_name: &Ident, crate_path: &syn::Path) -> TokenS
                                                        }
                                                };
 
-                                               [hinting, encoding]
+                                               [hinting, encoding, index]
                                        },
                                        Fields::Unnamed(ref fields) => {
                                                let fields = &fields.unnamed;
@@ -378,7 +378,7 @@ fn impl_encode(data: &Data, type_name: &Ident, crate_path: &syn::Path) -> TokenS
                                                        }
                                                };
 
-                                               [hinting, encoding]
+                                               [hinting, encoding, index]
                                        },
                                        Fields::Unit => {
                                                let hinting = quote_spanned! { f.span() =>
@@ -394,13 +394,14 @@ fn impl_encode(data: &Data, type_name: &Ident, crate_path: &syn::Path) -> TokenS
                                                        }
                                                };
 
-                                               [hinting, encoding]
+                                               [hinting, encoding, index]
                                        },
                                }
                        });
 
-                       let recurse_hinting = recurse.clone().map(|[hinting, _]| hinting);
-                       let recurse_encoding = recurse.clone().map(|[_, encoding]| encoding);
+                       let recurse_hinting = recurse.clone().map(|[hinting, _, _]| hinting);
+                       let recurse_encoding = recurse.clone().map(|[_, encoding, _]| encoding);
+                       let recurse_indices = recurse.clone().map(|[_, _, index]| index);
 
                        let hinting = quote! {
                                // The variant index uses 1 byte.
@@ -411,6 +412,23 @@ fn impl_encode(data: &Data, type_name: &Ident, crate_path: &syn::Path) -> TokenS
                        };
 
                        let encoding = quote! {
+                                       const _: () = {
+                                               let indices = [#( #recurse_indices ,)*];
+                                               let len = indices.len();
+
+                                               // Check each pair for uniqueness
+                                               let mut i = 0;
+                                               while i < len {
+                                                       let mut j = i + 1;
+                                                       while j < len {
+                                                               if indices[i] == indices[j] {
+                                                                       panic!("TODO: good error message with variant names and indices");
+                                                               }
+                                                               j += 1;
+                                                       }
+                                                       i += 1;
+                                               }
+                                       };
                                match *#self_ {
                                        #( #recurse_encoding )*,
                                        _ => (),
```

derive/src/utils.rs
pkhry: applied

gui1117:
I'm good with the current PR.
I just didn't know it was possible with const evaluation.

If we go the const evaluation way, we should do for both decode and encode. There won't be overhead as it is const evaluation. The error will be a bit less pretty but probably ok.

Anyway we can go with the PR as it is.

gui1117:
don't hate me, but I am conflicted, the current PR is good but it is a breaking change. We no longer support enums with discriminant being an expression rather than a literal.

And there to achieve the goal of this PR we can do it without a breaking change, by using constant evaluation.

People regularly complains that libraries are breaking too often, maybe we should avoid this breaking change and implement with the constant evaluation.

pkhry:
> don't hate me, but I am conflicted, the current PR is good but it is a breaking change. We no longer support enums with discriminant being an expression rather than a literal.
> 
> And there to achieve the goal of this PR we can do it without a breaking change, by using constant evaluation.
> 
> People regularly complains that libraries are breaking too often, maybe we should avoid this breaking change and implement with the constant evaluation.

All good, I'll re-do it

pkhry:
> don't hate me, but I am conflicted, the current PR is good but it is a breaking change. We no longer support enums with discriminant being an expression rather than a literal.
> 
> And there to achieve the goal of this PR we can do it without a breaking change, by using constant evaluation.
> 
> People regularly complains that libraries are breaking too often, maybe we should avoid this breaking change and implement with the constant evaluation.



> don't hate me, but I am conflicted, the current PR is good but it is a breaking change. We no longer support enums with discriminant being an expression rather than a literal.
> 
> And there to achieve the goal of this PR we can do it without a breaking change, by using constant evaluation.
> 
> People regularly complains that libraries are breaking too often, maybe we should avoid this breaking change and implement with the constant evaluation.

Ok, it seems like string interpolation is not possible without external dependency inside constant blocks, see alternate [PR's](https://github.com/paritytech/parity-scale-codec/pull/653/files) resulting error message. the only way to have string interpolation in constant blocks is to use `const_format` from what i've understood.

PR #630, opened 2024-09-22
Introduce `CountedInput`. a wrapper which counts the number of bytes successfully read

gui1117:
I saw some code which tries to track the number of byte read with `remaining_len`. But `remaining_len` don't always give information.

I think it is better to provide an accurate counter.

I used u64 because it made the code simpler. And more general purpose.

Otherwise I was thinking using a type:
```rust
enum Count {
    Exact(u32),
    MaxReached,
}
```
and do the count in u32.

But most machine being 64bit, I think it should execute good.

### DIFF
--- a/src/counted_input.rs
+++ b/src/counted_input.rs
@@ -137,7 +137,7 @@ mod test {
 
 		counted_input.counter = max_exact_count;
 
-		assert_eq!(counted_input.count(), max_exact_count));
+		assert_eq!(counted_input.count(), max_exact_count);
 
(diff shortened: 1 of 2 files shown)


bkchr (approved):

shawntabrizi (approved):

shawntabrizi:
fyi we also have `saturated_inc()` which mutates in place and adds one.

bkchr:
> fyi we also have `saturated_inc()` which mutates in place and adds one.

No in the standard library (which is only used in this crate)

ggwpez (approved):

PR #640, opened 2024-10-23
chore(deps): Update to syn 2

jsdw:
Note: `impl-trait-for-tuples` still uses syn 1, so need to update that to fully remove syn 1 from the tree


ggwpez (approved):
🙏

niklasad1 (approved):

jsdw:
PR in impl-trait-for-tuples here https://github.com/bkchr/impl-trait-for-tuples/pull/12

PR #641, opened 2024-10-23
Feature/refactor check repr

fewForce:
This PR :  

1. Resolved conflicts arising from upgradation of dependency  syn 1.98 to syn2
2. Added find_meta_item_in_path function to enhance utils.rs by reducing code redundancy. The new implementation uses Rust's functional programming features for improved readability & conciseness

### DIFF
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -37,6 +37,19 @@ where
 		attr.path().is_ident("codec").then(|| pred(attr.parse_args().ok()?)).flatten()
 	})
 }
+fn find_meta_item_in_path<F, R>(attrs: &[Attribute], word: &str, success_callback: F) -> Option<R>
+where
+	F: FnMut(&Path) -> R + Clone, // Ensure that F implements Clone
+{
+	find_meta_item(attrs.iter(), move |meta| {
+		if let Meta::Path(ref path) = meta {
+			if path.is_ident(word) {
+				return Some(success_callback.clone()(path));
+			}
+		}
+		None
+	})
+}
 
 /// Look for a `#[scale(index = $int)]` attribute on a variant. If no attribute
 /// is found, fall back to the discriminant or just the variant index.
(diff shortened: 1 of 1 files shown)


derive/src/utils.rs
```diff
@@ -37,6 +37,19 @@ where
 		attr.path().is_ident("codec").then(|| pred(attr.parse_args().ok()?)).flatten()
 	})
 }
+fn find_meta_item_in_path<F, R>(attrs: &[Attribute], word: &str, success_callback: F) -> Option<R>
```
gui1117, line 40: Personally I would name it `find_path_meta_item`.
I am not sure what `in_path` means in this context
```suggestion

/// Execute the callback for all the attribute matching `#[codec($path)]`.
fn find_path_meta_item<F, R>(attrs: &[Attribute], path: &str, success_callback: F) -> Option<R>
```

gui1117 (commented):
what is it fixing actually?

fewForce:
> what is it fixing actually?

@gui1117 
It addresses **Code Duplication** 
It reduces the lines of code.

**Explanation** 
Following code is used at 3 different places in the same file with following 2 changes - 

- `skip` / `dumb_trait_bound` is used instead of `compact`
- Completion behaviour is parameterized using Success callback

```
find_meta_item(field.attrs.iter(), |meta| {
		if let Meta::Path(ref path) = meta {
			if path.is_ident("compact") {
				let field_type = &field.ty;
				return Some(quote! {<#field_type as #crate_path::HasCompact>::Type});
			}
		}
```

gui1117:
If it is only for fixing the duplication of code, I personally think the new code is a bit less readable.

If some other people want this change I can re-open the PR.

fewForce:
> If it is only for fixing the duplication of code, I personally think the new code is a bit less readable.
> 
> If some other people want this change I can re-open the PR.

@gui1117 

Code Readability depends upon the experience of a Developer.
Experienced programmers would say that code should be concise & clear.
To develop for Polkadot’s substrate codebase is not for the light of heart.
Also, I believe removing inadequacies ( how so ever minor ) in code should be promoted by experienced engineers like yourself.

Code Duplication is a tangible metric. It makes code not  look ugly but feels more  like novice’s job.
So many people have seen this code and would notice the duplication.
Rather than trying to go through the process of making a PR, they choose to accept the lower quality code.
Now, waiting for others to say something is probably condoning bad coding practices by parity.

gui1117:
I also think the usage of FnMut + Clone feels not so good.

I actually appreciate the effort you put into making a PR, thank you for having taking the time.

That said I don't think the PR itself is worth merging. (Every review takes time)

If you want to contribute I think there is issue to do in polkadot-sdk that can be great to fix.

fewForce:
> I also think the usage of FnMut + Clone feels not so good.
> 
> I actually appreciate the effort you put into making a PR, thank you for having taking the time.
> 
> That said I don't think the PR itself is worth merging. (Every review takes time)
> 
> If you want to contribute I think there is issue to do in polkadot-sdk that can be great to fix.

@gui1117 

I believe there is no minimum bar for a PR that can be set to stop good code from merging.
- This is an open source project.
- I have followed all the guidelines mentioned in the repository. 
- Personal bias about code readability is not a criterion that can be applied. 

Please let me know if there is a process to appeal or ask for a different reviewer with little bit more experienced & open mind who can understand more complex code.

PR #646, opened 2024-10-24
Prep to release 3.7.0

jsdw:
I consistified some bits via workspace version/repository etc so that both crates align on 3.7.0

### Added

- Allow decoding with a memory limit. ([616](https://github.com/paritytech/parity-scale-codec/pull/616))
- Introduce `CountedInput`, an wrapper on `Input` that counts the bytes read. ([630](https://github.com/paritytech/parity-scale-codec/pull/630))

### Changed

- This release bumps some dependencies, primarily bumping `syn` to 2. ([#640](https://github.com/paritytech/parity-scale-codec/pull/640)).

### Fixed

- Fix MaxEncodedLen derive macro for enum with skipped variant ([#622](https://github.com/paritytech/parity-scale-codec/pull/622))
- Use MAX_PREALLOCATION consistently [#605](https://github.com/paritytech/parity-scale-codec/pull/605)


Cargo.toml
```diff
@@ -1,19 +1,19 @@
...
+edition.workspace = true
 build = "build.rs"
-rust-version = "1.60.0"
+rust-version.workspace = true
 
 [dependencies]
 arrayvec = { version = "0.7", default-features = false }
 serde = { version = "1.0.213", default-features = false, optional = true }
-parity-scale-codec-derive = { path = "derive", version = ">= 3.6.8", default-features = false, optional = true }
+parity-scale-codec-derive = { path = "derive", version = "3.6.8", default-features = false, optional = true }
```
jsdw, line 16: >= is the same as not having it at all

niklasad1 (approved):

derive/Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
 name = "parity-scale-codec-derive"
 description = "Serialization and deserialization derive macro for Parity SCALE Codec"
-version = "3.6.8"
-authors = ["Parity Technologies <admin@parity.io>"]
-license = "Apache-2.0"
-edition = "2021"
-rust-version = "1.56.1"
```
ggwpez, line 8: We do also bump the min rust version here, not sure if that is considered a breaking change? Maybe stay with the old version - if possible.

ggwpez (approved):
% the Rust version comment

derive/Cargo.toml
jsdw: Out of interest, I used `cargo msrv` to find the actual minimum supported Rust version, and it's currently actually 1.79 owing to a dependency using more modern features.

If we care about honouring a specific MSRV we should probably add `cargo msrv` to CI as eg a daily check that we are compliant, but my view is not to worry too much because any dependency that increases its MSRV in a non breaking release will force it on us all of a sudden anyway, unless we pin everything!

Anyway, in this PR I'll set it to 1.79 to align with reality!

CHANGELOG.md
```diff
@@ -5,6 +5,21 @@ All notable changes to this crate are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this crate adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.7.0] - 2024-10-24
+
+### Added
+
+- Allow decoding with a memory limit. ([616](https://github.com/paritytech/parity-scale-codec/pull/616))
```
gui1117, line 12: - Introduce `CountedInput`, an wrapper on `Input` that counts the byte read. ([630](https://github.com/paritytech/parity-scale-codec/pull/630))

CHANGELOG.md
```diff
@@ -5,6 +5,21 @@ All notable changes to this crate are documented in this file.
...
+
```
gui1117, line 13:
```suggestion
- Introduce `CountedInput`, an wrapper on `Input` that counts the bytes read. ([630](https://github.com/paritytech/parity-scale-codec/pull/630))

```

gui1117 (approved):

derive/Cargo.toml
ggwpez: Yea thats fine. I thought that we already checked it in CI, but apparently not 😆

rnbguy: Just reporting that this breaks dependency requirement for cargo.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

```
3.6  :=  >=3.6.0, <4.0.0
```

jsdw: > Just reporting that this breaks dependency requirement for cargo.

Sorry; what do you mean by this? What specifically breaks?

rnbguy: in our crates, we were using rust-version 1.75 with parity-scale-codec 3.6.12. after 3.7.0 release, our msrv builds are failing.

you should have released 4.0.0.

jsdw: Ah I see! 

It's up to the crate to determine whether MSRV breaks are indeed a breaking semver change or not. The general consensus in the Rust community is slightly torn but leans towards MSRV bumps _not_ requiring a major SemVer bump in crates.

In the case of this crate, it turned out that the actual Rust version supported prior to this release was already much higher than the reported MSRV (owing to a dependency havign a higher MSRV than we did). So by bumping the MSRV in this crate we're really just reflecting reality better!

In general, MSRV changes happen a bunch in minor and patch releases in transitive dependencies, making it very difficult to guarantee that this crate conforms to any particular MSRV short of pinning our dependency versions. This makes it very difficult to guarantee anything re MSRV in general.

PR #653, opened 2024-11-08
Disallow duplicate indexes using constant evaluation

pkhry:
### Description 

following a suggestion from here https://github.com/paritytech/parity-scale-codec/pull/628#issuecomment-2453814878

### Notes 

It seems that it's not possible to have a good error message this way, as `format!` in cant be used in constant_blocks, so string interpolation is not possible unless we pick up `const_format` and re-export it from the library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,6 +183,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "const_format"
+version = "0.2.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "126f97965c8ad46d6d9163268ff28432e8f6a1196a55578867832e3049df63dd"
+dependencies = [
+ "const_format_proc_macros",
+]
+
+[[package]]
+name = "const_format_proc_macros"
+version = "0.2.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


gui1117 (approved):

gui1117:
I was basing my PR on top of yours, so I re-opened it.

@bkchr if you like it we can merge it

bkchr (approved):

ggwpez:
This was actually a major change that we released as a patch... it broke the docs build here https://docs.rs/crate/polkadot-sdk/0.12.1/builds/1788762  
cc @gui1117 @nkpar

ggwpez:
Since this is arguably a breaking *fix*, do you think patch is still fine?

bkchr:
> Since this is arguably a breaking _fix_, do you think patch is still fine?

Yes, before this invalid code was emitted.

gui1117:
A fix should have been pushed for the creates in 2412, 2409 and 2407, maybe polkadot-sdk 0.12.1 is an earlier version? or some crates have been missed.

Because we released and yank versions, sorry for that, the automatic changelog missed, I manually added it.

serban300:
Since there hasn't been any error report from any customer could we just fix the docs and keep this release ?

Sorry, I thought this change wasn't major since it doesn't change any interface or structure. It just adds a check for duplicate indexes, which should prevent misusing the library.

Issue #662, opened 2024-12-01
Non-semver breaking change in v3.7.0


prestwich:
Unsure what it was exactly, but `ruint` no longer compiles on 3.7.0

CI output here:
https://github.com/recmo/uint/actions/runs/12097285631/job/33732497498?pr=407

bkchr:
Hey @prestwich, see https://github.com/paritytech/parity-scale-codec/issues/656 for the same discussion.

prestwich:
Please check the CI logs above. This issue is not due to the MSRV change, but due to some sort of bounds change on either an assoc type requirement or on a blanket impl

prestwich:
if you want to reproduce to assure yourself it's not MSRV related, you can compile on 1.80 as follows

```
cd ruint
cargo +1.80 update -p parity-scale-codec --precise 3.6.12 
cargo +1.80 c --features parity-scale-codec
cargo +1.80 update -p parity-scale-codec --precise 3.7.0
cargo +1.80 c --features parity-scale-codec
```

bkchr:
Ahh sorry!

This is again this: https://github.com/recmo/uint/pull/313

We forgot to revert the following changes on master: https://github.com/paritytech/parity-scale-codec/pull/545 :see_no_evil: 

@jsdw we should revert these changes on master, yank and release `3.7.1`. Could you do this? Then at some point we should finally tackle `4.0.0`.. :see_no_evil:

jsdw:
I don't 100% follow the issue right now (just packing for flight etc so don't have time to look much): how about I yank 3.7.0 now and then can look to fix the rest and release 3.7.1 later?

jsdw:
Ah I see, some PR to revert stuff was on a release branch instead of on master, so just need to apply that to master and we're good right? Maybe I have time lemme see

jsdw:
Ok, https://github.com/paritytech/parity-scale-codec/pull/663 opened which was my attempt at reverting said PRs on master and resolving the various conflicts, but I need to look over it in more detail to be confident it does the right thing and need to get packing etc so I'll have a look later and release 3.7.1  when thats' sorted out and I properly understand it

StackOverflowExcept1on:
@jsdw can you also yank scale-codec-derive plz? it failed our CI: https://github.com/gear-tech/gear/actions/runs/12124067535/job/33800965532?pr=4340#step:14:1197
Cargo.lock contains 3.7.0 codec-derive and 3.6.12 codec
[...]

PR #663, opened 2024-12-02
Revert #512 and #508

jsdw:
Quick attempt to revert these PRs mentioned in https://github.com/paritytech/parity-scale-codec/issues/662#issuecomment-2510797809

### DIFF
--- a/derive/src/max_encoded_len.rs
+++ b/derive/src/max_encoded_len.rs
@@ -43,7 +43,7 @@ pub fn derive_max_encoded_len(input: proc_macro::TokenStream) -> proc_macro::Tok
 		parse_quote!(#crate_path::MaxEncodedLen),
 		None,
 		has_dumb_trait_bound(&input.attrs),
-		&crate_path,
+		&crate_path
 	) {
 		return e.to_compile_error().into();
 	}
--- a/derive/src/max_encoded_len.rs
+++ b/derive/src/max_encoded_len.rs
@@ -86,9 +86,7 @@ fn fields_length_expr(fields: &Fields, crate_path: &syn::Path) -> proc_macro2::T
 		let ty = &field.ty;
 		if utils::is_compact(field) {
 			quote_spanned! {
-				ty.span() => .saturating_add(
-					<<#ty as #crate_path::HasCompact>::Type as #crate_path::MaxEncodedLen>::max_encoded_len()
-				)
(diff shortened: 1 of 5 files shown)


bkchr (approved):

jsdw:
For my own benefit: I realised just now that the recent releases prior to 3.7.0 all came off the 3.6.x branch, which had the above PRs reverted also. I totally missed this branch when doing the latest release, assuming everything was released off master!

I think it would be good to get master back to a sane point. Maybe we should not try to merge this PR at all (which we would do just to get 3.7.1 out, and then want to undo anyways) and instead release 4.0.0 directly and deal with the fallout of this. Does that sound fair?

serban300:
> For my own benefit: I realised just now that the recent releases prior to 3.7.0 all came off the 3.6.x branch, which had the above PRs reverted also. I totally missed this branch when doing the latest release, assuming everything was released off master!
> 
> I think it would be good to get master back to a sane point. Maybe we should not try to merge this PR at all (which we would do just to get 3.7.1 out, and then want to undo anyways for 4) and instead just release 4.0.0 directly now that we've yanked 3.7.0 and deal with the fallout of this major bump. Does that sound fair?

@jsdw what you're saying makes sense, but personally I would need a 3.7.1 release soon-ish in order to access the functionality introduced in https://github.com/paritytech/parity-scale-codec/pull/616 and I think we would have to wait quite a while for a 4.0 release

So could we have a v3.7.x branch for now that would include this PR, and then release 3.7.1 from it ? I can do it if you want. Just please let me know how you prefer

bkchr:
Let's merge this pr and add it to the `4.0` milestone (so we can take care of it in the future :P)

derive/src/max_encoded_len.rs
```diff
@@ -86,9 +86,7 @@ fn fields_length_expr(fields: &Fields, crate_path: &syn::Path) -> proc_macro2::T
 		let ty = &field.ty;
 		if utils::is_compact(field) {
 			quote_spanned! {
-				ty.span() => .saturating_add(
-					<<#ty as #crate_path::HasCompact>::Type as #crate_path::MaxEncodedLen>::max_encoded_len()
-				)
+				ty.span() => .saturating_add(<#crate_path::Compact::<#ty> as #crate_path::MaxEncodedLen>::max_encoded_len())
```
gui1117, line 89: The code before was actually: `ty.span() => .saturating_add(<#ty>::max_encoded_len())`

PR #681, opened 2025-01-16
Improve error message

gui1117:
@pkhry I tried to improve the error message, I just print all variant and there index, if there is a formula it will just be the formula expression without the evaluation but I still think it can be enough information for debugging, WDYT?

### DIFF
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -39,11 +39,26 @@ where
 }
 
 pub fn const_eval_check_variant_indexes(
-	recurse_indices: impl Iterator<Item = TokenStream>,
+	recurse_variant_indices: impl Iterator<Item = (syn::Ident, TokenStream)>,
 ) -> TokenStream {
+	let mut recurse_indices = vec![];
+	let mut variant_msg = vec![];
+	for (ident, index) in recurse_variant_indices {
+		variant_msg.push(format!("\tIndex for variant `{}` is `{}`\n", ident, index));
+		recurse_indices.push(index);
+	}
+	let len = recurse_indices.len();
+
+	let error_message = format!(
+		"Found Variants that have duplicate indexes. Use different indexes for each variant.\n\
+		List of variants:\n\
+		{}",
+		variant_msg.concat()
+	);
+
 	quote! {
 		const _: () = {
-			let indices = [#( #recurse_indices ,)*];
+			let indices: [usize; #len] = [#( #recurse_indices ,)*];
 			let len = indices.len();
 
 			// Check each pair for uniqueness
@@ -52,7 +67,7 @@ pub fn const_eval_check_variant_indexes(
 					let mut j = i + 1;
 					while j < len {
 							if indices[i] == indices[j] {
-									::core::panic!("Found Variants that have duplicate indexes. Use different indexes for each variant");
+									::core::panic!(#error_message);
 							}
 							j += 1;
 					}
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -40,37 +40,54 @@ where
 
 pub fn const_eval_check_variant_indexes(
 	recurse_variant_indices: impl Iterator<Item = (syn::Ident, TokenStream)>,
+	crate_path: &syn::Path,
 ) -> TokenStream {
 	let mut recurse_indices = vec![];
-	let mut variant_msg = vec![];
 	for (ident, index) in recurse_variant_indices {
-		recurse_indices.push((index, ident));
+		let ident_str = ident.to_string();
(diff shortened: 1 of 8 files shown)


pkhry (approved):
Lgtm

derive/src/utils.rs
```diff
@@ -39,11 +39,26 @@ where
...
 
 pub fn const_eval_check_variant_indexes(
-	recurse_indices: impl Iterator<Item = TokenStream>,
+	recurse_variant_indices: impl Iterator<Item = (syn::Ident, TokenStream)>,
 ) -> TokenStream {
+	let mut recurse_indices = vec![];
+	let mut variant_msg = vec![];
+	for (ident, index) in recurse_variant_indices {
+		variant_msg.push(format!("\tIndex for variant `{}` is `{}`\n", ident, index));
+		recurse_indices.push(index);
```
bkchr, line 48:
```suggestion
		recurse_indices.push((index, ident));
```

derive/src/utils.rs
```diff
@@ -39,11 +39,26 @@ where
 }
 
 pub fn const_eval_check_variant_indexes(
-	recurse_indices: impl Iterator<Item = TokenStream>,
+	recurse_variant_indices: impl Iterator<Item = (syn::Ident, TokenStream)>,
 ) -> TokenStream {
+	let mut recurse_indices = vec![];
+	let mut variant_msg = vec![];
+	for (ident, index) in recurse_variant_indices {
+		variant_msg.push(format!("\tIndex for variant `{}` is `{}`\n", ident, index));
```
bkchr, line 47:
```suggestion
```

derive/src/utils.rs
```diff
@@ -39,11 +39,26 @@ where
...
+		"Found Variants that have duplicate indexes. Use different indexes for each variant.\n\
+		List of variants:\n\
+		{}",
+		variant_msg.concat()
+	);
+
 	quote! {
 		const _: () = {
-			let indices = [#( #recurse_indices ,)*];
+			let indices: [usize; #len] = [#( #recurse_indices ,)*];
```
bkchr, line 61:
```suggestion
			let indices: [(usize, &'static str); #len] = [#( #recurse_indices ,)*];
```

derive/src/utils.rs
```diff
@@ -52,7 +67,7 @@ pub fn const_eval_check_variant_indexes(
 					let mut j = i + 1;
 					while j < len {
 							if indices[i] == indices[j] {
-									::core::panic!("Found Variants that have duplicate indexes. Use different indexes for each variant");
+									::core::panic!(#error_message);
```
bkchr, line 70: And then we can generate an even better error message here :P

gui1117: I thought it was impossible, but it is. Done in https://github.com/paritytech/parity-scale-codec/pull/681/commits/876a14ab5d93581f7f9b53e70f624ce4e73dbd24

Also are we ok adding a dependency to `const_format`, if not I can always write the concatenation of strings inlined.
(convertion to array, then calculate the new length in const, then instantiate the new array and copy content + plus special code to format u8).

derive/src/utils.rs
```diff
@@ -39,24 +39,55 @@ where
...
+					"Found variants that have duplicate indexes. Both `",
+					indices[DUP_INFO.1].1,
+					"` and `",
+					indices[DUP_INFO.2].1,
+					"` have the index `",
+					indices[DUP_INFO.1].0,
+					"`. Use different indexes for each variant."
+				);
+
+				::core::panic!("{}", msg);
```
bkchr, line 90:
```suggestion
				::core::panic!("Found variants that have duplicate indexes. Both `{}` and `{}` have index `{}`. Each variant is required to have an unique index.",
					indices[DUP_INFO.1].1,
					indices[DUP_INFO.2].1,
					indices[DUP_INFO.1].0,
				);
```

DOesn't that work? :D

gui1117: I wish constant evaluation gets a bit more love from rustc developers.

But formatting macro doesn't actually work at compile-time:
```
		const _: () = {
		    const FOO: &str = "a";
		    const BAR: &str = "b";
			panic!("{}{}", FOO, BAR);
		};
```
```
error[E0015]: cannot call non-const formatting macro in constants
 --> src/lib.rs:4:12
  |
4 |             panic!("{}{}", FOO, BAR);
  |                     ^^
  |
  = note: calls in constants are limited to constant functions, tuple structs and tuple variants
  = note: this error originates in the macro `$crate::const_format_args` which comes from the expansion of the macro `panic` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Issue #691, opened 2025-01-30
Issue with bounds


gui1117:
compiling polkadot-sdk with the latest version fails with:

```
error[E0277]: the trait bound `parity_scale_codec::Compact<Balance>: MaxEncodedLen` is not satisfied
   --> substrate/primitives/staking/src/lib.rs:451:2
    |
451 |     MaxEncodedLen,
    |     ^^^^^^^^^^^^^ the trait `MaxEncodedLen` is not implemented for `parity_scale_codec::Compact<Balance>`

```

gui1117:
So the PR 508 make use of compact type for max encoded len,
Then PR 512 also improve on the code.
Then PR 616 adds some bound for compact type.
Then the revert of 508 and 512 doesn't actually completely revert, and still make use of the compact type for max encoded len.

So now staking doesn't compile because `Compact<Balance>: MaxEncodedLen` is not bound.

solution:
* revert 508 completly
* try to do some fix, but they might introduce some breaking change:

```diff
diff --git a/derive/src/max_encoded_len.rs b/derive/src/max_encoded_len.rs
index 8a592e1..f09f6c8 100644
--- a/derive/src/max_encoded_len.rs
+++ b/derive/src/max_encoded_len.rs
@@ -44,7 +44,7 @@ pub fn derive_max_encoded_len(input: proc_macro::TokenStream) -> proc_macro::Tok
                None,
                has_dumb_trait_bound(&input.attrs),
                &crate_path,
-               false,
+               true,
        ) {
                return e.to_compile_error().into();
        }
@@ -87,7 +87,9 @@ fn fields_length_expr(fields: &Fields, crate_path: &syn::Path) -> proc_macro2::T
                let ty = &field.ty;
                if utils::is_compact(field) {
                        quote_spanned! {
-                               ty.span() => .saturating_add(<#crate_path::Compact::<#ty> as #crate_path::MaxEncodedLen>::max_encoded_len())
+                               ty.span() => .saturating_add(
+                                       <<#ty as #crate_path::HasCompact>::Type as #crate_path::MaxEncodedLen>::max_encoded_len()
+                               )
                        }
                } else {
                        quote_spanned! {
```

gui1117:
I don't have much time to work on it for now, I can start again tomorrow.

serban300:
@gui1117 I can take it

serban300:
Now I see that there are multiple issues left even after fixing this one:

1. 
```
[...]

PR #545, opened 2023-11-29
Revert breaking changes, bump version to `3.6.9`

ascjones:
Reverts #512 and #508, which are breaking changes, and bumps version to `3.6.9` for release.

PR #693, opened 2025-01-30
Follow-up on #663

serban300:
Fixes https://github.com/paritytech/parity-scale-codec/issues/691

Follow-up on #663 
Reverting one more thing as per https://github.com/paritytech/parity-scale-codec/pull/545


gui1117 (commented):
This code is now the same 3.6.9
So one less breaking change.

gui1117 (approved):

PR #687, opened 2025-01-27
Allow different reprs to still work

jsdw:
Don't output suffixes when generating the index array in the const fun, to avoid the suffix types from conflicting with the declared type of the index array (usize) which leads to compile errors when indexes are from discriminants and are not usizes.

Should fix #686

PR #695, opened 2025-01-31
Fix breaking change: Do not enforce discriminant to be a literal + don't expect usize.

gui1117:
Fix breaking change: https://github.com/paritytech/parity-scale-codec/pull/687
Fix forcing usize for variant expression introduced in: https://github.com/paritytech/parity-scale-codec/pull/653

Currently testing with polkadot-sdk: https://github.com/paritytech/polkadot-sdk/pull/7417

Maybe we can have a new attribute: allow duplicate index.


tests/mod.rs
```diff
@@ -931,3 +931,13 @@ fn deserializing_of_big_recursively_nested_enum_works() {
 	let obj_d2 = Enum::decode_with_depth_limit(40, &mut &data[..]).unwrap();
 	assert!(obj == obj_d2);
 }
+
+#[test]
+fn non_literal_variant_discriminant() {
+	const A: isize = 1;
```
gui1117, line 937: To be honest I don't know why rust requires a isize here.

Also writing:
```
enum A {
  B = -1,
  C = 0,
}
```

seems to give the discriminant 0 to both variants.

derive/src/utils.rs
```diff
@@ -112,21 +142,13 @@ pub fn variant_index(v: &Variant, i: usize) -> TokenStream {
...
-	// output the index with no suffix (ie 1 rather than 1usize) so that these can be quoted into
-	// an array of usizes and will work regardless of what type the discriminant values actually
-	// are.
-	let s = proc_macro2::Literal::usize_unsuffixed(index);
-	quote! { #s }
+	// then fallback to discriminant or just index
+	index.map(|i| quote! { #i }).unwrap_or_else(|| {
+		v.discriminant
+			.as_ref()
+			.map(|(_, expr)| quote! { (#expr) as isize})
```
gui1117, line 149: I wrote this so it compiles, but it is probably wrong. If the enum is `repr(u64)` it doesn't fit into isize

gui1117:
clippy fails with:
```
Run cargo +stable clippy --locked -- -Dwarnings
  cargo +stable clippy --locked -- -Dwarnings
  cargo +stable clippy --locked -p parity-scale-codec-derive -- -Dwarnings
  cargo +stable clippy --locked --test clippy -- -Dwarnings
  shell: sh -e {0}
  env:
    IMAGE: paritytech/ci-unified:bullseye-1.81.0
    CACHE_ON_FAILURE: false
    CARGO_INCREMENTAL: 0
    Updating crates.io index
error: the lock file /__w/parity-scale-codec/parity-scale-codec/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```

but it works for me locally, I will merge and try to trigger ci in another PR

PR #696, opened 2025-01-31
Enforce `parity-scale-codec-derive` version with `=x.y.z`.

gui1117:
image

Rust doc suggest this approach can be ok: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-metadata

I am not sure myself, but I think it can be better.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ rust-version.workspace = true
 [dependencies]
 arrayvec = { version = "0.7", default-features = false }
 serde = { version = "1.0.217", default-features = false, optional = true }
-parity-scale-codec-derive = { path = "derive", version = "3.6.8", default-features = false, optional = true }
+parity-scale-codec-derive = { path = "derive", version = "=3.7.3", default-features = false, optional = true }
 bitvec = { version = "1", default-features = false, features = ["alloc"], optional = true }
 bytes = { version = "1", default-features = false, optional = true }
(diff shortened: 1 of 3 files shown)


gui1117:
The CI job: `cargo publish -p parity-scale-codec --dry-run`
fails but it is expected. We would need to publish `parity-scale-codec-derive` before publishing `parity-scale-codec`.

serban300 (dismissed):

Cargo.toml
```diff
@@ -13,7 +13,7 @@ rust-version.workspace = true
 [dependencies]
 arrayvec = { version = "0.7", default-features = false }
 serde = { version = "1.0.217", default-features = false, optional = true }
-parity-scale-codec-derive = { path = "derive", version = "3.6.8", default-features = false, optional = true }
+parity-scale-codec-derive = { path = "derive", version = "=3.7.3", default-features = false, optional = true }
```
serban300, line 16: Makes sense, but can't we do `parity-scale-codec-derive = { path = "derive", default-features = false, optional = true }` ?

serban300: Oh, ok

serban300 (approved):

gui1117:
Then I changed the CI a bit.
* do publish dry-run for `parity-scale-codec-derive` instead of `parity-scale-codec`
* publish job: publish both crates in the correct order.

alvicsam (approved):

Issue #713, opened 2025-03-07
Clippy errros with "casting `usize` to `u8` may truncate the value"


cmichi:
IIRC there was a compilation error about the `usize` to `u8` cast a while ago, the release was yanked at the time, and it was fixed. clippy still errors because of it though:

`main.rs`:
```rust
use parity_scale_codec::{
    Decode,
    Encode,
};

#[derive(Encode, Decode)]
pub enum Foo {
    Bar,
}

fn main() { }
```

`Cargo.toml`:
```toml
[package]
name = "foobar"
edition = "2024"

[dependencies]
parity-scale-codec = { version = "=3.7.4", features = ["derive"] }

[workspace]
```

The following errors appear for `clippy`:
```bash
$ cargo +1.85 clippy -- -Dclippy::cast_possible_truncation  
    Checking foobar v0.0.0 (/tmp/foobar)
error: casting `usize` to `u8` may truncate the value
 --> src/main.rs:8:5
  |
8 |     Bar,
  |     ^^^
  |
  = help: if this is intentional allow the lint with `#[allow(clippy::cast_possible_truncation)]` ...
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation
  = note: requested on the command line with `-D clippy::cast-possible-truncation`
help: ... or use `try_from` and handle the error accordingly
  |
8 |     Bar::try_from(Encode),
  |     ~~~~~~~~~~~~~~~~~~~~~

error: casting `usize` to `u8` may truncate the value
 --> src/main.rs:8:5
  |
8 |     Bar,
  |     ^^^
  |
  = help: if this is intentional allow the lint with `#[allow(clippy::cast_possible_truncation)]` ...
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation
help: ... or use `try_from` and handle the error accordingly
  |
8 |     Bar::try_from(Encode),
  |     ~~~~~~~~~~~~~~~~~~~~~

error: casting `usize` to `u8` may truncate the value
 --> src/main.rs:8:5
  |
8 |     Bar,
  |     ^^^
  |
  = help: if this is intentional allow the lint with `#[allow(clippy::cast_possible_truncation)]` ...
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation
help: ... or use `try_from` and handle the error accordingly
  |
8 |     Bar::try_from(Decode),
[...]

PR #715, opened 2025-03-07
Make clippy happy for some casts

bkchr:
Closes: https://github.com/paritytech/parity-scale-codec/issues/713

### DIFF
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -45,7 +45,7 @@ pub fn const_eval_check_variant_indexes(
 	let mut recurse_indices = vec![];
 	for (ident, index) in recurse_variant_indices {
 		let ident_str = ident.to_string();
-		// We convert to usize same as in the generated code.
+		// We convert to usize, index should fit in usize.
 		recurse_indices.push(quote_spanned! { ident.span() =>
 			(
 				(#index) as ::core::primitive::usize,
--- a/derive/src/utils.rs
+++ b/derive/src/utils.rs
@@ -65,6 +65,32 @@ pub fn const_eval_check_variant_indexes(
 			#[allow(clippy::unnecessary_cast)]
 			const indices: [(usize, &'static str); #len] = [#( #recurse_indices ,)*];
 
+			const fn search_for_invalid_index(array: &[(usize, &'static str); #len]) -> (bool, usize) {
+				let mut i = 0;
+				while i < #len {
+					if array[i].0 > 255 {
+						return (true, i);
+					}
+
+					i += 1;
+				}
+
+				(false, 0)
+			}
+
+			const INVALID_INDEX: (bool, usize) = search_for_invalid_index(&indices);
+
+			if INVALID_INDEX.0 {
+				let msg = #crate_path::__private::concatcp!(
+					"Found variant `",
+					indices[INVALID_INDEX.1].1,
+					"` with invalid index: `",
+					indices[INVALID_INDEX.1].0,
+					"`. Max supported index is 255.",
+				);
+				::core::panic!("{}", msg);
+			}
+
 			// Returns if there is duplicate, and if there is some the duplicate indexes.
 			const fn duplicate_info(array: &[(usize, &'static str); #len]) -> (bool, usize, usize) {
 				let len = #len;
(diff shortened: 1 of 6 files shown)


derive/src/utils.rs
```diff
@@ -45,10 +45,10 @@ pub fn const_eval_check_variant_indexes(
 	let mut recurse_indices = vec![];
 	for (ident, index) in recurse_variant_indices {
 		let ident_str = ident.to_string();
-		// We convert to u8 same as in the generated code.
+		// We convert to usize same as in the generated code.
 		recurse_indices.push(quote_spanned! { ident.span() =>
 			(
-				(#index) as ::core::primitive::u8,
+				(#index) as ::core::primitive::usize,
```
gui1117, line 51: I think u8 is more accurate, in the implementation of decode and encode, the type is cast to u8, if user writes:
```rust
enum Foo {
  A = 256,
  B = 0,
}
```
then we want to give an error for conflicts.

The code in scale codec could be refactored to put the cast as part of the expression for variant indices but I think we are fine.

bkchr: We have an extra check for this [here](https://github.com/paritytech/parity-scale-codec/blob/49995c8ceeba6785c227e2d5b611e8a49dd5a818/derive/src/utils.rs#L523)

gui1117: This doesn't take the discriminant into account no?

bkchr: That is a good point. I will add a check there, still think this is better than having it fail somewhere with a weird error.

bkchr: https://github.com/paritytech/parity-scale-codec/blob/49995c8ceeba6785c227e2d5b611e8a49dd5a818/derive/src/utils.rs#L493 ha we have this :D

derive/src/utils.rs
```diff
@@ -60,14 +60,41 @@ pub fn const_eval_check_variant_indexes(
 	}
 
 	quote! {
+		#[automatically_derived]
 		const _: () = {
 			#[allow(clippy::unnecessary_cast)]
-			const indices: [(u8, &'static str); #len] = [#( #recurse_indices ,)*];
+			const indices: [(usize, &'static str); #len] = [#( #recurse_indices ,)*];
+
+			const fn search_for_invalid_index(array: &[(usize, &'static str); #len]) -> (bool, usize) {
```
bkchr, line 68: This way we are future proofs for const variant indices.

derive/src/utils.rs
bkchr: @gui1117 please have another look

derive/src/utils.rs
```diff
@@ -45,10 +45,10 @@ pub fn const_eval_check_variant_indexes(
 	let mut recurse_indices = vec![];
 	for (ident, index) in recurse_variant_indices {
 		let ident_str = ident.to_string();
-		// We convert to u8 same as in the generated code.
+		// We convert to usize same as in the generated code.
```
gui1117, line 48:
```suggestion
		// We convert to usize, index should fit in usize.
```

derive/src/utils.rs
gui1117, line 51: in some very rare case it could still trigger truncation if type is like `repr
(u128)` and he is using u128 discriminant, doesn't harm to also have the flag here: `cast_possible_truncation`

derive/src/utils.rs
```diff
@@ -458,12 +485,9 @@ fn check_variant_attribute(attr: &Attribute) -> syn::Result<()> {
 
 			Meta::NameValue(MetaNameValue {
 				path,
-				value: Expr::Lit(ExprLit { lit: Lit::Int(lit_int), .. }),
+				value: Expr::Lit(ExprLit { lit: Lit::Int(_), .. }),
 				..
-			}) if path.get_ident().map_or(false, |i| i == "index") => lit_int
-				.base10_parse::<u8>()
-				.map(|_| ())
-				.map_err(|_| syn::Error::new(lit_int.span(), "Index must be in 0..255")),
```
gui1117, line 466: In theory this check is still necessary, when we call `variant_index`:
https://github.com/paritytech/parity-scale-codec/blob/dacc95e2dc8eeb84b94175670b63cb7db1f09327/derive/src/utils.rs#L138-L140

somebody writing an index more than `2^64` will make the macro panic.

Though quite unlikely.

gui1117 (approved):

PR #721, opened 2025-03-17
Make `mem_size_of_btree` pub but gated

seadanda:
This allows it to be used in e.g. [`BoundedBTree` decoding](https://github.com/paritytech/parity-common/pull/906).

Feature gated to avoid the details of the implementation becoming part of the public API.
Could be persuaded to drop the gate, but it has already been done for `MaxEncodedLen` etc.


serban300 (commented):
LGTM ! But would also like to check if @bkchr or @gui1117 have any concerns related to this

PR #754, opened 2025-07-28
Fix derive(MaxEncodedLen) with codec(compact/encoded_as)

### LINKED ISSUE #302: MaxEncodedLen macro is not accurate for compact fields, and skipped fields and variants
AFAICT MaxEncodedLen doesn't read the attributes `#[codec(skip)]` and `#[codec(encode_as..)]` and `#[codec(compact)]`.

Thus it can result in an inaccurate len.

For instance the type:
```
#[derive(MaxEncodedLen, Encode)]
struct Foo(#[codec(compact)] u8);
```
will have max encoded len of 1 but can actually encode in 2 bytes.

zdave-parity:
Fixes #302.

### DIFF
--- a/derive/src/max_encoded_len.rs
+++ b/derive/src/max_encoded_len.rs
@@ -17,10 +17,13 @@
 
 use crate::{
 	trait_bounds,
-	utils::{codec_crate_path, custom_mel_trait_bound, has_dumb_trait_bound, should_skip},
+	utils::{
+		codec_crate_path, custom_mel_trait_bound, get_compact_type, get_encoded_as_type,
+		has_dumb_trait_bound, should_skip,
+	},
 };
-use quote::{quote, quote_spanned};
-use syn::{parse_quote, spanned::Spanned, Data, DeriveInput, Field, Fields};
+use quote::{quote, ToTokens};
+use syn::{parse_quote, spanned::Spanned, Data, DeriveInput, Error, Field, Fields};
 
 /// impl for `#[derive(MaxEncodedLen)]`
 pub fn derive_max_encoded_len(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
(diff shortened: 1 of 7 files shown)


derive/src/utils.rs
```diff
@@ -164,7 +164,7 @@ pub fn get_encoded_as_type(field: &Field) -> Option<TokenStream> {
 			if nv.path.is_ident("encoded_as") {
 				if let Expr::Lit(ExprLit { lit: Lit::Str(ref s), .. }) = nv.value {
 					return Some(
-						TokenStream::from_str(&s.value())
+						s.parse()
```
zdave-parity, line 167: AFAICT the span of `s` is properly preserved when using `s.parse()`. This improves error messages which refer to the encoded-as type.

derive/src/max_encoded_len.rs
```diff
@@ -79,19 +82,21 @@ fn fields_length_expr(fields: &Fields, crate_path: &syn::Path) -> proc_macro2::T
 	//   0
 	//     .saturating_add(<type of first field>::max_encoded_len())
 	//     .saturating_add(<type of second field>::max_encoded_len())
-	//
-	// We match the span of each field to the span of the corresponding
-	// `max_encoded_len` call. This way, if one field's type doesn't implement
-	// `MaxEncodedLen`, the compiler's error message will underline which field
```
zdave-parity, line 85: This was just wrong AFAICT. `quote_spanned` applies the given span only to tokens created by `quote_spanned` itself. The error message the compiler produces when `MaxEncodedLen` is not implemented refers to the non-implementing type. The type tokens were _not_ created by `quote_spanned` and so the span provided had no bearing on the error message.

zdave-parity:
> Looks good, but please add some test

Added a few tests.

bkchr (approved):

gui1117:
Note that this is a breaking change (although I agree it is a fix).

This compiled:
```diff
diff --git a/tests/max_encoded_len.rs b/tests/max_encoded_len.rs
index 5a41a10..4a86285 100644
--- a/tests/max_encoded_len.rs
+++ b/tests/max_encoded_len.rs
@@ -307,3 +307,10 @@ fn skip_enum_struct_test() {
        let _ = Enum::<NoCodecType, NoCodecNoDefaultType>::A(NoCodecNoDefaultType);
        let _ = StructNamed::<NoCodecType> { a: NoCodecType, b: 0 }.a;
 }
+
+use parity_scale_codec::HasCompact;
+#[derive(Encode, MaxEncodedLen)]
+struct Foo<T: HasCompact + MaxEncodedLen> {
+       #[codec(compact)]
+       t: T,
+}
```

But now fails with:
```
error[E0277]: the trait bound `<T as HasCompact>::Type: MaxEncodedLen` is not satisfied
   --> tests/max_encoded_len.rs:312:18
    |
312 | #[derive(Encode, MaxEncodedLen)]
    |                  ^^^^^^^^^^^^^ the trait `MaxEncodedLen` is not implemented for `<T as HasCompact>::Type`
    |
    = note: this error originates in the derive macro `MaxEncodedLen` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `<T as HasCompact>::Type: Encode` is not satisfied
   --> tests/max_encoded_len.rs:312:18
    |
312 | #[derive(Encode, MaxEncodedLen)]
    |                  ^^^^^^^^^^^^^ the trait `WrapperTypeEncode` is not implemented for `<T as HasCompact>::Type`
    |
    = note: required for `<T as HasCompact>::Type` to implement `Encode`
```

I think polkadot-sdk or some ecosystem crates may fail to compile, we should check their compilation before releasing a minor version.
If polkadot-sdk and other crates compile, maybe it is ok.

